PR #1, opened 2021-06-21
feat: add `GET /car/:cid`

olizilla:
- Get CARs via GET /car/:cid
- Get the size via HEAD /car/:cid
- Document the `wrangler.toml`
  - I figured out the wierdness with `worker_dev` not getting inherited and changing where it looks for account_id
  - Uses `javascript` deploy type instead of `webpack` so we use our own build step rather than opaque wrangler magic.
- Add prettier. Not adding pre-commit behaviours. Prettier can fix things in CI for us if we dont run it locally.

License: MIT


olizilla:
This is a first pass to wire things up. We would retrieve a CAR from cluster and filecoin when all the infra is in place.

packages/api/src/index.js
```diff
@@ -0,0 +1,41 @@
...
+router.all('*', () => new Response('Not Found.', { status: 404 }))
+
+async function serverError(error) {
+  return new Response(error.message || 'Server Error', {
+    status: error.status || 500,
+  })
+}
+
+addEventListener('fetch', (event) =>
+  event.respondWith(router.handle(event.request, {}, event).catch(serverError))
```
olizilla, line 40: This signature passes things through to the handlers in a way that mimics the pattern that will be in place when cloudflare fully supports sending them am ESM module that exports a handler as per https://developers.cloudflare.com/workers/cli-wrangler/configuration#modules 
```js
import html from './index.html'
export default {  
// * request is the same as `event.request` from the service worker format  
// * waitUntil() and passThroughOnException() are accessible from `ctx` instead of `event` from the service worker format  
// * env is where bindings like KV namespaces, Durable Object namespaces, Config variables, and Secrets  
// are exposed, instead of them being placed in global scope.  
  async fetch(request, env, ctx) {
    const headers = { 'Content-Type': 'text/html;charset=UTF-8' }
    return new Response(html, { headers })
  }}
```

but i think it'll be annoying that we can't just copy paste handlers from nft.storage, which just pass in the `event`, so am going to fix thist PR to pass the request around in the same way.

cc @alanshaw

olizilla: oh, the router `itty-router` expects the first argument to be the `event.request` so, even if i switch it i'm gonna need to update the method signatures on handlers from nft.storage from `handle(event)` to `handle(request, event)`.

olizilla: > The router handle expects only the request itself, but passes along any additional params to the handlers/middleware. For example, to access the event itself within a handler (e.g. for event.waitUntil()), you could simply do this:
```
addEventListener('fetch', event =>
  event.respondWith(router.handle(event.request, event))
)
```
– https://github.com/kwhitley/itty-router#a-few-notes-about-this

packages/api/src/car.js
```diff
@@ -0,0 +1,70 @@
...
+  const reader = response.body.getReader()
+  let size = 0
+  while (true) {
+    const { done, value } = await reader.read()
+    if (done) {
+      break
+    }
+    size += value.byteLength
+  }
+  return size
```
alanshaw, line 69: Previously I've thought it necessary to use `releaseLock()` e.g. https://github.com/alanshaw/stream-to-it/blob/fe19c61b9b5b3912c84a2bb91a41071422246455/source.js#L7-L19 I'm not sure if applicable here?

alanshaw (approved):

packages/api/src/index.js
olizilla: From conversation with @alanshaw we have a mild preference in favour of switching to bundling the worker as esm (which the wrangler docs suggest will be the recommended format in the future). It's assumed it will be easier to test than a _sort of service-worker_, with magic globals for secrets and other cloudflare features. In esm mode, the env is passed to the fetch function explicitly with all the magic properties on.

https://developers.cloudflare.com/workers/cli-wrangler/configuration#modules

packages/api/src/car.js
olizilla: > A reader or writer also has the capability to release its lock, which makes it no longer active, and allows further readers or writers to be acquired.
– https://streams.spec.whatwg.org/#lock

i'm not expecting anything else to read this stream, but I guess it'd be harmless to call `releaseLock` here incase this function gets extracted into a more general utils thing at some point.

PR #2, opened 2021-06-24
feat: client `.store` & `.get`

olizilla:
Skeleton of the client ported from nft.storage. 

- Implements .`store(car)` and `.get(cid)`.
- Test pass in browser, cjs, and esm flavour with 100% coverage.
- Use `smoke` to create api mock, as we do in nft.storage api tests, to simplify test set up.


packages/client/package.json
```diff
@@ -2,18 +2,75 @@
...
+    "typedoc": "typedoc --entryPoints src  --out ../../docs/client",
+    "prepare": "npm run build"
+  },
+  "dependencies": {
+    "@ipld/car": "^2.1.0",
+    "@web-std/blob": "^2.1.0",
+    "@web-std/fetch": "^2.0.1"
+  },
+  "devDependencies": {
+    "@ipld/dag-cbor": "^5.0.0",
```
olizilla, line 35: this list of devDeps was copied from nft.storage. We don't need all these just yet. I'll whittle it down.

packages/client/src/lib.js
```diff
@@ -0,0 +1,165 @@
...
+
+  /**
+   * @hidden
+   * @param {string} token
+   */
+  static headers(token) {
+    if (!token) throw new Error('missing token')
+    return {
+      Authorization: `Bearer ${token}`,
+      'X-Client': 'filecoin.storage',
```
olizilla, line 69: Add a custom header so we can figure out how many folks actually use this thing

packages/client/package.json
```diff
@@ -2,18 +2,66 @@
...
+    "uvu": "0.5.1"
+  },
+  "types": "./dist/src/lib.d.ts",
+  "module": "./src/lib.js",
+  "main": "./dist/src/lib.cjs",
+  "exports": {
+    ".": {
+      "browser": "./src/lib.js",
+      "require": "./dist/src/lib.cjs",
+      "node": "./src/lib.js"
```
alanshaw, line 56: Should this be "import"? I've not seen "node" before.

packages/client/src/lib.js
alanshaw: Are we interested in version or language?

alanshaw (approved):

packages/client/package.json
olizilla: hmm good question, this part was copied from nft.storage client, and I have not taken the time to fully understand it yet.

packages/client/src/lib.js
olizilla: Let's start with this and I'll ask around to see what folks think. I'm inclined to track less.

PR #3, opened 2021-06-25
feat: cors

alanshaw:
Adds CORS config from nft.storage, a test, a test runner and some config to run tests in CI.


olizilla:
Removed package-lock.json from gitignore and committed, which should fix the test error.
> Run bahmutov/npm-install@v1
> Error: ENOENT: no such file or directory, open '/home/runner/work/filecoin.storage/filecoin.storage/package-lock.json'

packages/api/src/index.js
```diff
@@ -1,12 +1,14 @@
...
 import { carHead, carGet, carPut, carPost } from './car.js'
 
 const router = Router()
 
-router.get('/car/:cid', carGet)
-router.head('/car/:cid', carHead)
-router.put('/car/:cid', carPut)
-router.post('/car', carPost)
+router.options('*', corsOptions)
+router.get('/car/:cid', withCorsHeaders(carGet))
```
olizilla, line 8: Neat 👍

packages/api/test/scripts/worker-globals.js
```diff
@@ -0,0 +1,6 @@
+export const caches = {
```
olizilla, line 1: Once we switch to sending ESM to cloudflare we can switch this to something we pass in explicitly to tests that need it, which'll is nice.

packages/api/package.json
```diff
@@ -11,9 +11,12 @@
     "publish": "wrangler publish --env $(whoami)",
     "build": "webpack",
     "format": "prettier --write '**/*.{js,css,json,md}'",
-    "test": "say sorry"
+    "test": "playwright-test \"test/**/*.spec.js\" --sw src/index.js"
```
olizilla, line 14: How does this actually work? Is `playwright-test` actually injecting this as a service worker in a chrome? `src/index.js` isn't a valid service worker without some pre-processing... it's esm but it uses a node style import which I would expect to fail in a browser service worker context. Also how does it know what origin the service worker can catch fetch events to... and https and and.... so many questions.

olizilla: woosh. it runs it through esbuild!

https://github.com/hugomrdias/playwright-test/blob/096b773636311eff6038fe7a71fc0f38eea6221a/src/utils/build-sw.js#L20-L61

packages/api/test/cors.spec.js
```diff
@@ -0,0 +1,13 @@
+/* eslint-env mocha, browser */
+import assert from 'assert'
+
+const endpoint = 'http://testing.filecoin.storage'
```
olizilla, line 4: There is a magic endpoint string in the baked into nft.storage router, but that isn't the case here. This test fails if you have no network connection. I think it's trying to hit ipfs.io to fetch the CAR file and fails becuase `res.ok` is not ok.

olizilla (commented):
This is good! 

There is an interesting question around what is the right test set up for cloudflare workers. Testing them as a real service worker might be the best way, but it's worth documenting what is going on, particularly that it's going to compile the worker with a different build tool. 

It's like this is kind of an integration test becuase it needs all the services or mocks to be available, but also kind of a parallel universe because workers are not local service workers.

olizilla:
We're gonna merge this and explore pushing esm to cloudlare in a subsequent PR

PR #4, opened 2021-06-26
feat: client.get: (cid) => Responsish

olizilla:
- Have client.get return an object with `files` and `filesIterator` methods on it.
- Add rollup config to make a single file esm bundle
- Add bundlesize so we can see the impact of adding CAR support
- Build and test the client in CI.

License: (Apache-2.0 AND MIT)


olizilla:
adding getFiles cia a dep on `ipfs-car` is not so bad for esm bundlesize

```console
❯ npm run test:size
> filecoin.storage@0.0.0 test:size
> bundlesize

 PASS  ./dist/bundle.esm.min.js: 94.49KB < maxSize 200KB (gzip) 
```

olizilla:
oh, that wasn't minified... with the terser plugin added to the rollup config it becomes

```
 PASS  ./dist/bundle.esm.min.js: 55.68KB < maxSize 200KB (gzip) 
```

packages/client/src/lib.js
```diff
@@ -155,6 +159,75 @@ class FilecoinStorage {
...
+ * @returns {Promise<import('./lib/interface.js').IpfsFile>}
+ */
+async function toIpfsFile(e) {
+  const chunks = []
+  for await (const chunk of e.content()) {
+    chunks.push(chunk)
+  }
+
+  // A Blob in File clothing
+  const file = Object.assign(new Blob(chunks), {
```
olizilla, line 191: @vasco-santos @alanshaw To make a browser `File` like thing, I am consuming the UnixFSEntry content... and i'm doing this for both the `res.file` method and the `res.filesIterator` which means, for big big files, you are probably going to have a bad, memory constrained time.

Should `filesIterator` be `unixFsEntryIterator` or similar... with that, we preserve all the information like directory nodes, and in the cli i could just import the existing `writeFiles` function from `ipfs-car` that takes the `AsyncIterable<UnixFsEntry>` and figures out how to write it to disk. 
https://github.com/vasco-santos/ipfs-car/blob/6935c9000703f7d2b0df529688c983987db6fce8/src/unpack/fs.ts#L29-L52

alanshaw: Sounds sensible to me.

packages/client/package.json
```diff
@@ -14,34 +14,41 @@
...
+    "browser-readablestream-to-it": "^1.0.2"
   },
   "devDependencies": {
+    "@rollup/plugin-commonjs": "^19.0.0",
+    "@rollup/plugin-node-resolve": "^13.0.0",
     "@types/mocha": "8.2.2",
+    "bundlesize": "^0.18.1",
     "hundreds": "0.0.9",
-    "ipfs-car": "^0.2.4",
+    "ipfs-car": "^0.3.5",
```
alanshaw, line 43: move to dependencies

packages/client/src/lib.js
```diff
@@ -155,6 +159,75 @@ class FilecoinStorage {
   }
 }
 
+/**
+ * Upgrade a ReadableStream to an AsyncIterable if it itn't already
```
alanshaw, line 163:
```suggestion
 * Upgrade a ReadableStream to an AsyncIterable if it isn't already
```

packages/client/src/lib.js
```diff
@@ -155,6 +159,75 @@ class FilecoinStorage {
...
+ *
+ * ReadableStream (e.g res.body) is asyncIterable in node, but not in chrome, yet.
+ * see: https://bugs.chromium.org/p/chromium/issues/detail?id=929585
+ *
+ * @param {ReadableStream<Uint8Array> | ReadableStream<Uint8Array> & AsyncIterable<Uint8Array>} readable
```
alanshaw, line 168:
```suggestion
 * @param {ReadableStream<Uint8Array>} readable
```

olizilla:
Replaced `filesIterator` with `unixFsIterator` and added a test for a CAR with dirs in.

olizilla:
Need to think some more about it tho, as we still have the problem of conuming the entire fetch response into an in mem CarReader... I think this is going to need something like fetch -> CarBlockIterator -> (indexdbBlockstore | fsBlockstore) -> unixfs-exporter to make it not be mem-bound.

packages/client/src/lib.js
vasco-santos: It can, but I would make it a `ReadableFile` per https://cryptpad.fr/code/#/2/code/edit/lLebfkgGeoMJmkkuhPQS8rTN/
which is really a subset of UnixFsEntry, but something more simple that a user can consume and we can create utilities to help consuming the get content.

I would see us doing something like

```js
const putFiles = readableFiles.fromFs(path: String)
const cid = await put (putFiles: Iterable<ReadableFile>)
const files = await get(cid: CID)
await readableFiles.toFs(files) // Back in fs
```

While for browser users just need to provide a ReadableFile Iterator (Blob, String, Uint8Array) + get one, which should be easy to parse. Using WebFile, we can also provide a `readableFiles.FromWebFile()`, which would convert Web File to Readable File, if users use the picker.

vasco-santos:
> Need to think some more about it tho, as we still have the problem of conuming the entire fetch response into an in mem CarReader... I think this is going to need something like fetch -> CarBlockIterator -> (indexdbBlockstore | fsBlockstore) -> unixfs-exporter to make it not be mem-bound.

Yes, we need to use the Blockstores here! I am wondering in Get if we want to keep the content fetched in the blockstore and return a readable for it. Otherwise, it will need to be an async generator.

packages/client/src/lib.js
vasco-santos: `readableFiles.fromFs` would go through the provided path and create an iterable of

```
{ path: String, content: ReadableStream }
```

which ipfsCar pack would accept.

alanshaw (approved):
This is good! We've discussed a further iteration and roughly documented in https://gist.github.com/alanshaw/fdeaed7df576231a38487e53a3a63c50

Lets get this merged and iterate again.

PR #9, opened 2021-06-30
fix: change filecoin storage to web3 storage in client

vasco-santos:


vasco-santos:
The API test is failling everywhere, but not locally for me. We need to figure this out tomorrow

olizilla:
Agree, issue is here for CI fails https://github.com/web3-storage/web3.storage/issues/11

packages/client/README.md
```diff
@@ -9,10 +9,10 @@ _This module is under developement and is not published yet._
 ## Usage
 
 ```js
-import FilecoinStorage from 'web3.storage'
+import { WebStorage } from 'web3.storage'
```
olizilla, line 12:
```suggestion
import { Web3Storage } from 'web3.storage'
```

packages/client/README.md
```diff
@@ -9,10 +9,10 @@ _This module is under developement and is not published yet._
...
 
 // Construct with token and endpoint
-const client = new FilecoinStorage({ token: apiKey })
+const client = new WebStorage({ token: apiKey })
```
olizilla, line 15:
```suggestion
const client = new Web3Storage({ token: apiKey })
```

olizilla (changes requested):
renaming stuff is harder than it sounds

olizilla (approved):

PR #10, opened 2021-06-30
feat: db

alanshaw:
* Adds a Fauna database schema
* Adds DB functions
    * Registering a user
    * Creating auth tokens
    * Verifying auth tokens
    * Registering an uploaded CAR (not yet used)
    * Finding a user by their "issuer"
    * Finding content by a CID (not yet used)
* Adds tools to import functions and graphql schema to a Fauna database
* Adds a DB client (a really thin wrapper around [`graphql-request`](https://www.npmjs.com/package/graphql-request)) and uses it in the API
* Adds 2 routes:
    * `/user/login` - for logging in using magic.link (not yet possible - needs UI)
    * `/user/tokens` - for creating a token (not yet possible - needs UI)
* Adds a `withAuth` middleware that verifies an auth token / magic.link token and adds an `auth` property onto the request object.
* Adds a script that creates a test user and auth token so we can use the authenticated API

### Notes

This is a small stepping stone towards the glorious future. I've extracted the minimum schema and some of the tools from niftysave and reused as much as possible of the auth code (jwt signing etc) from nft.storage.

Niftysave has migrations and a typed client generated from the schema which would be super cool but the process is currently complicated and still being ironed out so I've gone for the simpliest thing here and we'll upgrade later.

### Schema diagram

web3 storage schema(2)


packages/db/README.md
```diff
@@ -0,0 +1,46 @@
+# DB
```
alanshaw, line 1: @alanshaw add the schema diagram from the PR description here

packages/db/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Getting Started
```
vasco-santos, line 3: We need to update the main README pointing to the instruction in the single packages. Given that we still not have for the API, this can probably be a follow up PR

packages/db/fauna/schema.graphql
```diff
@@ -0,0 +1,306 @@
...
+  """
+  location: PinLocation! @relation
+
+  """
+  Pinning status at this location.
+  """
+  status: PinStatus!
+
+  """
+  Last time the status was last updated.
```
vasco-santos, line 117:
```suggestion
  Last time the status was updated.
```

vasco-santos (approved):

olizilla:
> Adds a DB client (a really thin wrapper around graphql-request) and uses it in the API
What convinced you to query the graphql api from the api? After Hugo's warnings about the latency on the API, I'd assumed we'd go with the FQL and the existing fauna db client https://github.com/fauna/faunadb-js

packages/api/README.md
```diff
@@ -21,6 +21,15 @@ workers_dev = true
 account_id = "<what does the `wrangler whoami` say>"
 ```
 
+- `npm run build` - Build the bundle
+- Add secrets
+
+    ```sh
+    wrangler secret put MAGIC_SECRET_KEY --env $(whoami) # Get from magic.link account
```
olizilla, line 28: I hit a problem setting up nft.storage where I could not set secrets before I had published the worker for the first time. It might have been a one off, but worth checking, or just move this step after the publish step to avoid the possibility.

packages/api/README.md
```diff
@@ -21,6 +21,15 @@ workers_dev = true
 account_id = "<what does the `wrangler whoami` say>"
 ```
 
+- `npm run build` - Build the bundle
```
olizilla, line 24: Did you need to build the bundle manually? `wrangler dev|publish` should be running the build command as defined in the wrangler.toml.

packages/api/scripts/get-auth-token.js
```diff
@@ -0,0 +1,56 @@
+import { DBClient, gql } from '@web3-storage/db'
+import * as JWT from '../src/utils/jwt.js'
+import subtle from 'subtle'
```
olizilla, line 3: node has a crispy fresh `webcrypto.subtle`!
https://nodejs.org/api/webcrypto.html#webcrypto_web_crypto_api

alanshaw:
> > Adds a DB client (a really thin wrapper around graphql-request) and uses it in the API
> 
> What convinced you to query the graphql api from the w3 api? After Hugo's warnings about the latency on the API, I'd assumed we'd go with the FQL and the existing fauna db client https://github.com/fauna/faunadb-js

* I did not see noticeable query latency, but also don't have any data in my database which could be a factor.
* I'm midly in favour of having an API to the database (GraphQL) where you can run queries that have been deemed safe.
* I'd like to avoid FQL as much as possible. It's not familiar to many web developers and not super easy to pick up.
* Using GraphQL will eventually allow us to use a typed client (as in niftysave) - I didn't use it here because it won't work in workers in it's current state and I wanted to get something up and running quickly.

packages/api/webpack.config.js
```diff
@@ -1,7 +1,22 @@
...
+const require = createRequire(import.meta.url)
+
 export default {
   target: 'webworker',
   mode: 'development',
   devtool: 'cheap-module-source-map', // avoid "eval": Workers environment doesn't allow it
-  plugins: [],
+  plugins: [
+    new webpack.ProvidePlugin({
+      process: 'process/browser',
```
olizilla, line 12: Which thing is using `process` in the browser? My assumption is the new `api/scripts` files are intended to be run from node

packages/api/README.md
alanshaw: When I did this it attempted to create a worker for me, but needed a built bundle - thats why I added the `npm run build` step prior to this one.

packages/api/README.md
alanshaw: Yes because first time round when you try to set secrets it'll automatically try and create a worker for you and needs a built bundle.

packages/api/scripts/get-auth-token.js
alanshaw: No way! Nice, will change.

packages/api/webpack.config.js
alanshaw: `@magic-sdk/admin`

olizilla:
TODO:
- add region to PinLocation
- add schema diagram
- use node webcrypto

olizilla (approved):
Happy for this to go in so we can unblock and iterate on it, the todos from our conversation are listed.

packages/api/package.json
```diff
@@ -16,11 +16,17 @@
...
-    "webpack": "^5.39.1",
+    "process": "^0.11.10",
+    "stream-browserify": "^3.0.0",
+    "webpack": "^5.42.0",
     "webpack-cli": "^4.7.2"
   },
   "dependencies": {
+    "@magic-ext/oauth": "^0.8.0",
+    "@magic-sdk/admin": "^1.3.0",
+    "@web3-storage/db": "^0.0.0",
```
olizilla, line 29: @alanshaw on the npms... do we want 
- `@web3-storage/db` 
- `@web3storage/db` - matches the hyphenless @nftstorage 
- `@dotstorage/web3.storage-db` - we talked about a `dotstrorage` org, but then there would be times you'd want to differentiate which app a module was actually for.

I guess we would not actually publish the db to npm, but we'd probably want to use a prefix that we controlled regardless, so this is mostly for illustration purposes

olizilla: am gonna go with `@web3-storage/db` as you have it.

alanshaw: We had `nftstorage` on github so I made `@nftstorage` on npm too for consistency and so we have `web3-storage` on github here so I went for `@web3-storage` on npm. Even though we won't publish any modules...yet, we might want to publish web3.storage related packages to `@web3-storage` eventually so worth land grabbing. WDYT?

The other packages use a prefix e.g. `web3.storage-api` - we should probably change these also.

I think it is ok for the JS client to use `web3.storage` is it's name though.

PR #12, opened 2021-07-01
feat: put files (client)

vasco-santos:
This PR adds:

- add Web3Files API in the client
- adds Node.js and Browser examples
  - They work on my local setup, but I flagged them as WIP until we get the end to end thing

Needs:
- [x] #9 with all the renames
- [x] https://github.com/web3-storage/web3-file/pull/4 + use fromText to create test/examples
- [x] https://github.com/web3-storage/ipfs-car/pull/49

PR #13, opened 2021-07-01
feat: client.get uses unpackStream with blockstore

olizilla:
- Update client.get to use https://github.com/web3-storage/ipfs-car/pull/46
  - uses a blockstore during unpack so should be able to deal with cars larger than mem now.
  - ipfs-car adopts the ReadableStream vs AsyncIterator shim which is nice
- Adds `clean` target called before `build` to ensure we're fresh.

Needs https://github.com/web3-storage/web3-file/issues/7

Builds on #12 

License: (Apache-2.0 AND MIT)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10791,6 +10791,18 @@
         "node": ">=8"
       }
     },
+    "node_modules/nyc/node_modules/strip-ansi": {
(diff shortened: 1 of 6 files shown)


olizilla:
Closing in favour of the mergable #16

PR #15, opened 2021-07-02
feat: onRootCidReady

vasco-santos:
We know the cid once the car is generated from the given files. We can give them to the user already, as this will allow use cases of parallel processing data.

This PR adds a function option, similar to the provided for tracking upload progress that a user can rely

### DIFF
--- a/packages/client/src/lib.js
+++ b/packages/client/src/lib.js
@@ -91,22 +91,26 @@ class Web3Storage {
    * @param {API.PutOptions} [options]
    * @returns {Promise<API.CIDString>}
    */
-  static async put({ endpoint, token }, files, { onStoredChunk, maxRetries = MAX_PUT_RETRIES } = {}) {
+  static async put({ endpoint, token }, files, { onCarCreated, onStoredChunk, maxRetries = MAX_PUT_RETRIES } = {}) {
     const url = new URL(`/car`, endpoint)
     const headers = Web3Storage.headers(token)
     const targetSize = MAX_CHUNK_SIZE
 
-    let root
+    let carRoot
     const blockstore = new Blockstore()
 
     try {
-      const { out } = await pack({
+      const { out, root } = await pack({
         input: Array.from(files).map((f) => ({
           path: f.name,
           content: f.stream()
         })),
         blockstore
       })
+      carRoot = root.toString()
+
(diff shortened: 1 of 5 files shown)


packages/client/src/lib.js
```diff
@@ -91,22 +91,26 @@ class Web3Storage {
    * @param {API.PutOptions} [options]
    * @returns {Promise<API.CIDString>}
    */
-  static async put({ endpoint, token }, files, { onStoredChunk, maxRetries = MAX_PUT_RETRIES } = {}) {
+  static async put({ endpoint, token }, files, { onCarCreated, onStoredChunk, maxRetries = MAX_PUT_RETRIES } = {}) {
```
olizilla, line 94: `onCarCreated` is accurate, but we could go with `onRootCidReady` or similar to keep the CARs hidden here... @alanshaw what say you?

alanshaw: SGTM

vasco-santos: renamed 🎉

packages/client/src/lib.js
```diff
@@ -143,16 +147,13 @@ class Web3Storage {
         }
       )
 
-      for await (const cid of upload(splitter.cars())) {
-        root = cid
-      }
+      for await (const _ of upload(splitter.cars())) {}
```
olizilla, line 150: Should we verify here that the service returns always returns the root CID we expect?

olizilla (approved):

packages/client/src/lib.js
vasco-santos: good point, I assume the service is trustworthy. Otherwise, nothing also guarantees that is sends back CID x, but the content has CID y.
What would you see happening if by any reason it was different CID? Would we throw?

olizilla: yes big throw! the contract is broken! the service has been compromised! pack your bags! RUN!

olizilla: We're aimiing for "service is a dumb store of the Blocks you give it", and the promse should be "I will store your blocks and pin your root CID", such that when you later `get` that CID, you get back the exact same blocks that we also locally verify.

alanshaw (approved):

PR #16, opened 2021-07-02
feat: feat: client.get uses unpackStream

olizilla:
- Update client.get to use https://github.com/web3-storage/ipfs-car/pull/46
  - uses a blockstore during unpack so should be able to deal with cars larger than mem now.
  - ipfs-car adopts the ReadableStream vs AsyncIterator shim which is nice
- Adds `clean` target called before `build` to ensure we're fresh.

License: (Apache-2.0 AND MIT)

### DIFF
--- a/packages/client/test/get.spec.js
+++ b/packages/client/test/get.spec.js
@@ -27,14 +27,14 @@ describe('get', () => {
     const files = await res.files()
     for (const file of files) {
       assert.is(
-        file.cid,
+        file.cid.toString(),
         cid,
         'in a CAR with 1 file, the file name should match the CAR root'
       )
       assert.is(
-        file.webkitRelativePath,
+        file.name,
         cid,
-        `webkitRelativePath (${file.webkitRelativePath}) should be (${cid})`
+        `name (${file.name}) should be (${cid})`
       )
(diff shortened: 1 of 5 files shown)


packages/client/test/get.spec.js
```diff
@@ -49,6 +49,15 @@ describe('get', () => {
     assert.ok(res.ok)
     const files = await res.files()
     assert.is(files.length, 3, 'should contain 3 files')
+    assert.is(files[0].name, 'dr-is-tired.jpg')
```
vasco-santos, line 52: we should not depend on network requests in CI, but let's deal with this later

vasco-santos (approved):
LGTM

packages/client/test/get.spec.js
olizilla: is a local mock 🎉

olizilla: https://github.com/web3-storage/web3.storage/blob/a95bcb2f7d8a2164cc6ef7e90b7d72bb1b0f096a/packages/client/test/mocks/api/get_car%23%40cid.js#L5-L10

PR #21, opened 2021-07-05
feat: temporary UI

alanshaw:
Mostly copied from the nft.storage website and bits trimmed out.

This is COMPLETELY disposable and only exists temporarily to try things out.

Screenshot 2021-07-03 at 11 37 43

### DIFF
--- a/packages/website/next-env.d.ts
+++ b/packages/website/next-env.d.ts
@@ -1,0 +1,3 @@
+/// <reference types="next" />
+/// <reference types="next/types/global" />
+/// <reference types="next/image-types/global" />
--- a/packages/website/package.json
+++ b/packages/website/package.json
@@ -1,0 +1,81 @@
+{
+  "name": "@web3-storage/website",
+  "version": "0.0.0",
+  "description": "web3.storage website",
+  "private": true,
+  "scripts": {
+    "start": "next dev",
+    "build": "next build && next export",
+    "test": "eslint './**/*.js' && tsc --build"
+  },
+  "dependencies": {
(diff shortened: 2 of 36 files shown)


olizilla:
pls to resolve conflicts @alanshaw

packages/website/next-env.d.ts
```diff
@@ -0,0 +1,3 @@
+/// <reference types="next" />
```
olizilla, line 1: what is?

alanshaw: IDK next.js make it

packages/website/package.json
```diff
@@ -0,0 +1,81 @@
...
+        }
+      ],
+      "react/display-name": 1
+    },
+    "settings": {
+      "react": {
+        "version": "detect"
+      }
+    }
+  },
```
olizilla, line 75: note to self: we should unify this with the prettier config at the root of the project

packages/website/styles/global.css
```diff
@@ -0,0 +1,12 @@
...
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+  }
+  .table-responsive {
+    display: block;
+    width: 100%;
+    overflow-x: auto;
+  }
```
olizilla, line 11: _it's the smell_

olizilla (approved):

packages/website/styles/global.css
alanshaw: 😉 https://github.com/ipfs-shipyard/nft.storage/pull/61#discussion_r618455074

PR #22, opened 2021-07-05
chore: staging api config

olizilla:
- update wrangler.toml with staging env
- update github action to publish staging api on merge to main

and out-of-band we have:
- web3.storage site created on Cloudflare
- nameservers updated to point web3.storage at Cloudflare
- DNSSEC enabled on web3.storage
- Scoped api token for editing workers on web3.storage added as `CF_API_TOKEN`

License: (Apache-2.0 AND MIT)

### DIFF
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -3,9 +3,15 @@ on:
   push:
     branches:
       - main
+    paths:
+      - 'packages/api/**'
+      - '.github/workflows/api.yml'
   pull_request:
     branches:
       - main
+    paths:
+      - 'packages/api/**'
+      - '.github/workflows/api.yml'
 jobs:
   test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


olizilla:
This is a reduced version of what is powering deploys in nft.storage. @hugomrdias can you give it a quick review pls.

.github/workflows/api.yml
```diff
@@ -3,9 +3,15 @@ on:
   push:
     branches:
       - main
+    paths:
+      - 'packages/api/**'
+      - '.github/workflows/api.yml'
   pull_request:
```
alanshaw, line 9: Should we just remove this part and the `if` condition below?

olizilla: No, we want the test jon to run for pull requests that affect the api code. I could split these out into MOAR yamlz, if you'd prefer.

alanshaw: Ah, no, no more yamls - I missed the existing test job for some reason 🤦 .

alanshaw (approved):

PR #23, opened 2021-07-05
feat: user tokens get

alanshaw:
* Adds an endpoint for retrieving user tokens `GET /user/tokens`
* Adds tests for `GET /user/tokens` and `POST /user/tokens` (creating a token)
* Adds a smoke mock for `db`


packages/api/src/env.js
```diff
@@ -8,14 +8,12 @@ import { DBClient } from '@web3-storage/db'
...
 
   env.db = new DBClient({
-    endpoint: getConfig('FAUNA_ENDPOINT'),
-    token: getConfig('FAUNA_KEY')
+    endpoint: env.FAUNA_ENDPOINT || FAUNA_ENDPOINT,
+    token: env.FAUNA_KEY || FAUNA_KEY
   })
 
-  env.SALT = getConfig('SALT')
+  env.SALT = env.SALT || SALT
```
alanshaw, line 18: esbuild was thinking these globals were "dead code" and removing them.

packages/api/src/index.js
```diff
@@ -35,12 +37,13 @@ router.get('/', () => {
...
 
-router.all('*', () => new Response('Not Found.', { status: 404 }))
+router.all('*', () => new JSONResponse({ message: 'Not Found' }, { status: 404 }))
 
-async function serverError(error) {
-  return new Response(error.message || 'Server Error', {
-    status: error.status || 500,
-  })
+function serverError (error) {
+  console.error(error.stack)
```
vasco-santos, line 43:
```suggestion
```

vasco-santos: do we want to keep this?

packages/api/test/scripts/worker-globals.js
```diff
@@ -1,6 +1,11 @@
 export const caches = {
   default: {
-    match: () =>  null,
+    match: () => null,
     put: () => {}
   }
 }
+
+export const SALT = 'test-salt'
```
vasco-santos, line 8: Will we do the ENV setup in follow up?

packages/api/src/env.js
olizilla: yes, i'm not super comfortable with their being an alternative bundling chain that is only used in the playwright-test --sw flow.

packages/api/package.json
```diff
@@ -11,15 +11,20 @@
     "publish": "wrangler publish --env $(whoami)",
     "build": "webpack",
     "format": "prettier --write '**/*.{js,css,json,md}'",
-    "test": "playwright-test \"test/**/*.spec.js\" --sw src/index.js"
+    "test": "npm run test:e2e",
+    "test:e2e": "npm-run-all -p -r test:e2e:**",
+    "test:e2e:mock:db": "smoke -p 9086 test/mocks/db",
+    "test:e2e:playwright": "playwright-test \"test/**/*.spec.js\" --sw src/index.js"
```
olizilla, line 17: minor: if it so pleases you, I think the targets in the client are a little more pleasing. Not much but it would be good to have a similar pattern in both, either way.
```suggestion
    "test": "npm-run-all -p -r mock:api test:e2e",
    "test:e2e": "playwright-test \"test/**/*.spec.js\" --sw src/index.js"
    "mock:db": "smoke -p 9086 test/mocks/db",
```

packages/api/src/index.js
olizilla: i think is helpful to see errors in the log

packages/api/src/user.js
```diff
@@ -172,3 +177,28 @@ export async function userTokensPost (request, env) {
...
+ * Retrieve user auth tokens.
+ *
+ * @param {AuthenticatedRequest} request
+ * @param {import('./env').Env} env
+ */
+export async function userTokensGet (request, env) {
+  const res = await env.db.query(gql`
+    mutation FindAuthTokensByUser($user: ID!) {
+      # Paginated but users are probably not going to have tons of these.
+      # Note: 100,000 is the max page size.
```
olizilla, line 191: Should we just put a cap on it? Let's prevent users making more than 100,000 tokens. They can get in touch for the enterprise license if they need more.

packages/api/src/user.js
```diff
@@ -172,3 +177,28 @@ export async function userTokensPost (request, env) {
...
+
+/**
+ * Retrieve user auth tokens.
+ *
+ * @param {AuthenticatedRequest} request
+ * @param {import('./env').Env} env
+ */
+export async function userTokensGet (request, env) {
+  const res = await env.db.query(gql`
+    mutation FindAuthTokensByUser($user: ID!) {
```
olizilla, line 189: Why is `FindAuthTokensByUser` a `mutation`? Sounds like a `query`

packages/api/src/index.js
alanshaw: Yes keep, for my sanity!

packages/api/test/scripts/worker-globals.js
alanshaw: no? not sure I understand?

packages/api/package.json
alanshaw: Yes lets do it that way.

packages/api/src/utils/fetch.js
```diff
@@ -0,0 +1 @@
+export default globalThis.fetch.bind(globalThis)
```
olizilla, line 1: who needs this?

packages/api/test/scripts/worker-globals.js
vasco-santos: If we will merge this and do the environment set up with wrangler secrets as a follow up PR

olizilla: This simulates having the env passed in as magic globals

vasco-santos: yes, my point was for production, but @alanshaw already pointed me to the README with this. I didn't realize we already had it

packages/api/package.json
```diff
@@ -11,15 +11,19 @@
     "publish": "wrangler publish --env $(whoami)",
     "build": "webpack",
     "format": "prettier --write '**/*.{js,css,json,md}'",
-    "test": "playwright-test \"test/**/*.spec.js\" --sw src/index.js"
+    "test": "npm-run-all -p -r mock:api test:e2e",
+    "test:e2e": "playwright-test \"test/**/*.spec.js\" --sw src/index.js"
+    "mock:db": "smoke -p 9086 test/mocks/db",
```
olizilla, line 16: sorry!
```suggestion
    "test:e2e": "playwright-test \"test/**/*.spec.js\" --sw src/index.js",
    "mock:db": "smoke -p 9086 test/mocks/db"
```

packages/api/src/utils/fetch.js
alanshaw: nobody!

packages/api/src/user.js
alanshaw: it does

vasco-santos (approved):

packages/api/src/user.js
alanshaw: The default page size is 64. I can remove the size param to get that many...

I'll open an issue to prevent token creation abuse 😄

packages/api/src/env.js
alanshaw: Me either. @hugomrdias is there a way we can use our own build in the sw? There's no code that playwright has to add to it...right?

PR #25, opened 2021-07-06
feat: post car api

vasco-santos:
This PR adds /POST car to the API.

Needs:

- [x] docs
- [ ] DB fix for follow up chunk uploads


packages/api/src/car.js
```diff
@@ -19,44 +23,90 @@ export async function carHead(request, env, ctx) {
...
-      'Content-Disposition': `attachment; filename="${cid}.car"`,
-    }
-    // // compress if asked for? is it worth it?
-    // if (request.headers.get('Accept-Encoding').match('gzip')) {
-    //   headers['Content-Encoding'] = 'gzip'
-    // }
-    res = new Response(res.body, { ...res, headers })
-    ctx.waitUntil(cache.put(request, res.clone()))
+
+  if (res) {
```
vasco-santos, line 27: just short circuiting here to avoid extra indentation

packages/api/src/user.js
```diff
@@ -100,6 +100,7 @@ export function withAuth (handler) {
    */
   return async (request, env, ctx) => {
     const auth = request.headers.get('Authorization') || ''
+    // TODO: Should this throw if no auth token with meaningful error?
```
vasco-santos, line 103: Currently sending empty Authotization seems to throw a magic link error. Should we make this more friendly?

packages/api/src/car.js
```diff
@@ -19,44 +23,90 @@ export async function carHead(request, env, ctx) {
...
+
+  const { cid } = await cluster.add(content, {
+    // When >2.5MB, use local add, because waiting for blocks to be sent to
+    // other cluster nodes can take a long time. Replication to other nodes
+    // will be done async by bitswap instead.
+    local: blob.size > LOCAL_ADD_THRESHOLD,
+  })
+
+  // Store in DB
+  await env.db.query(gql`
```
vasco-santos, line 93: Not really sure on this query, I think I got the logic from the user queries, but 😅

packages/api/src/car.js
```diff
@@ -19,44 +23,90 @@ export async function carHead(request, env, ctx) {
...
-  console.log('Auth Token ID', authToken && authToken._id)
-  return new JSONResponse({})
+export async function carPost(request, env) {
+  const { authToken } = request.auth
+
+  const { headers } = request
+  const contentType = headers.get('content-type')
+
+  if (!contentType.includes('application/car')) {
+    return new Response(`${request.method} /car did not receive a car content type`, { status: 400 })
```
alanshaw, line 71: We don't need to enforce correct content type at this endpoint - we shouldn't be getting any other types and 'application/car' is non-standard so it won't just be added automatically for users.

packages/api/src/car.js
```diff
@@ -19,44 +23,90 @@ export async function carHead(request, env, ctx) {
...
+  }
+
+  let name = headers.get('name')
```
alanshaw, line 74:
```suggestion
  let name = headers.get('x-name')
```

packages/api/src/car.js
```diff
@@ -19,44 +23,90 @@ export async function carHead(request, env, ctx) {
...
+  if (!name || typeof name !== 'string') {
+    // throw new Error('invalid name') // Is it required?
+    name = 'testname'
```
alanshaw, line 77:
```suggestion
    name = `Upload at ${new Date().toISOString()}`
```

packages/api/src/car.js
```diff
@@ -19,44 +23,90 @@ export async function carHead(request, env, ctx) {
...
+    data: { 
+      authToken,
+      cid,
+      name
+      // dagSize: undefined // TODO: should we default to chunk car behavior?
+    }
+  })
+
+  // TODO: Improve response type with pin information
+  return new JSONResponse({ ok: true, value: { cid, name } })
```
alanshaw, line 109: I'm not sure what the benefit of the `{ ok, value }` pattern is. We end up with users having to check `Response.ok` and also `data.ok`...seems like a duplication.

Maybe @hugomrdias or @Gozala know the reasons?

packages/api/src/constants.js
```diff
@@ -1,2 +1,12 @@
...
 export const JWT_ISSUER = 'web3-storage'
+
+export const cluster = {
+  apiUrl: typeof CLUSTER_API_URL !== 'undefined'
+    ? CLUSTER_API_URL
+    : '',
+  basicAuthToken:
+    typeof CLUSTER_BASIC_AUTH_TOKEN !== 'undefined'
+      ? CLUSTER_BASIC_AUTH_TOKEN
+      : '',
```
alanshaw, line 11: We won't be able to do this when we switch to esm - can you create the cluster client in the `envAll.js`?

packages/api/src/user.js
alanshaw: We do some of this in nft.storage https://github.com/ipfs-shipyard/nft.storage/blob/aad125902fe4f8e6b5144ed52e60be3988f24e1a/packages/api/src/errors.js#L41-L68

packages/api/src/car.js
alanshaw: Did it work? You should be able to start this up in dev mode, point the JS client at it, and send a CAR.

packages/api/src/car.js
```diff
@@ -19,44 +23,90 @@ export async function carHead(request, env, ctx) {
...
+      importCar(data: $data) {
+        _id
+      }
+    }
+  `, {
+    data: { 
+      authToken,
+      cid,
+      name
+      // dagSize: undefined // TODO: should we default to chunk car behavior?
```
alanshaw, line 104: We kinda want this to be `null` if partial DAG or `blob.size` otherwise...

Not sure how to do this without reading through the CAR to see if it has any missing blocks...which I guess we _could_ do.

packages/api/test/mocks/db/post_graphql.js
```diff
@@ -8,6 +8,10 @@ const gqlOkResponse = data => gqlResponse(200, { data })
  * @param {{ body: GraphQLRequest }} request
  */
 module.exports = ({ body }) => {
+  if (body.query.includes('importCar')) {
+    return gqlOkResponse({ importCar: { _id: 'test-auth-token' } })
```
alanshaw, line 12:
```suggestion
    return gqlOkResponse({ importCar: { _id: 'test-upload-id' } })
```

packages/api/wrangler.toml
```diff
@@ -46,6 +48,7 @@ account_id = "8c3da25233263bd7a26c0e2e04569ded"
 [env.vsantos]
 workers_dev = true
 account_id = "7ec0b7cf2ec201b2580374e53ba5f37b"
+vars = { ENV = "dev", DEBUG = "*", CLUSTER_API_URL = "https://vsantos-cluster-api-web3-storage.loca.lt" }
```
alanshaw, line 51: Please can you add for everyone?

packages/api/src/car.js
alanshaw: If we did then we could default the name to `hamster.jpg` when the user uploads a CAR file of a directory with 1 file.

alanshaw (changes requested):

packages/api/src/user.js
vasco-santos: ok, so should I create an issue to get that into here?

packages/api/src/car.js
vasco-santos: Then we will get the pin status from cluster, and I think we can get the size from it if single upload?

So, I was thinking that to be a better approach than identifying if this is a partial car. wdyt?

packages/api/src/car.js
vasco-santos: Yes, this was related to my comment in your PR: https://github.com/web3-storage/web3.storage/pull/30#discussion_r664653009

packages/api/src/car.js
vasco-santos: I still need to setup magic link and fauna locally. I will do it

packages/api/src/car.js
alanshaw: I don't understand...?

vasco-santos: The next step on POST car is to get the pin status from cluster. If the cluster receives the entire CAR in this request, it will already have the size for us in its reponse.
But, I was now thinking and it might go to the request queue

vasco-santos: > If we did then we could default the name to hamster.jpg when the user uploads a CAR file of a directory with 1 file.

humm, so you mean really unpack the entire CAR file?

packages/api/wrangler.toml
```diff
@@ -34,18 +37,22 @@ route = "api-staging.web3.storage/*"
...
 
 [env.yusef]
 workers_dev = true
 account_id = "8c3da25233263bd7a26c0e2e04569ded"
+vars = { CLUSTER_API_URL = "https://yusef-cluster-api-web3-storage.loca.lt", FAUNA_ENDPOINT = "https://graphql.fauna.com/graphql" }
 
 [env.vsantos]
 workers_dev = true
 account_id = "7ec0b7cf2ec201b2580374e53ba5f37b"
+vars = { CLUSTER_API_URL = "https://vsantos-cluster-api-web3-storage.loca.lt", FAUNA_ENDPOINT = "https://graphql.fauna.com/graphql" }
```
vasco-santos, line 55: I needed to add the FAUNDA_ENDPOINT otherwise it was undefined in the worker. @alanshaw given that this is always the same, should we make a constant out of it ?

alanshaw: 👍

packages/api/src/car.js
vasco-santos: Given we know in the client if it is chunked, I think a simple and more efficient approach would be to add a Header `X-Chunked`. If it exists we do not care about the bytes returned by `cluster.add`, otherwise we set them as size in DB.
What do you think?

vasco-santos: after syncing, we will start by having the size when pinned, for simplicity and guarantee data correction.

packages/api/src/user.js
alanshaw: yes good plan

packages/api/src/car.js
alanshaw: It's a good idea, but I don't think we can rely on it being sent, or it being sent in the correct situations. We're in charge of our JS client but it's a public API and there's nothing stopping others from uploading chunks and not sending the header.

I think for now lets just not set dag size and wait for the cron job to set it.

packages/api/src/car.js
```diff
@@ -19,44 +22,86 @@ export async function carHead(request, env, ctx) {
...
+    data: { 
+      authToken: _id,
+      cid,
+      name
+      // dagSize: undefined // TODO: should we default to chunk car behavior?
+    }
+  })
+
+  // TODO: Improve response type with pin information
+  return new JSONResponse({ car: { cid, name }}, { status: 200 })
```
alanshaw, line 104: Why not just return `{ cid }` or even just the CID itself? 200 status is the default so not necessary...

```suggestion
  return new JSONResponse({ cid })
```

packages/api/src/constants.js
```diff
@@ -1,2 +1,3 @@
 export const GATEWAY = new URL('https://ipfs.io')
 export const JWT_ISSUER = 'web3-storage'
+export const FAUNA_ENDPOINT = 'https://graphql.fauna.com/graphql'
```
alanshaw, line 3:
```suggestion
```

packages/api/src/env.js
```diff
@@ -11,9 +13,13 @@ export function envAll (_, env) {
   env.magic = new Magic(env.MAGIC_SECRET_KEY || MAGIC_SECRET_KEY)
 
   env.db = new DBClient({
-    endpoint: env.FAUNA_ENDPOINT || FAUNA_ENDPOINT,
+    endpoint: env.FAUNA_ENDPOINT || FAUNA_ENDPOINT || FAUNA_DEFAULT_ENDPOINT,
```
alanshaw, line 16: This is optional, but you can't access the value of an undefined global so you need to type check it first.

```suggestion
    endpoint: env.FAUNA_ENDPOINT || (typeof FAUNA_ENDPOINT === 'undefined' ? undefined : FAUNA_ENDPOINT)
```

Screenshot 2021-07-07 at 11 50 02

packages/api/src/car.js
vasco-santos: sounds good

packages/api/src/car.js
```diff
@@ -19,44 +22,86 @@ export async function carHead(request, env, ctx) {
...
+
+  // Ensure car blob.type is set; it is used by the cluster client to set the foramt=car flag on the /add call.
+  const content = blob.slice(0, blob.size, 'application/car')
+
+  const { cid } = await env.cluster.add(content, {
+    metadata: { size: content.size.toString() },
+    // When >2.5MB, use local add, because waiting for blocks to be sent to
+    // other cluster nodes can take a long time. Replication to other nodes
+    // will be done async by bitswap instead.
+    local: blob.size > LOCAL_ADD_THRESHOLD,
```
olizilla, line 84: Why not `--local` add everything? with a 3 of 3 replication, we could get stuck waiting for content to add to all 3 nodes if 1 node is backed up

packages/client/examples/node.js/files.js
```diff
@@ -1,8 +1,11 @@
...
+import { Web3Storage, File } from '../../src/lib.js'
 
 // TODO
-const endpoint = 'https://api.web3.storage' // the default
-const token = 'API_KEY' // your API key from https://web3.storage/manage
+// const endpoint = 'https://api.web3.storage' // the default
+// const token = 'API_KEY' // your API key from https://web3.storage/manage
+
+const endpoint = 'http://127.0.0.1:8787'
+const token = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJpc3N1ZXItMTYyNTU5Mjk3MjgwNiIsImlzcyI6IndlYjMtc3RvcmFnZSIsImlhdCI6MTYyNTU5Mjk3MzIzNywibmFtZSI6InRva2VuLTE2MjU1OTI5NzI4MDYifQ.FxhcrVxOOqiy6jzyLywq6p34hzFk4V1xC_WufLd7E7s'
```
olizilla, line 8: beep boop token alert

packages/api/README.md
```diff
@@ -27,7 +43,8 @@ One time set up of your cloudflare worker subdomain for dev:
     ```sh
     wrangler secret put MAGIC_SECRET_KEY --env $(whoami) # Get from magic.link account
     wrangler secret put SALT --env $(whoami) # open `https://csprng.xyz/v1/api` in the browser and use the value of `Data`
-    wrangler secret put FAUNA_KEY --env $(whoami) # Get from fauna.com after creating a dev DB
+    wrangler secret put FAUNA_KEY --env $(whoami) # Get from fauna.com after creating a dev Classic DB
+    wrangler secret put CLUSTER_BASIC_AUTH_TOKEN --env $(whoami) # Get from web3.storage vault in 1password
```
alanshaw, line 47:
```suggestion
    wrangler secret put CLUSTER_BASIC_AUTH_TOKEN --env $(whoami) # Get from web3.storage vault in 1password (not required for dev)
```

olizilla (commented):
I need at least the `FAUNA_ENDPOINT` default change to land in main, I'd be happy with landing this PR and iterating on it in a next PR.

packages/api/src/env.js
```diff
@@ -11,9 +12,13 @@ export function envAll (_, env) {
...
-    endpoint: env.FAUNA_ENDPOINT || FAUNA_ENDPOINT,
+    endpoint: env.FAUNA_ENDPOINT || (typeof FAUNA_ENDPOINT === 'undefined' ? undefined : FAUNA_ENDPOINT),
     token: env.FAUNA_KEY || FAUNA_KEY
   })
 
   env.SALT = env.SALT || SALT
+
+  env.cluster = new Cluster(env.CLUSTER_API_URL || CLUSTER_API_URL, {
+    headers: { Authorization: `Basic ${env.CLUSTER_BASIC_AUTH_TOKEN || CLUSTER_BASIC_AUTH_TOKEN}` }
+  })
```
alanshaw, line 23:
```suggestion
  const clusterAuthToken = env.CLUSTER_BASIC_AUTH_TOKEN || (typeof CLUSTER_BASIC_AUTH_TOKEN === 'undefined' ? undefined : CLUSTER_BASIC_AUTH_TOKEN)
  const headers = clusterAuthToken ? { Authorization: `Basic ${clusterAuthToken}` } : null
  env.cluster = new Cluster(env.CLUSTER_API_URL || CLUSTER_API_URL, { headers })
```

packages/client/examples/node.js/files.js
alanshaw: woops!
```suggestion
const endpoint = 'https://api.web3.storage' // the default
const token = 'API_TOKEN' // your API token from https://web3.storage/tokens
```

packages/client/src/lib.js
```diff
@@ -26,7 +26,7 @@ import {
   Blockstore
 } from './platform.js'
 
-const MAX_PUT_RETRIES = 5
+const MAX_PUT_RETRIES = 1
```
alanshaw, line 29: Just 1?

packages/client/examples/node.js/files.js
vasco-santos: oh no, just took it out

alanshaw (changes requested):

packages/client/src/lib.js
vasco-santos: reverting, I was using this to test and accidentally put it in

packages/api/src/car.js
alanshaw: I've asked hector what happens...

alanshaw (approved):
Awesome! Lets merge this with the partial CAR makes multiple upload records as a known issue. I'll look into fixing it now.

packages/api/src/car.js
alanshaw: > During the add cluster does block-puts for every block. As long as block-put works in 1 destination it keeps going and the adding does not error.
> If block-put fails in one place once, then that place is not tried again for subsequent blocks.
> At the end of the adding process you can be sure at least 1 place has all the blocks.

...so it sounds like we don't get stuck :)

PR #29, opened 2021-07-06
fix: webpack build process

yusefnapora:
I was getting this error trying to run `npm run build` in the api package:

```bash
npm run build

> web3.storage-api@0.0.0 build
> webpack

1 asset
201 modules

ERROR in ./src/index.js 52:13-20
Module not found: Error: Can't resolve 'process/browser' in '/Users/yusef/work/repos/web3.storage/packages/api/src'
Did you mean 'browser.js'?
BREAKING CHANGE: The request 'process/browser' failed to resolve only because it was resolved as fully specified
(probably because the origin is a '*.mjs' file or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.

webpack 5.42.0 compiled with 1 error in 466 ms
```

Using node v16.4.0, npm 7.18.1 on an M1 mac.

Changing the webpack config entry to `'process/browser.js'` instead of just `'process/browser'` fixes it up on my machine.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18031,7 +18031,7 @@
         "assert": "^2.0.0",
         "buffer": "^6.0.3",
         "npm-run-all": "^4.1.5",
-        "playwright-test": "^4.1.0",
+        "playwright-test": "^5.0.0",
         "process": "^0.11.10",
         "smoke": "^3.1.1",
         "stream-browserify": "^3.0.0",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -18031,7 +18031,7 @@
         "assert": "^2.0.0",
         "buffer": "^6.0.3",
         "npm-run-all": "^4.1.5",
-        "playwright-test": "^4.1.0",
+        "playwright-test": "^5.0.0",
```
olizilla, line 18034: unexpected, but is ok, looks like it should have been updated in a previous PR.

olizilla (approved):
👍

PR #30, opened 2021-07-06
feat: user token delete

alanshaw:
* Adds the endpoint `DELETE /user/tokens/:id`
* Adds a GQL function `deleteAuthToken(user: ID!, authToken: ID!)` which sets the `deleted` flag on the token provided the token owner is the passed `user`.
* Fixes for bad codings

### DIFF
--- a/packages/api/src/user.js
+++ b/packages/api/src/user.js
@@ -159,7 +159,7 @@ export function withAuth (handler) {
 export async function userTokensPost (request, env) {
   const { name } = await request.json()
   if (!name || typeof name !== 'string') {
-    throw new Error('invalid name')
+    throw Object.assign(new Error('invalid name'), { status: 400 })
   }
 
   const { _id, issuer } = request.auth.user
@@ -167,7 +167,7 @@ export async function userTokensPost (request, env) {
   const iss = JWT_ISSUER
   const secret = await JWT.sign({ sub, iss, iat: Date.now(), name }, env.SALT)
 
-  await env.db.query(gql`
+  const res = await env.db.query(gql`
     mutation CreateAuthToken($data: CreateAuthTokenInput!) {
       createAuthToken(data: $data) {
         _id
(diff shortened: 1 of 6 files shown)


packages/api/src/user.js
```diff
@@ -159,23 +159,23 @@ export function withAuth (handler) {
...
+  const res = await env.db.query(gql`
     mutation CreateAuthToken($data: CreateAuthTokenInput!) {
       createAuthToken(data: $data) {
         _id
       }
     }
   `, { data: { user: _id, name, secret } })
 
-  return new JSONResponse()
+  return new JSONResponse(res.createAuthToken, { status: 201 })
```
vasco-santos, line 178: Is this the pattern we want? To always return the auth token?

vasco-santos (approved):
LGTM
just a small question inline

packages/api/src/user.js
```diff
@@ -198,7 +198,26 @@ export async function userTokensGet (request, env) {
...
+export async function userTokensDelete (request, env) {
+  const res = await env.db.query(gql`
+    mutation DeleteAuthToken($user: ID!, $authToken: ID!) {
+      deleteAuthToken(user: $user, authToken: $authToken) {
+        _id
+      }
+    }
+  `, { user: request.auth.user._id, authToken: request.params.id })
+
+  return new JSONResponse(res.deleteAuthToken, { status: 410 })
```
olizilla, line 222: `410 Gone` is cool, but i had to look it up, and status > 400 feels like the user did something wrong, like i tried to delete a thing that is already deleted. I think should probably just be another 200 or 204.

packages/api/src/user.js
alanshaw: Just selecting the ID from the query and returning it.

olizilla:
Are you re-running the fauna db import scripts to your local db for these PRs? What's the rough plan for handling this in staging and prod? re-import to staging db on merge to main? Import to prod db from the `release-please` workflow? The api and the db are tightly coupled... i wonder if we might actually want blue/green deploys here... but how does Fauna handle things like cloning live data... hmm, sounds messy... just thinking out loud, what are your thoughts here?

olizilla (approved):

PR #32, opened 2021-07-07
chore: install before staging deploy

olizilla:
we no longer rely on wranglers built in webpack, so we must npm install before deploying

License: (Apache-2.0 AND MIT)


olizilla:
Should fix https://github.com/web3-storage/web3.storage/runs/3007870198

PR #35, opened 2021-07-07
feat: find uploads

alanshaw:
* Adds endpoint `GET /user/uploads?before&size` - get user uploads `before` a given ISO date with the passed page `size`. Size is optional with default 25 and max 100.
* Adds FQL index for uploads by user sorted by `created` DESC
* Adds FQL function to get paginated uploads

Apologies, this PR also includes some UI changes to make auth token stuff work.


packages/api/src/user.js
```diff
@@ -221,3 +221,49 @@ export async function userTokensDelete (request, env) {
...
   return new JSONResponse(res.deleteAuthToken)
 }
+
+/**
+ * Retrieve a page of user uploads.
+ *
+ * @param {AuthenticatedRequest} request
+ * @param {import('./env').Env} env
+ */
+export async function userUploadsGet (request, env) {
```
vasco-santos, line 231: nit: better `userUploadsList`? While it is paginated, a get usually ties with an id to be fetched

packages/website/pages/tokens.js
```diff
@@ -12,7 +12,7 @@ import Loading from '../components/loading.js'
 export function getStaticProps() {
   return {
     props: {
-      title: 'Manage API Keys - NFT Storage',
+      title: 'Manage API Tokens - NFT Storage',
```
vasco-santos, line 15:
```suggestion
      title: 'Manage API Tokens - web3.storage',
```

packages/db/fauna/resources/Function/findUploads.js
```diff
@@ -0,0 +1,62 @@
...
+          )
+        ),
+        page: If(
+          Equals(Var('before'), null),
+          If(
+            Equals(Var('after'), null),
+            Paginate(Var('match'), { size: Var('size') }),
+            Paginate(Var('match'), { size: Var('size'), after: Var('after') })
+          ),
+          Paginate(Var('match'), { size: Var('size'), before: Var('before') })
```
vasco-santos, line 50: Do we only support before or after?

alanshaw: Yes, this is slightly confusing.

The top level `before` and `after` are cursor values that can also be used to paginate through the result set while retaining the `where.before` (timestamp value).

The top level `before` and `after` are added automatically when we specify `@paginate` in GraphQL. The `after` cursor gets the next page and `before` gets the previous. We have the same in `findAuthTokensByUser`.

In practice we change the `where.before` value to page through the results instead of using the cursor. I implemented it this way because we'll need to support that if we ever implement the pinning services API.

packages/api/src/user.js
alanshaw: Currently they are all suffixed with the HTTP method used so this is consistent with that.

packages/db/fauna/resources/Function/findUploads.js
alanshaw: We could rename this to `createdBefore`?

vasco-santos: I see, thanks for the details. `createdBefore` 👍🏼

vasco-santos (approved):
LGTM

packages/website/lib/api.js
```diff
@@ -93,7 +92,7 @@ export async function getFiles({ limit, before }) {
   })
 
   if (!res.ok) {
-    throw new Error(`failed to get files: ${await res.text()}`)
+    throw new Error(`failed to get uplaods: ${await res.text()}`)
```
olizilla, line 95:
```suggestion
    throw new Error(`failed to get uploads: ${await res.text()}`)
```

olizilla (approved):

PR #36, opened 2021-07-07
chore: force wrangler to deploy to custom domain

olizilla:
enforce https on api route and dont deploy to workers.dev domain

see: https://developers.cloudflare.com/workers/platform/routes#route-patterns-may-optionally-begin-with-http-or-https

License: (Apache-2.0 AND MIT)


olizilla:
I'm self merging these PRs as I can only see if they work by running them in CI on main, and they are trivial config changes (while we are not yet live)

PR #40, opened 2021-07-08
chore: update playwright

alanshaw:
This PR updates the `playwright-test` dependency.

The new version ensures the service worker is registered after the tests bundle is requested. This prevents the service worker responding to a fetch request for the bundle that contains the tests.

The worker no longer has to distinguish between requests for the worker and requests for other resources as everything should be for the worker (as it would be in production).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "web3.storage",
       "version": "1.0.0",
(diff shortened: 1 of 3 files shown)


vasco-santos (approved):

olizilla (approved):
Oh neat! good fixing.

alanshaw:
> Oh neat! good fixing.

All thanks goes to @hugomrdias

PR #42, opened 2021-07-08
fix: split CAR uploads

alanshaw:
* Creates an index of `Upload`s by `user` and `content`
* Fixes `importCar` to check for an existing `Upload` from the user for the given CID and returns that if found
    * This effectively allows split CAR uploads by not creating a new `Upload` object for every chunk of CAR that is uploaded

### DIFF
--- a/packages/api/src/car.js
+++ b/packages/api/src/car.js
@@ -63,7 +63,7 @@ export async function carGet(request, env, ctx) {
  * @param {import('./env').Env} env
  */
 export async function carPost(request, env) {
-  const { _id } = request.auth.authToken
+  const { user, authToken } = request.auth
   const { headers } = request
 
   let name = headers.get('x-name')
@@ -92,8 +92,9 @@ export async function carPost(request, env) {
       }
     }
   `, {
-    data: { 
-      authToken: _id,
+    data: {
+      user: user._id,
+      authToken: authToken?._id,
       cid,
       name
       // dagSize: undefined // TODO: should we default to chunk car behavior?
--- a/packages/api/src/env.js
+++ b/packages/api/src/env.js
@@ -20,6 +20,6 @@ export function envAll (_, env) {
   env.SALT = env.SALT || SALT
 
   const clusterAuthToken = env.CLUSTER_BASIC_AUTH_TOKEN || (typeof CLUSTER_BASIC_AUTH_TOKEN === 'undefined' ? undefined : CLUSTER_BASIC_AUTH_TOKEN)
-  const headers = clusterAuthToken ? { Authorization: `Basic ${clusterAuthToken}` } : null
+  const headers = clusterAuthToken ? { Authorization: `Basic ${clusterAuthToken}` } : {}
   env.cluster = new Cluster(env.CLUSTER_API_URL || CLUSTER_API_URL, { headers })
 }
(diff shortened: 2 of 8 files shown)


packages/db/fauna/resources/Function/findUploads.js
```diff
@@ -31,7 +31,7 @@ const body = Query(
       {
         match: Filter(
           Match(
-            Index('upload_sort_by_created_desc'),
+            Index('upload_by_user_sort_by_created_desc'),
```
alanshaw, line 34: Just giving this a better name...not related to the purpose of the PR - sorry.

packages/api/src/car.js
```diff
@@ -92,8 +92,9 @@ export async function carPost(request, env) {
       }
     }
   `, {
-    data: { 
-      authToken: _id,
+    data: {
+      user: user._id,
+      authToken: authToken?._id,
```
alanshaw, line 97: `authToken` may be null here if the user is using a Magic.link token.

packages/api/src/env.js
```diff
@@ -20,6 +20,6 @@ export function envAll (_, env) {
   env.SALT = env.SALT || SALT
 
   const clusterAuthToken = env.CLUSTER_BASIC_AUTH_TOKEN || (typeof CLUSTER_BASIC_AUTH_TOKEN === 'undefined' ? undefined : CLUSTER_BASIC_AUTH_TOKEN)
-  const headers = clusterAuthToken ? { Authorization: `Basic ${clusterAuthToken}` } : null
+  const headers = clusterAuthToken ? { Authorization: `Basic ${clusterAuthToken}` } : {}
```
alanshaw, line 23: Cloudflare workers was _not_ happy with `null` for headers FML:

```console
TypeError: Incorrect type for the 'headers' field on 'RequestInitializerDict': the provided value is not of type 'variant'.
    at Cluster.add (worker.js:875:28)
    at carPost (worker.js:37751:37)
```

packages/db/fauna/schema.graphql
```diff
@@ -45,12 +45,12 @@ type AuthToken {
   """
   Uploads made using this token.
   """
-  uploads: [Upload!]! @relation @index
+  uploads: [Upload!]! @relation
```
alanshaw, line 48: `@index` with no name is a noop - this doesn't actually tell Fauna to create an index on this field, just allows you to name the index to use.

https://docs.fauna.com/fauna/current/api/graphql/directives/d_index

packages/db/fauna/resources/Function/importCar.js
```diff
@@ -28,47 +29,71 @@ const body = Query(
...
-              cid: Var('cid'),
-              content: Select('ref', Get(Var('contentMatch'))),
-              created: Now(),
-              name: Select('name', Var('data'), null)
-            }
-          }),
-          Let(
-            {
-              content: Create('Content', {
+        Exists(Var('userRef')),
```
olizilla, line 40: pls to abort early

packages/db/fauna/resources/Index/uploadByUserAndContent.js
```diff
@@ -0,0 +1,52 @@
...
+} = fauna
+
+/**
+ * Usage:
+ *
+ * Match(
+ *   Index('upload_by_user_and_content'),
+ *   Ref(Collection('user'), Var('id')),
+ *   Ref(Collection('content'), Var('id')),
+ *   true // not_deleted
```
olizilla, line 24: would be lovely to not have to remember to pass an unnamed boolean to ensure you dont count deleted things which is mostly what you want

olizilla (approved):

PR #48, opened 2021-07-09
feat: delete user upload

vasco-santos:
Adds DELETE /user/uploads/:id to API and wires it up in UI

Given we are going with DELETE /user/uploads/:id I didn't add a delete to the client, as this is not a /car route. Happy to add it as well

All flow tested with local setup from website

### DIFF
--- a/packages/website/lib/api.js
+++ b/packages/website/lib/api.js
@@ -97,3 +97,22 @@ export async function getUploads({ size, before }) {
 
   return res.json()
 }
+
+/**
+ * Deletes upload
+ *
+ * @param {string} _id
+ */
+export async function deleteUpload (_id) {
+  const res = await fetch(`${API}/user/uploads/${_id}`, {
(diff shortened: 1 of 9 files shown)


packages/website/lib/api.js
```diff
@@ -97,3 +97,22 @@ export async function getUploads({ size, before }) {
...
+  const res = await fetch(`${API}/user/uploads/${_id}`, {
+    method: 'DELETE',
+    headers: {
+      'Content-Type': 'application/json',
+      Authorization: 'Bearer ' + (await getToken()),
+    }
+  })
+
+  if (!res.ok) {
+    throw new Error(`failed to delete auth token: ${await res.text()}`)
```
alanshaw, line 116:
```suggestion
    throw new Error(`failed to delete upload: ${await res.text()}`)
```

alanshaw (approved):

PR #51, opened 2021-07-09
feat: storage broker API

alanshaw:
This PR makes changes to allow the storage broker service to interact with the site via GraphQL.

It involves splitting the `Deal` collection out into `BatchEntry`, `Batch` and `Deal` to more closely follow the [schema](https://raw.githubusercontent.com/nftstorage/dagcargo/master/maint/pg_schema_diagram.svg) in the storage broker.

`Content` _will_ be included in multiple batches. This may happen during initial deal creation if the same content is referenced by multiple DAGs, when deals are "renewed" or if we decide to change the format of the batch DAG (and want to "re-do" a bunch of deals in the new format).

`BatchEntry` is _not_ denormalized into `Content`, just incase we are ever given a batch root CID and want to easily know what entries it contains. It also ensures we don't hold write locks on thousands of `Content` records when creating a batch.

**Find new CIDs:**

```gql
query {
  findUploadsCreatedAfter(since: "2021-07-08T15:13:08.318949Z" _cursor: null _size: 1000) {
    data {
      content {
        cid
      }
      user {
        issuer
      }
      created
    }
    # cursor value to be used as _cursor for the next page
    after
  }
}
```

**Create a batch:**

```gql
mutation {
  createBatch(data: {
    batchCid: "BATCHROOTCID"
    pieceCid: "PIECECID"
    entries: [{
      cid: "ENTRYCID"
      dataModelSelector: "Links/0/Links/1/..."
    }]
  }) {
    _id
    cid
    pieceCid
  }
}
```

**Create or update a deal:**

```gql
mutation {
  createOrUpdateDeal(data: {
    batchCid: "BATCHROOTCID" # required (for create, ignored for update)
    miner: "f99"
    network: "mainnet"
    chainDealId: 12345 # required
    activation: "2021-07-08T15:13:08.318949Z"
    renewal: "2022-07-08T15:13:08.318949Z" # (expiry date)
    status: "Active" # required
    statusReason: "It worked!"
  }) {
    _id
    batch {
      cid
    }
    status
  }
}
```

**Updated schema:**

web3 storage schema(3)


packages/db/fauna/resources/Function/createBatch.js
```diff
@@ -0,0 +1,81 @@
...
+          Collection('Batch'),
+          {
+            data: {
+              cid: Var('batchCid'),
+              pieceCid: Var('pieceCid')
+            }
+          }
+        ),
+        batchRef: Select('ref', Var('batch')),
+        entries: Select('entries', Var('data'))
```
ribasushi, line 43: You can/should make all of these mandatory fields to catch errors on my side.

packages/db/fauna/resources/Function/createBatch.js
```diff
@@ -0,0 +1,81 @@
...
+} = fauna
+
+const name = 'createBatch'
+const body = Query(
+  Lambda(
+    ['data'],
+    Let(
+      {
+        batchCid: Select('batchCid', Var('data')),
+        pieceCid: Select('pieceCid', Var('data')),
```
ribasushi, line 32: When I call this API I know the exact byte-size of the resulting car (I can't calculate the PieceCID without the file - so I obviously know its size). You could store that info if of interest.

ribasushi (approved):
Looks great! (api-surface wise, I can't comment competently on the code)

Added two nits, ok to go without any changes.

packages/db/fauna/resources/Function/createOrUpdateDeal.js
```diff
@@ -0,0 +1,95 @@
...
+        ),
+        Let(
+          {
+            deal: Get(Var('dealMatch'))
+          },
+          Update(
+            Select('ref', Var('deal')),
+            {
+              data: {
+                miner: Select('miner', Var('data'), Select(['data', 'miner'], Var('deal'), null)),
```
vasco-santos, line 75: nit: better readability without Selects inside selects? Probably better to declare in Let

vasco-santos (approved):
LGTM! This is great 🎉

packages/db/fauna/resources/Function/createBatch.js
alanshaw: `Select` will throw if the property doesn't exist in the object and no default is given (like here).

alanshaw: Screenshot 2021-07-12 at 14 33 02

They are also enforced required by GraphQL (the `!`)

ribasushi: Ah got it, all good!

packages/db/fauna/schema.graphql
```diff
@@ -195,43 +196,74 @@ type Upload {
...
   """
-  contents: [Content!]! @relation
+  cid: String! @unique
 
   """
-  ID of miner this deal was made with.
+  Piece CID.
   """
-  miner: String
+  pieceCid: String
```
olizilla, line 210: CreateBatchInput has a non-nullable pieceCID, so this could be non-nullable too, if we care.
```suggestion
  pieceCid: String!
```

packages/db/fauna/schema.graphql
```diff
@@ -195,43 +196,74 @@ type Upload {
...
+
+  """
+  ID of miner this deal was made with.
+  """
+  miner: String
+
+  """
+  Filecoin network deal is made on.
+  """
+  network: String
```
olizilla, line 261: why nullable `miner` and `network`?

packages/db/fauna/schema.graphql
```diff
@@ -195,43 +196,74 @@ type Upload {
...
+
+  """
+  Identifier for the deal stored on chain.
+  """
+  chainDealId: Int @unique
```
olizilla, line 266: does @unique imply not-nullable? i think we want chainDealId: Int!
```suggestion
  chainDealId: Int! @unique
```

olizilla: of note, not changed in this PR, but `activation` and `renewal` and `statusReason` are all still nullable as well.

packages/db/fauna/schema.graphql
```diff
@@ -195,43 +196,74 @@ type Upload {
 }
 
 """
-Filecoin deal for a pin.
+A batch of content for inclusion in a Filecoin deal.
 """
-type Deal {
+type Batch {
```
olizilla, line 201: We could track when a Batch is created so we can periodically delete batches that got dumped in favour of others and will never make it on chain.

packages/db/fauna/schema.graphql
```diff
@@ -286,10 +321,37 @@ type Mutation {
...
+
+input CreateBatchInput {
+  batchCid: String!
+  pieceCid: String!
+  entries: [BatchEntryInput!]!
+}
+
+input BatchEntryInput {
+  cid: String!
+  dataModelSelector: String
```
olizilla, line 336: is `dataModelSelector` optional?

olizilla (changes requested):
Is very good. Minor schema tweaks suggested.

packages/db/fauna/schema.graphql
ribasushi: Note - I do not intend to send dealinfo up until there's an actual on-chain publish (prior to seal). But still worthwhile to do for the future.

packages/db/fauna/schema.graphql
ribasushi: Yeah, `miner` should be required. We should just dump `network` - testnetworks are too much faff for little benefit at this stage.

packages/db/fauna/resources/Function/findUploadsCreatedAfter.js
```diff
@@ -0,0 +1,52 @@
...
+        ),
+        page: If(
+          Equals(Var('before'), null),
+          If(
+            Equals(Var('after'), null),
+            Paginate(Var('range'), { size: Var('size') }),
+            Paginate(Var('range'), { size: Var('size'), after: Var('after') })
+          ),
+          Paginate(Var('range'), { size: Var('size'), before: Var('before') })
+        )
```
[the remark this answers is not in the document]
alanshaw: I have opened an issue here https://github.com/web3-storage/web3.storage/issues/75

olizilla (approved):

PR #57, opened 2021-07-09
feat: w3 --put /path/to/files

olizilla:
adds a `--put` command to `w3`.

```console
$ w3 --put pics --token <your token here... this is temporary>

+ /pics/dr-is-tired.jpg
+ /pics/not-distributed.jpg
+ /pics/youareanonsense.jpg
⁂ https://dweb.link/ipfs/bafybeidd2gyhagleh47qeg77xqndy2qy3yzn4vkxmk775bg2t5lpuy7pcu
```

glob-source is ported from ipfs-utils to be esm and make files in the shape that client.put expects them.

License: (Apache-2.0 AND MIT)

### DIFF
--- a/packages/w3/bin.js
+++ b/packages/w3/bin.js
@@ -12,7 +12,7 @@ const cli = meow(
     flags: {
       api: {
         type: 'string',
-        default: 'https://api-staging.web3.storage'
+        default: 'https://api.web3.storage'
       },
       token: {
         type: 'string',
--- a/packages/w3/bin.js
+++ b/packages/w3/bin.js
@@ -5,15 +5,22 @@ import w3 from './index.js'
 const cli = meow(
   `Usage
     --get bafybeidd2gyhagleh47qeg77xqndy2qy3yzn4vkxmk775bg2t5lpuy7pcu --output pics --api http://127.0.0.1:8787
+    --put /path/to/files
   `,
   {
     importMeta: import.meta,
     flags: {
       api: {
         type: 'string',
-        default: 'https://api.web3.storage'
+        default: 'https://api-staging.web3.storage'
+      },
+      token: {
+        type: 'string',
       },
       get: {
+        type: 'string',
+      },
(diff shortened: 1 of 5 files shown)


packages/w3/bin.js
```diff
@@ -5,17 +5,24 @@ import w3 from './index.js'
...
     --get bafybeidd2gyhagleh47qeg77xqndy2qy3yzn4vkxmk775bg2t5lpuy7pcu --output pics --api http://127.0.0.1:8787
+    --put /path/to/files
   `,
   {
     importMeta: import.meta,
     flags: {
       api: {
         type: 'string',
-        default: 'https://api.web3.storage'
+        default: 'https://api-staging.web3.storage'
```
alanshaw, line 15:
```suggestion
        default: 'https://api.web3.storage'
```

packages/w3/bin.js
```diff
@@ -5,17 +5,24 @@ import w3 from './index.js'
 const cli = meow(
   `Usage
     --get bafybeidd2gyhagleh47qeg77xqndy2qy3yzn4vkxmk775bg2t5lpuy7pcu --output pics --api http://127.0.0.1:8787
+    --put /path/to/files
```
alanshaw, line 8: Not for this PR but I think it would be nice if these were subcommands not flags...i.e. `w3 get CID` and `w3 put /file`.

The [carbites CLI](https://github.com/nftstorage/carbites-cli/) does this with meow.

alanshaw (approved):

packages/w3/bin.js
olizilla: agree! i want commands too. I was thinking of using a lib other than meow that was more geared towards clis with sub commands. https://github.com/lukeed/sade looks nice, as used by the admin tooling for nft.storage https://github.com/ipfs-shipyard/nft.storage/blob/main/packages/tools/cli.js

PR #60, opened 2021-07-10
feat: subcommands for w3

olizilla:
- `w3 get <cid>` fetch files by cid and verify them
- `w3 put <path>` upload files to web3.storage
- switches `meow` for `sade` for the subcommand support.

**upload in progress**

```console
$ w3 put ~/Desktop/output/week-in-web3-2021-07-02.mov
# Packed 1 file (181.4MB)
# bafybeigzwx5ychvapyslpmgy7hwdwh6nabyapxkjmpmtwwqevv4zpwfauy
⠧ Storing 48%
```

**success ✨**

```console
$ w3 put ~/Desktop/output/week-in-web3-2021-07-02.mov
# Packed 1 file (181.4MB)
# bafybeigzwx5ychvapyslpmgy7hwdwh6nabyapxkjmpmtwwqevv4zpwfauy
⁂ Stored 1 file
⁂ https://dweb.link/ipfs/bafybeigzwx5ychvapyslpmgy7hwdwh6nabyapxkjmpmtwwqevv4zpwfauy
```

License: (Apache-2.0 AND MIT)

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,8 @@
     "format": "prettier --write '**/*.{js,jsx,ts,tsx}'"
   },
   "devDependencies": {
-    "prettier": "^2.3.1"
+    "prettier": "^2.3.1",
+    "standard": "^16.0.3"
   },
   "prettier": {
     "trailingComma": "es5",
--- a/package.json
+++ b/package.json
@@ -15,8 +15,7 @@
     "format": "prettier --write '**/*.{js,jsx,ts,tsx}'"
   },
   "devDependencies": {
-    "prettier": "^2.3.1",
-    "standard": "^16.0.3"
+    "prettier": "^2.3.1"
   },
   "prettier": {
     "trailingComma": "es5",
--- a/packages/w3/bin.js
+++ b/packages/w3/bin.js
@@ -12,13 +12,13 @@ cli
   .example('get bafkreigh2akiscaildcqabsyg3dfr6chu3fgpregiymsck7e7aqa4s52zy -o room-guardian.jpg')
 
 cli.command('get <cid>')
-  .describe('Get and verify files from web3.storage')
+  .describe('Fetch and verify files from web3.storage')
   .option('-o, --output', 'Write to the output file name')
   .example('get bafkreigh2akiscaildcqabsyg3dfr6chu3fgpregiymsck7e7aqa4s52zy -o room-guardian.jpg')
   .action(get)
 
-cli.command('put <path...>')
-  .describe('Upload files to web3.storage')
+cli.command('put <path>')
+  .describe('Upload a file or directory to web3.storage')
   .action(put)
 
 cli.parse(process.argv)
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -15,7 +15,8 @@
     "format": "prettier --write '**/*.{js,jsx,ts,tsx}'"
   },
   "devDependencies": {
-    "prettier": "^2.3.1"
+    "prettier": "^2.3.1",
+    "standard": "^16.0.3"
```
alanshaw, line 19: These two are not compatible. We should choose one. My preference is standard, because it lints as well as formats and there's plugins that can show warnings in your editor as they occur.

packages/w3/bin.js
```diff
@@ -1,41 +1,24 @@
...
-}
-if (cli.flags.get && cli.flags.put) {
-  cli.showHelp()
-}
+const cli = sade('w3')
 
-w3(cli)
+cli
+  .option('--api', 'URL for API to use')
+  .option('--token', 'API token to use')
```
alanshaw, line 10: Not blocking, but we should have something like `w3 config set token <AUTH_TOKEN>` that writes to a `.w3rc` file in your home directory.

packages/w3/bin.js
```diff
@@ -1,41 +1,24 @@
...
+  .example('get bafkreigh2akiscaildcqabsyg3dfr6chu3fgpregiymsck7e7aqa4s52zy -o room-guardian.jpg')
+
+cli.command('get <cid>')
+  .describe('Get and verify files from web3.storage')
+  .option('-o, --output', 'Write to the output file name')
+  .example('get bafkreigh2akiscaildcqabsyg3dfr6chu3fgpregiymsck7e7aqa4s52zy -o room-guardian.jpg')
+  .action(get)
+
+cli.command('put <path>')
+  .describe('Upload files to web3.storage')
```
alanshaw, line 21:
```suggestion
  .describe('Upload a file or directory to web3.storage')
```

alanshaw (approved):

packages/w3/bin.js
olizilla: Agree! The plan is to use [`conf`](https://github.com/sindresorhus/conf)  which recommends storing things in

>    macOS: ~/Library/Preferences/MyApp-nodejs
>    Windows: %APPDATA%\MyApp-nodejs\Config (for example, C:\Users\USERNAME\AppData\Roaming\MyApp-nodejs\Config)
>    Linux: ~/.config/MyApp-nodejs (or $XDG_CONFIG_HOME/MyApp-nodejs)

and the command to set it would be `w3 token <token>` for fewer words.

For CI world, we could have it check a .w3 file in the cwd too.

package.json
olizilla: i'm gonna revert this change for this PR, so we can land the linting change as a seperate thing.

olizilla: Done

PR #62, opened 2021-07-12
fix!: cid prop on web3file is a string

olizilla:
Previously calling `res.files()` on the response from `client.get` would return an array of Files with a `.cid` property that was a CID instance. This PR makes it a string, to be consistent with the our other usages of CID which are all strings. If folks need to upgrade their string CIDs to instances of CID then the `CID.parse(str)` method is the one to use.

A breaking change for the client API! 

License: (Apache-2.0 AND MIT)


alanshaw (approved):

vasco-santos (approved):
LGTM

PR #66, opened 2021-07-12
chore: deploy staging db from ci

olizilla:
- add db workflow that runs the fauna resource import script when a change is merged to main
- `STAGING_FAUNA_KEY` is added to repo.

License: (Apache-2.0 AND MIT)

### DIFF
--- a/.github/workflows/db.yml
+++ b/.github/workflows/db.yml
@@ -17,7 +17,6 @@ jobs:
     name: Deploy Staging DB
     # if: github.event_name == 'push' && github.ref == 'refs/heads/main'
     runs-on: ubuntu-latest
-    needs: test
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
--- a/.github/workflows/db.yml
+++ b/.github/workflows/db.yml
@@ -24,7 +24,7 @@ jobs:
         with:
           node-version: 16
       - uses: bahmutov/npm-install@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/db.yml
```diff
@@ -0,0 +1,30 @@
...
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+    runs-on: ubuntu-latest
+    needs: test
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 16
+      - uses: bahmutov/npm-install@v1
+      - name: Update FuanaDB resources
```
olizilla, line 27:
```suggestion
      - name: Update FaunaDB resources
```

olizilla:
When this PR is merged, the `import` task in the db packge should be re-run against the staging db. It's idempotent, so that should have no impact on the staging site, but it is worth calling it out that this idempotency is now required, and changes should be written so they can be deployed on top of the previous schema. In time we will no doubt find issues where a schema change is awkward to deploy without some migration. At that point, this auto-deploy to the staging db may fail, and this should be treated as an alarm bell that the same problem will occur when deploying to prod, and should be resolved before proceeding.

olizilla:
i'm going to remove the branch guard from this PR so i can see it run before letting it loose.

alanshaw (approved):

PR #68, opened 2021-07-12
feat: add home, login and about pages

jdiogopeixoto:
- Changes the website from `tachyons` to `tailwind` ✨ 
- Adds the homepage structure with some CSS
- Changes the login page according to the new design
- Adds global typography configuration - so we don't have to worry about it every time
- Adds "Space Grotesk" and "Helvetica" fonts
- Adds the about page


packages/website/styles/global.css
```diff
@@ -1,12 +1,114 @@
...
+    font-style: normal;
+    src: url("/fonts/space-grotesk/space-grotesk-medium.woff2") format("woff2");
+    src: url("/fonts/space-grotesk/space-grotesk-medium.woff") format("woff");
+}
+
+@font-face {
+    font-family: "Helvetica";
+    font-weight: 400;
+    font-style: normal;
+    src: url("/fonts/helvetica/helvetica.woff2") format("woff2");
```
rafaelramalho19, line 29: I would change this approach from using self-hosted font files to just using google fonts.

Pros:
- Easier to cache through websites (let's say the user already browsed a website with Helvetica in google fonts, they would get that cached in the browser and it speeds up the first load ever)
- Removes font bloatware, which you currently have with 8 fonts.

Cons:
- The only con is that it adds an extra domain to fetch from the initial load, but with current gen techonologies, adding pre-load links to google fonts easily fixes it.

packages/website/components/button.js
```diff
@@ -28,34 +28,23 @@ export default function Button({
...
-        'chicagoflf',
-        'f5',
+        'w-full',
+        'border',
+        'border-black',
+        'rounded-md',
+        'px-4',
         { pointer: !disabled },
-        'bg-white',
+        'bg-black',
```
rafaelramalho19, line 42: The button doesn't have a hover effect now, you can add:
```css
hover:bg-gray-700
```

packages/website/styles/global.css
olizilla: The website should be self contained, so we can add it to IPFS and host it entirely from the root cid. In general we want to limit our dependence on third party domains for all our sites.

olizilla: AKA: please let's self host the fonts, unless there is a very strong reason not to.

rafaelramalho19: That's a strong point @olizilla, go fo self-host then 👍

packages/website/components/button.js
jdiogopeixoto: I would say it's not worth adding it for now since the design is not finished yet. I will also need to add a light variant to the button, so it's better to deal with the hover states later on.

alanshaw (approved):
LGTM!

PR #69, opened 2021-07-13
chore: add standard linting

olizilla:
Adds snazzy standard js linting and removes all our accumulated lint, and checks for lint in CI. We tried out annotating the PRs with lint errors, but it is too noisy, and makes PRs hard to review. As it is there will be a polite check error which links to the list of lint errors.

Adds:
`npm run lint` to the root that checks standard and gives snazzy output
`npm run lint:fix` adds the fix flag if you forget how many `--` you need to add to pass flags to npm run tasks.

```console
$ npm run lint

> web3.storage@1.0.0 lint
> standard --verbose | snazzy

standard: Use JavaScript Standard Style (https://standardjs.com)
standard: Run `standard --fix` to automatically fix some problems.

/Users/oli/Code/web3-storage/web3.storage/packages/api/pw-test.config.cjs
  18:2  error  Newline required at end of file but not found  eol-last

/Users/oli/Code/web3-storage/web3.storage/packages/db/fauna/resources/Function/deleteUserUpload.js
  28:62  error  Unexpected trailing comma  comma-dangle

✖ 2 problems
```

Worstcase example output for a big bad run is here: https://github.com/web3-storage/web3.storage/pull/69/checks?check_run_id=3055923078

The `website` package is ignored for now, to avoid lint based rebase pain, and it also has more opinionated lint/formatting config already.

License: (Apache-2.0 AND MIT)


alanshaw (approved):
🎉

packages/client/src/lib.js
```diff
@@ -30,8 +29,17 @@ const MAX_PUT_RETRIES = 5
...
 const MAX_CHUNK_SIZE = 1024 * 1024 * 10 // chunk to ~10MB CARs
 
+/** @typedef { import('./lib/interface.js').API } API */
+/** @typedef { import('./lib/interface.js').Service } Service */
+/** @typedef { import('./lib/interface.js').Filelike } Filelike */
+/** @typedef { import('./lib/interface.js').CIDString} CIDString */
+/** @typedef { import('./lib/interface.js').Web3File} Web3File */
+/** @typedef { import('./lib/interface.js').PutOptions} PutOptions */
+/** @typedef { import('./lib/interface.js').UnixFSEntry} UnixFSEntry */
+/** @typedef { import('./lib/interface.js').Web3Response} Web3Response */
```
olizilla, line 39: @alanshaw pls to note. The type importing in the client lib has be converted to many simple typedefs, as I think it makes the rest of the JSDoc types through the file more readable.

I was obliged to stop importing the interface as a module as standard did not approve of it, but then had to fix all the tsc warnings and it was not happy with trying to use a namespace like API.Service, so this seemed the least worst, least magic solution to appease both. I could litter the JSDoc comments with these imports, but it felt nicer to have this block that we can ignore, for more readable method comments.

PR #70, opened 2021-07-13
feat: get metrics

vasco-santos:
Adds metrics endpoint to API and created DB needed query and indexes.

The current metrics being computed were inspired from the ones exposed in nft.storage. They seemed reasonable to me at this point.

A preview: 

image


packages/db/fauna/resources/Index/pinsSizes.js
```diff
@@ -0,0 +1,38 @@
...
+const index = {
+  name: 'pins_sizes',
+  source: [
+    Collection('Pin')
+  ],
+  terms: [
+    { field: ['data', 'status'] }
+  ],
+  values: [
+    { field: ['data', 'content', 'data', 'dagSize'] }
```
alanshaw, line 29: I think this is not working.

vasco-santos: you are right! Before I added the status condition it seemed to be working. I will look into it

alanshaw: I think you can't traverse a relation.

vasco-santos: I think this was working for a weird combination of merging the index updates in Fauna as I know cannot make it work lol

vasco-santos: Yes, I am working on a new index/query for this

vasco-santos: image
Should now be fixed. I changed the index logic to get all the pins content ref and then sum their dagSize. Not sure if this is the most efficient though

packages/db/fauna/resources/Function/findMetrics.js
```diff
@@ -0,0 +1,82 @@
...
+          'Active'
+        )),
+        dealsQueuedTotal: Count(Match(
+          Index("deals_by_status"),
+          'Queued'
+        )),
+        pinsTotal: Count(Documents(Collection("Pin"))),
+        pinsTotalBytes: Sum(
+          Select('data', Map(
+            Paginate(
```
vasco-santos, line 45: Using this to convert set to array, couldn't find an alternative 🤷🏼

vasco-santos:
rebased PR and fixed linting

olizilla:
plz to resolve conflicts! (sorry)

vasco-santos:
> plz to resolve conflicts! (sorry)

@olizilla rebased 🌲

packages/api/src/metrics.js
```diff
@@ -0,0 +1,74 @@
...
+
+      '# HELP web3storage_uploads_total Total number of uploads.',
+      '# TYPE web3storage_uploads_total counter',
+      `web3storage_uploads_total ${findMetrics.uploadsTotal}`,
+
+      '# HELP web3storage_content_bytes_total Total bytes of all web3s.',
+      '# TYPE web3storage_content_bytes_total counter',
+      `web3storage_content_bytes_total ${findMetrics.contentTotalBytes}`,
+
+      '# HELP web3storage_content_filecoin_total Total number of content stored on Filecoin in active deals.',
```
olizilla, line 41:
```suggestion
      '# HELP web3storage_content_filecoin_total Total number of unique DAGs in active Filecoin deals.',
```

packages/api/src/metrics.js
```diff
@@ -0,0 +1,74 @@
...
+      '# TYPE web3storage_content_filecoin_total counter',
+      `web3storage_content_filecoin_total ${findMetrics.dealsActiveTotal}`,
+
+      '# HELP web3storage_content_filecoin_queued_total Total number of content queued for the next deal batch.',
```
olizilla, line 45:
```suggestion
      '# HELP web3storage_content_filecoin_queued_total Total number of unique DAGs queued for a Filecoin deal',
```

packages/api/src/metrics.js
```diff
@@ -0,0 +1,74 @@
...
+      '# TYPE web3storage_content_filecoin_queued_total counter',
+      `web3storage_content_filecoin_queued_total ${findMetrics.dealsQueuedTotal}`,
+
+      '# HELP web3storage_pins_total Total number of pins on IPFS.',
```
olizilla, line 49:
```suggestion
      '# HELP web3storage_pins_total Total number of pins on the IPFS Cluster',
```

packages/api/src/metrics.js
```diff
@@ -0,0 +1,74 @@
...
+      '# TYPE web3storage_pins_total counter',
+      `web3storage_pins_total ${findMetrics.pinsTotal}`,
+
+      '# HELP web3storage_pins_bytes_total Total size of pinned items on IPFS.',
```
olizilla, line 53:
```suggestion
      '# HELP web3storage_pins_bytes_total Total size in bytes of pins on the IPFS Cluster',
```

packages/api/test/fixtures/find-metrics.json
```diff
@@ -0,0 +1,15 @@
+{
```
olizilla, line 1: I wasted an hour whilst tired last night trying to figure out why my code worked in test and not on live and it was due to the mismatch of the response shape that you get from the graphql explorer / curl against fuana having a top level `data` property and our test code and db client hiding that. Could we make them all the same please? like have the fixtures match the response shape you get from the db, match the reponse shape you get from the db client?

cc @alanshaw

packages/api/test/metrics.spec.js
```diff
@@ -0,0 +1,23 @@
+/* eslint-env mocha, browser */
+import assert from 'assert'
+import { endpoint } from './scripts/constants.js'
+
+describe('GET /metrics', () => {
+  it('retrieves metrics', async () => {
+    const res = await fetch(new URL('metrics/', endpoint).toString())
```
olizilla, line 7:
```suggestion
    const res = await fetch(new URL('metrics/', endpoint))
```

> const fetchResponsePromise = fetch(resource [, init])
> Parameters
> `resource`
>     This defines the resource that you wish to fetch. This can either be: A string or any other object with a stringifier — including a URL object — that provides the URL of the resource you want to fetch.
– https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch

packages/db/fauna/resources/Function/findMetrics.js
```diff
@@ -0,0 +1,82 @@
...
+        dealsQueuedTotal: Count(Match(
+          Index('deals_by_status'),
+          'Queued'
+        )),
+        pinsTotal: Count(Documents(Collection('Pin'))),
+        pinsTotalBytes: Sum(
+          Select('data', Map(
+            Paginate(
+              Match(
+                Index('pins_content_by_status'),
```
olizilla, line 47: Hmm, it feels a bit like `dagSize` should be a property of `Pin` as we really don't know the dag size other than when cluster Pins it.

re #83 

cc @alanshaw

packages/api/test/mocks/db/post_graphql.js
```diff
@@ -63,6 +63,10 @@ module.exports = ({ body }) => {
     return gqlOkResponse(require('../../fixtures/find-content-by-cid.json'))
   }
 
+  if (body.query.includes('findMetrics')) {
+    return gqlOkResponse(require('../../fixtures/find-metrics.json'))
```
olizilla, line 67: yeah, this is the part that bit me... my fixture was a copy/paste of the graphql response, and our `gqlOkResponse` adds the `data` prop for you, but then the db client removes that prop from the response that it returns to your code. I think i'd prefer the redundant `.data` prop to just appear everywhere, so fixtures match what you get from the db, and you can look at the fixture and write your api code against it.

packages/api/test/metrics.spec.js
vasco-santos: 👌🏼  There are other tests with the same pattern here, I will also update them

packages/db/fauna/resources/Function/findMetrics.js
vasco-santos: It would ease things, but duplicate data, as it should also be in the content

packages/api/test/fixtures/find-metrics.json
vasco-santos: If we do not return with `data`, the query simply does not return anything (this was the issue I was having yesterday).

```
const body = Query(
  Lambda(
    [],
    {
      data: {
        usersTotal: Count(Documents(Collection('User')))
        // ....
```

as opposed to:

```
const body = Query(
  Lambda(
    [],
    {
      usersTotal: Count(Documents(Collection('User')))
      // ....
```

With this in mind, I agree that we should return the same format always, but probably we will need to rely on data

olizilla: ok... a proposal
- the fixtures should include the `data` property. A fixture should match exactly what the db returns
- i'm going to PR an update to the api code where we call the db client to call response object `data` rather than `res`, so it's easier to guess that it has given you the `data` prop of the response rather than the full response with the `data` prop contained with in it.... which is what graphql-request does in all it's examples https://github.com/prisma-labs/graphql-request#quickstart

vasco-santos: I am a bit confused, we return with `data` in fauna but then the response that we get in the client has no data. I am receiving in the client the response just with `findMetrics`

vasco-santos: apparently only the status DB query returns data?

vasco-santos: https://github.com/web3-storage/web3.storage/blob/main/packages/db/fauna/resources/Function/findContentByCid.js#L25

I think what we want is to simply add a select to get the data from the Match here

olizilla: oh my, that could be the source of my confusion... is it a quirk of the `findContentByCid` function? why would the db client be stripping that part of the object???

vasco-santos: A [Match](https://docs.fauna.com/fauna/current/api/fql/functions/match?lang=shell#returns) returns a SetRef with data and then [Get](https://docs.fauna.com/fauna/current/api/fql/functions/get?lang=shell#returns) returns the first reference with data / ref too.

vasco-santos: Apparently `findUserByIssuer` also returns in the same format, but it is not being mocked. There seem to be the only two cases

alanshaw: This is to do with returning a `Document` from FQL. A document is `{ ref, data, ttl, ... }`. When you return a _document_ from FQL, the value of `data` is what is returned in your GraphQL response.

This code is making a regular object _look_ like a document so that this happens - as vasco said, nothing is returned otherwise.

Note that `findContentByCid` is using `Get` which returns a `Document`.

---

Separately, when we use `Paginate` in FQL we return an array of documents. The directive `@resolver(paginated: true)` in GraphQL does a magic transform on that array of documents to turn it into `{ data, after, before }`, where `data` is the array of documents and `after`/`before` are auto generated cursor values to allow you to get the next/previous page of data.

e.g.

```gql
findUploadsCreatedAfter(since: Time!): [Upload!]! @resolver(paginated: true)
```

```gql
query {
  findUploadsCreatedAfter(since: "2021-07-08T15:13:08.318949Z" _cursor: null _size: 1000) {
    data {
      content {
        cid
      }
      user {
        issuer
      }
      created
    }
    # cursor value to be used as _cursor for the next page
    after
  }
}
```

A similar thing happens with `@relation` when the relation is a _many_ relationship:

```gql
batchEntries: [BatchEntry!]! @relation
```

```gql
  query {
      findContentByCid(cid: "bafy...") {
        dagSize
        batchEntries {
          data {
            dataModelSelector
            batch {
              cid
            }
          }
        }
      }
  }
```

Note that `BatchEntry` has a 1:1 relation with `Batch` so we don't end up with a `data` property in the selection. i.e. `batch { cid }` NOT `batch { data { cid } }`:

```gql
batch: Batch! @relation
```

alanshaw: Screenshot 2021-07-15 at 12 14 46

See what has happened here - `batchEntries` is actually a resolver function (with nullable params) that returns a page.

packages/api/test/mocks/db/post_graphql.js
alanshaw: yerp, it's a different `data` prop for a GQL response https://graphql.org/learn/serving-over-http/#response - it _has_ to be added in at this level.

packages/db/fauna/resources/Function/findMetrics.js
alanshaw: We did discuss this but decided against. I think it's appropriate for it to be on the `Content` object. If on the pin we'd need to sync the value across all nodes where it is being pinned...

packages/api/src/metrics.js
```diff
@@ -0,0 +1,74 @@
...
+        dealsQueuedTotal,
+        pinsTotal,
+        pinsTotalBytes,
+        pinsQueuedTotal,
+        pinsPinningTotal,
+        pinsPinnedTotal,
+        pinsFailedTotal
+      }
+    }
+    `)
```
alanshaw, line 25: As the dataset gets bigger we might need to split this out into multiple queries.

packages/db/fauna/resources/Function/findMetrics.js
```diff
@@ -0,0 +1,82 @@
...
+const body = Query(
+  Lambda(
+    [],
+    {
+      data: {
+        usersTotal: Count(Documents(Collection('User'))),
+        uploadsTotal: Count(Documents(Collection('Upload'))),
+        contentTotal: Count(Documents(Collection('Content'))),
+        contentTotalBytes: Sum(Match(Index('content_sizes'))),
+        dealsActiveTotal: Count(Match(
```
alanshaw, line 34: We should be aware that `Count` and `Sum` work up to 100K items then we must paginate and page through.

packages/db/fauna/resources/Function/findMetrics.js
```diff
@@ -0,0 +1,82 @@
...
+          Index('deals_by_status'),
+          'Active'
+        )),
+        dealsQueuedTotal: Count(Match(
+          Index('deals_by_status'),
```
alanshaw, line 39: There are no queued deals - a "queued deal" is the existence of a `Batch` object with no associated `Deal` object.

packages/db/fauna/resources/Function/findMetrics.js
```diff
@@ -0,0 +1,82 @@
...
+          'Queued'
+        )),
```
alanshaw, line 41: I did not think the number of active or queued deals was interesting for NFT.Storage and no-one has asked for it yet! The metric we have in NFT.storage is the number of uploads that are in queued/active deals.

packages/db/fauna/resources/Function/findMetrics.js
alanshaw: This will give you just 1 page of data - 64 items by default. You can increase the page size up to 100K but after that we'll need to page through.

packages/db/fauna/resources/Index/pinsContentByStatus.js
```diff
@@ -0,0 +1,36 @@
...
+ * )
+ */
+const index = {
+  name: 'pins_content_by_status',
+  source: Collection('Pin'),
+  terms: [
+    { field: ['data', 'status'] }
+  ],
+  values: [
+    { field: ['data', 'content'] }
```
alanshaw, line 27: We have a pin object for each node a piece of content is pinned on. So if pinned on 3 nodes in the cluster then 3 pins and I assume 3 results for the same content here - they need to be deduplicated before suming.

I'm not sure if this is the right/scalable way to aggregate this information.

alanshaw (changes requested):
My gut instinct is that we'll very quickly find that we do not have enough time in the worker to complete these tasks (you get 30 seconds execution time) and we probably shouldn't expose this endpoint for anyone to hit without some kind of caching.

In NFT.Storage we calculate metrics on a cron and put them in the KV so they can be generated at our own pace and served quickly.

packages/db/fauna/resources/Index/pinsContentByStatus.js
vasco-santos: Exactly. I was thinking the total sum of bytes pinned would be 3x the actual content if we have 3 pin locations. How is this in nft.storage?

packages/api/src/metrics.js
vasco-santos: so probably split between:

- `findUserMetrics`
  - `usersTotal`
- `findUploadMetrics`
  - `uploadsTotal`
  - `contentTotal`
  - `contentTotalBytes`
- `findPinMetrics`
  - `pinsTotal`
  - `pinsTotalBytes`
  - `pinsQueuedTotal`
  - `pinsPinningTotal`
  - `pinsPinnedTotal`
  - `pinsFailedTotal`
- `findDealMetrics`
  - `dealsActiveTotal`
  - `dealsQueuedTotal`

wdyt?

vasco-santos:
> In NFT.Storage we calculate metrics on a cron and put them in the KV so they can be generated at our own pace and served quickly.

We can also do this here, but we need to keep updating a single table entry, which is strange in Fauna

packages/db/fauna/resources/Function/findMetrics.js
vasco-santos: so, I can probably just create a Paginate Function that goes through all the pages.

olizilla:
> My gut instinct is that we'll very quickly find that we do not have enough time in the worker to complete these tasks (you get 30 seconds execution time) and we probably shouldn't expose this endpoint for anyone to hit without some kind of caching.

Are you thinking this query will take more than 30s on Fauna??

packages/db/fauna/resources/Function/findMetrics.js
vasco-santos: dropping for now

packages/db/fauna/resources/Function/findMetrics.js
vasco-santos: dropping for now

vasco-santos:
- added cache
- removed deals metrics for now
- make paginate with max size for now

@olizilla @alanshaw I think we are good to go with this initial iteration and iterate with #93 approach

packages/db/fauna/resources/Function/findMetrics.js
olizilla: A metric that gets referred to a lot is % of successful deals. We should try and add that. Can be a subsequent PR.

vasco-santos: Yes, I had the total deals metric. But Riba might also give us this information from his side, so we might not even need to track it. @alanshaw suggested we wait for now

olizilla (approved):
of note we now have no metrics around filecoin deals. I think we'll want at least one for tracking % of succesful deals, to compare with other endevours.

olizilla:
Let's land this and iterate! I'd like to get a metrics endpoint I can scrape from prometheus!

PR #80, opened 2021-07-13
chore: add api bundlesize check

### LINKED ISSUE #79: wrangler publish timeout
@hugomrdias mentioned this happened in nft.storage recently too... we just need to minify our worker bundle... and keep an eye on the bundle size.

```
> web3.storage-api@0.0.0 build
> webpack

1 asset
202 modules
webpack 5.43.0 compiled successfully in 1517 ms
 Build completed successfully!
Error: Something went wrong! Status: 400 Bad Request, Details {
  "result": null,
  "success": false,
  "errors": [
    {
      "code": 10021,
      "message": "Script startup timed out.\n"
    }
  ],
  "messages": []
}

```

olizilla:
- Add api bundlesize check
- minify bundle

fixes #79 (hopefully)

License: (Apache-2.0 AND MIT)

### DIFF
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -22,6 +22,7 @@ jobs:
         with:
           node-version: 16
       - uses: bahmutov/npm-install@v1
+      - run: npm run build --workspace packages/api
       - run: npm test --workspace packages/api
 
   deploy-staging:
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -12,7 +12,8 @@
     "dev": "wrangler dev --env $(whoami)",
     "publish": "wrangler publish --env $(whoami)",
     "build": "webpack",
-    "test": "npm-run-all -p -r mock:cluster mock:db test:e2e",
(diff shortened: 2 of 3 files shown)


alanshaw (approved):

vasco-santos (commented):
I have been thinking about creating a PR to decrease the max bundle size expectation of the client. And I think we should consider the maxSize here too.

In the client, we have ~30% of the maxSIze which is great (66.25KB < maxSize 200KB ) and here we have ~50% (498.62KB < maxSize 1MB)

I suggest decreasing these. I see this CI job as an alert if we add something considerably big in a PR, before is too late. With this in mind, of course as time goes we might need to increase the maxSize to get in some PRs, but the flag of increasing the bundle size can help us look into better alternatives.

So, I would go for now with like:

client: 80KB
api: 600KB

olizilla:
@vasco-santos good idea. I'd like to put those tighter caps in once we're launched, as we don't know what our baseline bundlesize is until we have the v1 feature set landed.

PR #84, opened 2021-07-14
Update config.yml

jnthnvctr:


alanshaw:
I added some templates with nice green buttons instead.

PR #87, opened 2021-07-14
refactor: allow chunked adding of batch entries

alanshaw:
The max payload for Fauna is around 10MB, but even then the query backing this operation can only deal with around 10,000 entries before it throws "An internal server error occurred. Please contact Fauna support.".

This PR splits the `createBatch` call into two: 1) `createBatch` and 2) `addEntries`, so that a batch can be created and then entries can be added to the batch. The `createBatch` method no longer takes a list of entries.

You use it like this now:

```gql
# Create a new batch
mutation {
  createBatch(data: { batchCid: "bafybatch", pieceCid: "baga" }) {
    # returns a Batch object
    _id
  }
}

# Adds new entries to the batch, existing entries are ignored.
# Call it multiple times, call it with the same entries, nothing goes wrong.
mutation {
  addBatchEntries(
    batchCid: "bafybatch",
    entries: [{ cid: "QmFile", dataModelSelector: "Links/0/Hash" }]
  ) {
    # returns a Batch object also
    _id
  }
}
```

### DIFF
--- a/packages/db/fauna/resources/Index/batchEntryByBatchAndContent.js
+++ b/packages/db/fauna/resources/Index/batchEntryByBatchAndContent.js
@@ -1,0 +1,35 @@
+import fauna from 'faunadb'
+
+const {
+  CreateIndex,
+  Collection,
+  If,
+  Index,
+  Exists,
+  Not
+} = fauna
+
+/**
+ * Usage:
+ *
+ * Match(
+ *   Index('batchEntry_by_batch_and_content'),
+ *   Ref(Collection('Batch'), Var('batchId')),
+ *   Ref(Collection('Content'), Var('contentId'))
+ * )
+ */
+const index = {
+  name: 'batchEntry_by_batch_and_content',
+  source: Collection('BatchEntry'),
+  terms: [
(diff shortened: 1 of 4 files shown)


packages/db/fauna/resources/Index/batchEntryByBatchAndContent.js
```diff
@@ -0,0 +1,35 @@
...
+const index = {
+  name: 'batchEntry_by_batch_and_content',
+  source: Collection('BatchEntry'),
+  terms: [
+    { field: ['data', 'batch'] },
+    { field: ['data', 'content'] }
+  ]
+}
+
+// indexes cannot be updated
```
ribasushi, line 30: What do you mean by this comment? We will be adding stuff incrementally ( over the course of a minute ) - will that not require an index update?

alanshaw: Content of indexes can be updated - I mean you can't change what an index is indexing. You have to delete it and re-create or create one with a different name. Functions on the other hand _can_ be updated.

vasco-santos (approved):
LGTM

Issue #33, opened 2021-07-07
Throw meaningful Magic Errors


vasco-santos:
When authenticated requests fail, we should provide human friendly errors.

Previous context: https://github.com/web3-storage/web3.storage/pull/25#discussion_r664628316

How nft.storage does it: https://github.com/ipfs-shipyard/nft.storage/blob/aad125902fe4f8e6b5144ed52e60be3988f24e1a/packages/api/src/errors.js#L41-L68

vasco-santos:
I'm tackling this as part of the work to integrate sentry

PR #90, opened 2021-07-15
feat: add sentry

vasco-santos:
Adds sentry monitoring to the API and add meaningful errors to Magic Errors

A preview:

image

TODO: 
- [x] add secrets for sentry staging and production
- [ ] add wrangler secrets for staging/prod

Closes: #33

### DIFF
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -39,6 +39,8 @@ jobs:
       - name: Publish api worker
         # workaround for https://github.com/cloudflare/wrangler-action/issues/59 to use node 16
         uses: web3-storage/wrangler-action@node16
+        env:
+          SENTRY_TOKEN: ${{ secrets.STAGING_SENTRY_TOKEN}}
         with:
           apiToken: ${{ secrets.CF_API_TOKEN }}
           workingDirectory: 'packages/api'
--- a/packages/api/src/env.js
+++ b/packages/api/src/env.js
@@ -1,15 +1,32 @@
-/* global MAGIC_SECRET_KEY FAUNA_ENDPOINT FAUNA_KEY SALT CLUSTER_BASIC_AUTH_TOKEN CLUSTER_API_URL */
+/* global MAGIC_SECRET_KEY FAUNA_ENDPOINT FAUNA_KEY SALT CLUSTER_BASIC_AUTH_TOKEN CLUSTER_API_URL SENTRY_DSN */
+import Toucan from 'toucan-js'
 import { Magic } from '@magic-sdk/admin'
 import { DBClient } from '@web3-storage/db'
 import { Cluster } from '@nftstorage/ipfs-cluster'
 
+import pkg from '../package.json'
+
 /** @typedef {{ magic: Magic, db: DBClient, SALT: string }} Env */
 
 /**
- * @param {Request} _
+ * @param {Request} req
  * @param {Env} env
+ * @param {FetchEvent} event
  */
-export function envAll (_, env) {
+export function envAll (_, env, event) {
+  console.log('sentry', env.SENTRY_DSN || SENTRY_DSN)
+  env.sentry = new Toucan({
+    dsn: env.SENTRY_DSN || SENTRY_DSN,
+    event,
+    allowedHeaders: ['user-agent'],
+    allowedSearchParams: /(.*)/,
+    debug: false,
+    rewriteFrames: {
+      root: '/'
+    },
+    // version: env.VERSION || VERSION, // TODO: version endpoint?
+    pkg
+  })
   env.magic = new Magic(env.MAGIC_SECRET_KEY || MAGIC_SECRET_KEY)
 
   env.db = new DBClient({
--- a/packages/api/src/env.js
+++ b/packages/api/src/env.js
@@ -14,8 +14,7 @@ import pkg from '../package.json'
  * @param {FetchEvent} event
  */
 export function envAll (_, env, event) {
-  console.log('sentry', env.SENTRY_DSN || SENTRY_DSN)
-  env.sentry = new Toucan({
+  env.sentry = (env.SENTRY_DSN || typeof SENTRY_DSN !== 'undefined') && new Toucan({
     dsn: env.SENTRY_DSN || SENTRY_DSN,
     event,
     allowedHeaders: ['user-agent'],
--- a/packages/api/src/error-handler.js
+++ b/packages/api/src/error-handler.js
@@ -1,0 +1,60 @@
+import { ErrorCode as MagicErrors } from '@magic-sdk/admin'
+import { JSONResponse } from './utils/json-response.js'
+import { HTTPError, UserNotFoundError, TokenNotFoundError } from './errors.js'
+
+/**
+ * @param {Error & {status?: number;code?: string;}} err
+ */
+export function errorHandler (err) {
+  let error = {
+    code: err.code,
+    message: err.message
+  }
+  let status = err.status || 500
+
+  if (err instanceof HTTPError) {
+    return new JSONResponse(error, { status })
+  }
+
+  if (err instanceof UserNotFoundError || err instanceof TokenNotFoundError) {
+    return new JSONResponse(error, { status })
+  }
+
+  switch (err.code) {
+    // Magic SDK errors
+    case MagicErrors.TokenExpired:
+      status = 401
+      error.message = 'API token has expired.'
+      break
+    case MagicErrors.ExpectedBearerString:
+      status = 401
+      error.message =
+        'API token is missing, make sure the `Authorization` header has a value in the following format `Bearer {token}`.'
+      break
+    case MagicErrors.MalformedTokenError:
+      status = 401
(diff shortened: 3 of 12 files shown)


packages/api/src/errors.js
```diff
@@ -0,0 +1,89 @@
...
+        // catch all server errors
+        if (status >= 500) {
+          error = {
+            code: err.name,
+            message: err.message || 'Server Error'
+          }
+          sentry.captureException(err)
+        } else {
+          // Custom HTTPError
+          error = {
```
vasco-santos, line 79: changed the previous error to have a code and a message in the body. Not sure if we want to go this way, or drop the error codes from Magic, no strong feelings

.github/workflows/api.yml
```diff
@@ -39,6 +39,8 @@ jobs:
       - name: Publish api worker
         # workaround for https://github.com/cloudflare/wrangler-action/issues/59 to use node 16
         uses: web3-storage/wrangler-action@node16
+        env:
+          SENTRY_TOKEN: ${{ secrets.SENTRY_TOKEN}}
```
vasco-santos, line 43: needs to be added

packages/api/src/errors.js
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  /**
+   * @param {string} message
+   * @param {number} [status]
+   * @returns {never}
+   */
+  static throw (message, status) {
+    throw new this(message, status)
+  }
```
alanshaw, line 23: I think `throw new HTTPError('boom', 500)` is fine, and would argue that `HTTPError.throw('boom', 500)` is not any more convenient.

```suggestion
```

packages/api/src/env.js
```diff
@@ -1,15 +1,32 @@
...
+    event,
+    allowedHeaders: ['user-agent'],
+    allowedSearchParams: /(.*)/,
+    debug: false,
+    rewriteFrames: {
+      root: '/'
+    },
+    // version: env.VERSION || VERSION, // TODO: version endpoint?
+    pkg
+  })
```
alanshaw, line 29: Please can sentry be optional? We don't need it when developing.

packages/api/src/errors.js
```diff
@@ -0,0 +1,102 @@
...
+    this.name = 'UserNotFound'
+    this.status = 401
+    this.code = ErrorUserNotFound.CODE
+  }
+}
+
+ErrorUserNotFound.CODE = 'ERROR_USER_NOT_FOUND'
+
+export class ErrorTokenNotFound extends Error {
+  constructor (msg = 'API Key not found.') {
```
alanshaw, line 94:
```suggestion
  constructor (msg = 'API token not found.') {
```

packages/api/src/errors.js
```diff
@@ -0,0 +1,102 @@
...
+
+    switch (err.code) {
+      case ErrorUserNotFound.CODE:
+      case ErrorTokenNotFound.CODE:
+        break
+
+      // Magic SDK errors
+      case MagicErrors.TokenExpired:
+        status = 401
+        error.message = 'API Key has expired.'
```
alanshaw, line 35:
```suggestion
        error.message = 'API token has expired.'
```

packages/api/src/errors.js
```diff
@@ -0,0 +1,102 @@
...
+        break
+      case MagicErrors.ExpectedBearerString:
+        status = 401
+        error.message =
+          'API Key is missing, make sure the `Authorization` header has a value in the following format `Bearer {api key}`.'
```
alanshaw, line 40:
```suggestion
          'API token is missing, make sure the `Authorization` header has a value in the following format `Bearer {token}`.'
```

packages/api/src/errors.js
```diff
@@ -0,0 +1,102 @@
...
+        break
+      case MagicErrors.MalformedTokenError:
+        status = 401
+        error.message = 'API Key is malformed or failed to parse.'
```
alanshaw, line 44:
```suggestion
        error.message = 'API token is malformed or failed to parse.'
```

packages/api/src/errors.js
```diff
@@ -0,0 +1,102 @@
...
+          }
+        }
+        break
+    }
+
+    return new JSONResponse({ error }, { status })
+  }
+}
+
+export class ErrorUserNotFound extends Error {
```
alanshaw, line 82: I think it's much more common to see `Error` as a suffix than a prefix. Including within JS itself (`TypeError`, `RangeError` etc.)

```suggestion
export class UserNotFoundError extends Error {
```

packages/api/src/errors.js
```diff
@@ -0,0 +1,102 @@
...
+    super(msg)
+    this.name = 'UserNotFound'
+    this.status = 401
+    this.code = ErrorUserNotFound.CODE
+  }
+}
+
+ErrorUserNotFound.CODE = 'ERROR_USER_NOT_FOUND'
+
+export class ErrorTokenNotFound extends Error {
```
alanshaw, line 93:
```suggestion
export class TokenNotFoundError extends Error {
```

packages/api/src/errors.js
```diff
@@ -0,0 +1,102 @@
...
+          // Custom HTTPError
+          error = {
+            code: err.code || 'HTTP_ERROR',
+            message: err.message || 'Server Error'
+          }
+        }
+        break
+    }
+
+    return new JSONResponse({ error }, { status })
```
alanshaw, line 78: Do we need an extra property to traverse in the response? It's like having `value`/`data` in a non-error resposne.

```suggestion
    return new JSONResponse(error, { status })
```

packages/api/src/index.js
```diff
@@ -1,14 +1,24 @@
...
 
-const router = Router()
+// Extended itty router for custom error handler
+const ThrowableRouter = (options = {}) =>
+  new Proxy(Router(options), {
+    get: (obj, prop) => (...args) =>
+      prop === 'handle'
+        ? obj[prop](...args).catch((err) => serverError(err, ...args))
+        : obj[prop](...args)
+  })
```
alanshaw, line 19: 😱 why do we need this?

alanshaw (changes requested):

packages/api/src/index.js
vasco-santos: // Extended itty router for custom error handler

We need to be able to give `env` to the serverError function below, so that we can use `env.sentry` when catching errors from handlers.

Also, this is based on their recommendations for error handling: https://github.com/kwhitley/itty-router#extending-itty-router

packages/api/src/env.js
vasco-santos: sure!

packages/api/src/errors.js
```diff
@@ -0,0 +1,34 @@
...
+    this.name = 'HTTPError'
+    this.status = status
+  }
+}
+
+export class UserNotFoundError extends HTTPError {
+  constructor (msg = 'User not found.') {
+    super(msg)
+    this.name = 'UserNotFound'
+    this.status = 401
```
alanshaw, line 18:
```suggestion
    super(msg, 401)
    this.name = 'UserNotFound'
```

packages/api/src/errors.js
```diff
@@ -0,0 +1,34 @@
...
+  }
+}
+
+UserNotFoundError.CODE = 'ERROR_USER_NOT_FOUND'
+
+export class TokenNotFoundError extends HTTPError {
+  constructor (msg = 'API token not found.') {
+    super(msg)
+    this.name = 'TokenNotFound'
+    this.status = 401
```
alanshaw, line 29:
```suggestion
    super(msg, 401)
    this.name = 'TokenNotFound'
```

packages/api/src/error-handler.js
```diff
@@ -0,0 +1,56 @@
...
+ * @param {Error & {status?: number;code?: string;}} err
+ */
+export function errorHandler (err) {
+  let error = {
+    code: err.code,
+    message: err.message
+  }
+  let status = err.status || 500
+
+  if (err instanceof HTTPError || err instanceof UserNotFoundError || err instanceof TokenNotFoundError) {
```
alanshaw, line 15: `err instanceof HTTPError` should be true when `err` is a `UserNotFoundError` or `TokenNotFoundError`.

packages/api/src/index.js
```diff
@@ -50,19 +51,24 @@ router.get('/', () => {
...
+ * @param {Error} error
+ * @param {Request} request
+ * @param {import('./env').Env} env
+ */
+function serverError (error, request, env) {
   console.error(error.stack)
-  const message = error.message || 'Server Error'
-  const status = error.status || 500
-  return addCorsHeaders(request, new JSONResponse({ message }, { status }))
+  env.sentry && env.sentry.captureException(error)
```
alanshaw, line 61: Please can this move into the error handler, and only report when the status is >=500...also the `console.error` line?

alanshaw (changes requested):
Some tiny tweaks and then I think we're done. Thank you! 🙏

alanshaw (approved):

PR #95, opened 2021-07-15
feat: client.status(cid)

olizilla:
- fetch the /status/:cid data for a given cid
- return undefined on 404... like asking a Map for a key it doesn't have
- throw on server error

License: (Apache-2.0 AND MIT)

### DIFF
--- a/packages/client/src/lib.js
+++ b/packages/client/src/lib.js
@@ -130,7 +130,7 @@ class Web3Storage {
 
           const res = await pRetry(
             async () => {
-              const request = await fetch(url.toString(), {
+              const request = await fetch(url, {
                 method: 'POST',
                 headers,
                 body: carFile
@@ -161,12 +161,12 @@ class Web3Storage {
 
   /**
    * @param {Service} service
-   * @param {string} cid
+   * @param {CIDString} cid
    * @returns {Promise<Web3Response | null>}
    */
   static async get ({ endpoint, token }, cid) {
     const url = new URL(`/car/${cid}`, endpoint)
-    const res = await fetch(url.toString(), {
+    const res = await fetch(url, {
       method: 'GET',
       headers: Web3Storage.headers(token)
     })
@@ -184,13 +184,33 @@ class Web3Storage {
 
   /**
    * @param {Service} service
-   * @param {string} cid
+   * @param {CIDString} cid
    * @returns {Promise<CIDString>}
    */
   /* c8 ignore next 4 */
   static async delete ({ endpoint, token }, cid) {
     console.log('Not deleteing', cid, endpoint, token)
     throw Error('.delete not implemented yet')
+  }
+
+  /**
+   * @param {Service} service
+   * @param {CIDString} cid
+   * @returns {Promise<CIDString>}
+   */
+  static async status ({ endpoint, token }, cid) {
+    const url = new URL(`/status/${cid}`, endpoint)
+    const res = await fetch(url, {
+      method: 'GET',
+      headers: Web3Storage.headers(token)
+    })
+    if (res.status === 404) {
+      return undefined
+    }
+    if (!res.ok) {
+      throw new Error(res.statusText)
+    }
+    return res.json()
   }
 
   // Just a sugar so you don't have to pass around endpoint and token around.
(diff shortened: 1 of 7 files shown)


olizilla:
Somthing is off with the types we're using for fetch... it things a URL is not a valid parameter to fetch, but the spec would disagree.

vasco-santos:
> Somthing is off with the types we're using for fetch... it thinks a URL is not a valid parameter to fetch, but the spec would disagree.

Here is the problem: 
- https://github.com/web-std/io/blob/main/fetch/%40types/index.d.ts#L175
- type RequestInfo = string | Body;

packages/client/src/lib.js
```diff
@@ -193,6 +193,26 @@ class Web3Storage {
     throw Error('.delete not implemented yet')
   }
 
+  /**
+   * @param {Service} service
+   * @param {CIDString} cid
+   * @returns {Promise<CIDString>}
```
alanshaw, line 199: Incorrect return type!

packages/client/src/lib.js
```diff
@@ -193,6 +193,26 @@ class Web3Storage {
...
+      method: 'GET',
+      headers: Web3Storage.headers(token)
+    })
+    if (res.status === 404) {
+      return undefined
+    }
+    if (!res.ok) {
+      throw new Error(res.statusText)
+    }
+    return res.json()
```
alanshaw, line 213: Shall we transform the date strings to `Date` objects? Asking because we do in nft.storage.

packages/client/src/lib/interface.ts
```diff
@@ -40,6 +40,11 @@ export interface API {
    * Remove a users record of an upload. Does not make CID unavailable.
    */
   delete(service: Service, cid: CIDString): Promise<CIDString>
+
+  /**
+   * Get info on Filecoin deals and IPFS pins that a CID is replicated in.
+   */
+  status(service: Service, cid: CIDString): Promise<CIDString>
```
alanshaw, line 47: Needs correct return type.

alanshaw (changes requested):

packages/client/src/lib.js
olizilla: i'm disinclined to do that. The user might want moments, or luxons or date-fns, or might not care about those fields. An ISO string is a nice thing.

olizilla:
Return types are fixed up. Updating to the latest @web-std/fetch did not appease tsc, so I have reverted to called in url.toString for now, as it's harmless, just a little annoying.

packages/client/src/lib/interface.ts
```diff
@@ -62,3 +67,30 @@ export interface Web3Response extends Response {
...
+  region: string,
+  status: 'Pinned' | 'Pinning' | 'PinQueued'
+}
+
+export interface Deal {
+  dealId: number,
+  miner: string,
+  status: 'Queued' | 'Published' | 'Active'
+  pieceCid: string,
+  dealCid: string,
```
alanshaw, line 83:
```suggestion
  dataCid: string,
```

packages/client/src/lib/interface.ts
```diff
@@ -62,3 +67,30 @@ export interface Web3Response extends Response {
...
+  dataModelSelector: string,
+  activation: string,
+  created: string
+  updated: string
+}
+
+export interface Status {
+  cid: string
```
alanshaw, line 91:
```suggestion
  cid: CIDString
```

packages/client/src/lib/interface.ts
```diff
@@ -62,3 +67,30 @@ export interface Web3Response extends Response {
   unixFsIterator: () => AsyncIterable<UnixFSEntry>
   files: () => Promise<Array<Web3File>>
 }
+
+export interface Pin {
+  peerId: string,
+  peerName: string,
+  region: string,
+  status: 'Pinned' | 'Pinning' | 'PinQueued'
```
alanshaw, line 75:
```suggestion
  status: 'Pinned' | 'Pinning' | 'PinQueued',
  updated: string
```

alanshaw (changes requested):

olizilla:
Yikes. Thanks for the fixes. I'm gonna step away from the keys; I must be more tired than I thought

alanshaw (approved):

PR #97, opened 2021-07-16
feat: update typography and add colors

jdiogopeixoto:
- Changes the typography to use "Open Sans" instead of "Helvetica"
- Adds colors to the tailwind config using the same naming as the ones in the [docs](https://github.com/web3-storage/docs/blob/feat/initial-branding/docs/.vuepress/theme/styles/palette.styl)
- Uses tailwind `purge` to reduce bundle size


rafaelramalho19 (approved):
LGTM

PR #98, opened 2021-07-16
feat: `w3 token` and `w3 status` commands

olizilla:
`w3 token` will prompt you to paste in your api key

```console
$ w3 token
✔ Paste your API token for api.web3.storage · eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJk...etc

⁂ API token saved
```

and as with all other commands, you can pass in a custom --api

```console
w3 token --api https://api-staging.web3.storage
✔ Paste your API token for api-staging.web3.storage · eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJk...etc

⁂ API token saved
```

and now w3 will save that token and api (via [`conf`](https://github.com/sindresorhus/conf) to an appropriate config dir per you OS. AND! it warns you if you are using an API other than PROD! 🎉
```console
w3 put ~/Pictures/pmf.jpg
⁂ Stored 1 file
⁂ https://dweb.link/ipfs/bafybeicwehe4cx74f5gf2duol7g5kgqmkh4z7ivnqikgqgspckqazz6ygi
```

as a bonus, the `w3 status <cid>` command is also ready to print some JSON about what we know of IPFS pins and Filecoin deals for a given CID, pending #95 landing.

License: (Apache-2.0 AND MIT)

### DIFF
--- a/packages/w3/bin.js
+++ b/packages/w3/bin.js
@@ -14,7 +14,7 @@ cli
 cli.command('token')
   .option('--api', 'URL for API to use')
   .option('--token', 'API token to use')
-  .describe('Save an API token use for all requests')
+  .describe('Save an API token to use for all requests')
   .action(token)
 
 cli.command('status <cid>')
--- a/packages/w3/bin.js
+++ b/packages/w3/bin.js
@@ -6,21 +6,18 @@ import { get, put, status, token } from './index.js'
 const cli = sade('w3')
 
 cli
-  .option('--api', 'URL for API to use')
-  .option('--token', 'API token to use')
   .example('put path/to/files')
   .example('get bafkreigh2akiscaildcqabsyg3dfr6chu3fgpregiymsck7e7aqa4s52zy -o room-guardian.jpg')
 
 cli.command('token')
-  .option('--api', 'URL for API to use')
-  .option('--token', 'API token to use')
-  .describe('Save an API token to use for all requests')
+  .option('--api', 'URL for the Web3 Storage API. Default: https://api.web3.storage')
+  .option('--delete', 'Delete your saved token')
+  .describe('Save an API token to use for all requests.')
   .action(token)
 
-cli.command('status <cid>')
-  .describe('Get the Filecoin deals and IPFS pins that contain a given CID, as JSON.')
-  .example('status bafkreigh2akiscaildcqabsyg3dfr6chu3fgpregiymsck7e7aqa4s52zy')
-  .action(status)
+cli.command('put <path>')
+  .describe('Upload a file or directory to web3.storage')
+  .action(put)
 
 cli.command('get <cid>')
   .describe('Fetch and verify files from web3.storage')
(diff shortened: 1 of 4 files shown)


packages/w3/bin.js
```diff
@@ -11,6 +11,17 @@ cli
   .example('put path/to/files')
   .example('get bafkreigh2akiscaildcqabsyg3dfr6chu3fgpregiymsck7e7aqa4s52zy -o room-guardian.jpg')
 
+cli.command('token')
+  .option('--api', 'URL for API to use')
+  .option('--token', 'API token to use')
```
vasco-santos, line 16: should we drop this and only allow token via prompt to not be pasted into commands history?

packages/w3/index.js
```diff
@@ -1,8 +1,71 @@
...
+
+/**
+ * Get a new API client configured either from opts or config
+ * @param {object} opts
+ * @param {string} [opts.api]
+ * @param {string} [opts.token]
+ */
+function getClient ({
+  api = config.get('api') || API,
+  token = config.get('token')
```
vasco-santos, line 19: I would show the prompt if not existent and drop support to add via CLI param

vasco-santos (commented):
Overall looking good.
Left some feedback on the token insertion

packages/w3/bin.js
olizilla: it's your command history. if you wanna paste in a token, go nuttsss i say!

packages/w3/index.js
olizilla: hmm maybe that is a good compromise tho. catches the cases where you didn't know you had to set a token too,

packages/w3/bin.js
vasco-santos: yes, my point here is mostly that this will appear in the help of the command, which will guide people towards providing the token as a CLI parameter, as they will likely not understand they will be prompted

packages/w3/bin.js
```diff
@@ -11,6 +11,17 @@ cli
...
+  .describe('Save an API token use for all requests')
```
alanshaw, line 17:
```suggestion
  .describe('Save an API token to use for all requests')
```

packages/w3/index.js
```diff
@@ -1,8 +1,71 @@
...
+    })
+    token = response.token
+  }
+  config.set('token', token)
+  config.set('api', api)
+  console.log('⁂ API token saved')
+}
+
+/**
+ * Get data on Filecoin deals and IPFS pins that contain a given CID, as JSON.
```
alanshaw, line 55: It would be nice to have a pretty print by default and an option to print as JSON.

alanshaw (approved):

packages/w3/index.js
olizilla: but JSON is pretty!

olizilla:
@vasco-santos how abou this? https://github.com/web3-storage/web3.storage/pull/98/commits/12595b84e4931143e8940cc0230fd8ff17abd22c

the global `--api` and `--token` commands are now secret! only for fancy developers who know. They do not appear in the command help, but still work if you choose to use them. 

I've added a `token --delete`, just so you can clear it out if you want. I've left it prompting the user to run the `token` command if they dont have one set, as I dont want someone running `we get` in CI and then wondering why the command is hanging. It should just fail fast and prompt to add a token.

vasco-santos (approved):
This is great! 🎉 🚀

PR #99, opened 2021-07-16
chore: use files-from-path module

vasco-santos:
Removes `globSource` code in favour of https://github.com/web3-storage/glob-to-files which is essentially a copy of it for now.

This also aims to make it easier to use the client in Node.js and read files from `fs`. However, given that `client.put` receives an iterable of Files, we can even make this easier by making `glob-to-files` to also export a function with the final files list as an array.

This way we could do in Node.js:

```
await client.put(globToFiles(fsPath))
```

The async iterator is probably nice to have, so I would export both function.

Accepting better names for the module (Friday is even a more difficult day to name things)

Let me know your thoughts @olizilla


alanshaw:
My vote would be to avoid the word "glob" if possible...I think we kinda want to accept file path(s) and get back file objects, but we support "globs" in those paths. Something like `filepathToFile`/`filesFromPath` (needs work), but the idea is that we make it easy to convert a _file path_ to a _file object(s)_, but the path can be a directory or a glob.

vasco-santos:
`filesFromPath` seems a good naming

olizilla (approved):
That is neat! My only (very minor) thought is `files-from-path` doesn't feel like it's related to `web3-storage` so might look a little odd in the docs when explaining how to do this for node devs.

```js
import { filesFromPath } from '@web3-stroage/files'
```

might tie it in better or just exposing it as an export from `web3.storage`

```js
import { Web3Storage, filesFromPath } from 'web3.storage'
```

but not a blocker here, just surfacing my feels.

olizilla:
also note that `filesFromPath` does not strictly return "Files" in the web sense, but Filelike things that can have many more properties. Again, I don't think we should use the term `Filelike` here, and this PR is probably fine as is, just surfacing thoughts.

vasco-santos:
I agree @olizilla ! I will create a follow up PR adding it to the client and exporting it for users

olizilla:
It of course might be a good thing that `files-from-path` looks like a totally speperate modules, as it would be fine for node devs to bring any module they like to recurisvely find all the paths in this dir and map them to something that `web3.storage.put` can handle.

@terichadbourne do you have any feels on this? This PR is providing a module to make it easy for a node dev to pass all the files, recursively, in a directory to the `web3.storage` js client `put` method.

vasco-santos:
I think it should be a separate module, but we should also export it from web3storage, so that from the user stand point the objective of it is clear.

So, I would go as you suggest:

```js
import { Web3Storage, filesFromPath } from 'web3.storage'
```

where basically web3.storage imports and exports the util

vasco-santos:
I added it to the client already

packages/client/src/platform.web.js
```diff
@@ -7,3 +7,7 @@ export const Response = globalThis.Response
 export const Blob = globalThis.Blob
 export const File = globalThis.File
 export const Blockstore = MemoryBlockStore
+// @ts-ignore filesFromPath any type
+export const filesFromPath = undefined
+// @ts-ignore getFilesFromPath any type
+export const getFilesFromPath = undefined
```
alanshaw, line 13: Perhaps these should be more like https://github.com/ipld/js-car/blob/master/lib/indexed-reader-browser.js so that anyone trying to use them in the browser is given a reason why they're not there.

packages/client/src/platform.js
```diff
@@ -10,5 +11,7 @@ export {
   Headers,
   Blob,
   File,
-  Blockstore
+  Blockstore,
+  filesFromPath,
+  getFilesFromPath
```
alanshaw, line 16: What is the difference?

vasco-santos: `getFilesFromPath` iterates the `filesFromPath` async iterator and returns an array with all the files. I think users will likely rely on the `getFilesFromPath`, but in the CLI we do a transformation so the iterator is also nice for such use cases.

packages/client/src/platform.web.js
vasco-santos: yes, this is a good idea!

yusefnapora:
Nice! I'll add this to the storage how-to once it lands.

alanshaw (approved):
LGTM - it would be awesome to get a new node.js example that uses this.

vasco-santos:
> LGTM - it would be awesome to get a new node.js example that uses this.

Yes! I am waiting on this to be merged to add it here: https://github.com/web3-storage/web3.storage/pull/103

vasco-santos:
I will rebase this PR now and merge it 🎉

PR #100, opened 2021-07-16
feat: style sign-in-page

rafaelramalho19:
- Styles sign-in-page
- Adds page background variant possibility

## ⚠️ TODO IN OTHER PRs: 

- Fix navbar background (@jdiogopeixoto)
- Fix footer positioning

## 👀  Visuals 
image


alanshaw:
> Adds page background variant possibility

I'm not sure I understand this bit!

The background looks much darker than the design - I'm not sure it's on purpose?

Screenshot 2021-07-16 at 21 22 56

packages/website/components/layout.js
```diff
@@ -32,7 +34,7 @@ export default function Layout({
   const shouldWaitForUser = needsUser && status === 'loading'
 
   return (
-    <div>
+    <div className={clsx(pageBgColor, 'min-h-screen')}>
```
jdiogopeixoto, line 37: We need to add `flex flex-col` here to solve the layout issue

packages/website/pages/login.js
```diff
@@ -1,15 +1,22 @@
...
+ *
+ * @returns {{ props: import('../components/types.js').LayoutProps}}
+ */
 export function getStaticProps() {
   return {
     props: {
       title: 'Login - Web3 Storage',
       redirectTo: '/files',
       redirectIfFound: true,
+      pageBgColor: 'bg-w3storage-red-dark',
```
jdiogopeixoto, line 19: This should be using the color `w3storage-red` according to Figma

jdiogopeixoto (changes requested):
Looks good! Only left 2 small comments

packages/website/pages/login.js
rafaelramalho19: You're right good catch

packages/website/components/layout.js
rafaelramalho19: Was gonna leave that for another task, but I added now

jdiogopeixoto (approved):
🚀

PR #102, opened 2021-07-19
feat!: fold in content into user uploads list

olizilla:
"content" is a feature of our internal schema. I don't think it helps users. So it's removed from the `GET/ user/uploads` response by spreading the properties into each upload info.

**before**

```json
[{
  "name": "pinpie.jpg",
  "content": {
    "cid": "bafkreiajkbmpugz75eg2tmocmp3e33sg5kuyq2amzngslahgn6ltmqxxfa",
    "dagSize": null
  },
  "created": "2021-07-09T10:36:05.862862Z"
}]
```

**after**

```json
[{
  "name": "pinpie.jpg",
  "cid": "bafkreiajkbmpugz75eg2tmocmp3e33sg5kuyq2amzngslahgn6ltmqxxfa",
  "dagSize": null,
  "created": "2021-07-09T10:36:05.862862Z"
}]
```

Is the plan for pagaination that the caller should grab the `created` property of the last item and pass it in as the `before` prop on the next request? We could lift that out into an envelope and have responses be like

```js
{
  data: [ /* the payload */ ]
  next: '/user/uploads?before=<iso-data-stamp>
}
```

or similar to make that a litte more friendly... Interested in feedback here. Am going to build out the js client method for it and see how it feels.

License: (Apache-2.0 AND MIT)


vasco-santos (commented):
Folding the content looks good to me.
Regarding the pagination, I think we should do it as you suggest, by changing to a data+next object. So probably, we should land that here in this PR

alanshaw (approved):
LGTM

re: pagination

I'm just wondering if there's any benefit to having the next/previous links as HTTP header(s) ([github uses the `Link` header](https://docs.github.com/en/rest/guides/traversing-with-pagination#basics-of-pagination)) since it is metadata?

I'd also be fine with next/prev links in the response. Both approaches seem to be common. Perhaps we could call the `data` property `results` instead in this case?

If we have a client method we'd have to figure out how to call that method again to get the next page and I'm not sure how we'd use the `next` URL here (unless it can be passed to the method...).

olizilla:
ooh good idea @alanshaw, I think Link header is the way to go for RESTish thingsm and means we dont have to change the shape of the response.

olizilla:
This PR now adds a`Link` header where it is returning a full page of things like

```
Link: </user/uploads?size=25&before=<iso date">; rel="next"
```

This is the start of how we might do pagination on listing endpoints for restful apis... but there is more to figure out here. We could ask the db for size +1 from the db to figure out if there actually is an next page, to avoid the edge case of the end of the result set falling on a page boundary...

it'd be nice to figure out how to add a "rel=last" link a la github pagination, to inform the caller of the url for the last page, but this would need more smarts int the query. Same for `rel=prev` as we'd need to fetch the prev page to know the `created` time of the first result of that page. But I'm inclined to leave that for subsequent PRs if folks would find it useful

alanshaw (approved):

PR #108, opened 2021-07-19
feat: self-host via add-to-web3 action

olizilla:
see: https://github.com/web3-storage/add-to-web3/pull/1

License: (Apache-2.0 AND MIT)


.github/workflows/website.yml
```diff
@@ -42,16 +42,14 @@ jobs:
...
-        uses: ipfs-shipyard/ipfs-github-action@v2.1.0
+      # Add the site to web3.storage, output the cid as `steps.ipfs.outputs.cid`
+      - name: Add to web3.storage
+        uses: web3-storage/add-to-web3@mvp
         id: ipfs
         with:
           path_to_add: packages/website/out
-          cluster_host: /dnsaddr/filecoin.storage.ipfscluster.io
-          cluster_user: ${{ secrets.CLUSTER_USER }}
-          cluster_password: ${{ secrets.CLUSTER_PASSWORD }}
```
alanshaw, line 54: Just a reminder to remove these two secrets from the repo after merge as this is the only thing that was using them.

alanshaw (approved):

olizilla:
This is blocked on deciding how and what params to expose in ipfs-car https://github.com/web3-storage/ipfs-car/pull/61

olizilla:
...becuase we're always wrapping with a directory right now, and the github action wants to be able to let folks pin a directory and publish it as a website, for which the wrapping directory is not wanted.

.github/workflows/website.yml
olizilla: IT IS DONE

PR #110, opened 2021-07-20
feat: POST /upload

vasco-santos:
Adds POST /file to the API, which accepts both multipart and single blob.

This gets https://github.com/ssttevee/js-multipart-parser code in for the multipart parse, similarly to nft.storage API. We should probably just get our own module of this.

~~Note: Add directory test is still not working (but it works with Postman...)~~
- needs to run on webkit in playwright, given is the only place that supports reading body


packages/api/src/upload.js
```diff
@@ -0,0 +1,92 @@
...
+    const files = /** @type {File[]} */ (form.getAll('file'))
+    const dirSize = files.reduce((total, f) => total + f.size, 0)
+
+    const dir = await env.cluster.addDirectory(files, {
+      metadata: { size: dirSize.toString() }
+    })
+    const { cid: cCId, size } = dir[dir.length - 1]
+
+    cid = cCId
+    dagSize = size
```
alanshaw, line 47:
```suggestion
    const entries = await env.cluster.addDirectory(files, {
      metadata: { size: dirSize.toString() },
      // When >2.5MB, use local add, because waiting for blocks to be sent to
      // other cluster nodes can take a long time. Replication to other nodes
      // will be done async by bitswap instead.
      local: dirSize > LOCAL_ADD_THRESHOLD
    })
    const dir = entries[entries.length - 1]

    cid = dir.cid
    dagSize = dir.size
```

packages/api/src/upload.js
```diff
@@ -0,0 +1,92 @@
...
+      // When >2.5MB, use local add, because waiting for blocks to be sent to
+      // other cluster nodes can take a long time. Replication to other nodes
+      // will be done async by bitswap instead.
+      local: blob.size > LOCAL_ADD_THRESHOLD
+    })
+
+    cid = cCId
+    dagSize = size
+
+    return new JSONResponse({ cid })
```
alanshaw, line 65:
```suggestion
```

packages/api/src/upload.js
```diff
@@ -0,0 +1,92 @@
...
+    status: toPinStatusEnum(status),
+    location: { peerId, peerName }
+  }))
+
+  if (!pins.length) { // should not happen
+    throw new Error('not pinning on any node')
+  }
+
+  // Store in DB
+  await env.db.query(IMPORT_CAR, {
```
alanshaw, line 80: ha, okay would you please open an issue for us to identify the upload method in the `Upload` collection, and change this mutation to take the method as a param?

packages/api/src/utils/multipart/utils.js
```diff
@@ -0,0 +1,45 @@
...
+  if (a.length !== b.length) {
+    return false
+  }
+  for (let i = 0; i < a.length; i++) {
+    if (a[i] !== b[i]) {
+      return false
+    }
+  }
+  return true
+}
```
alanshaw, line 45: Please open an issue to pull this out as a dependency (or just do it 😉). It'll be useful for NFT.Storage also.

packages/api/test/mocks/cluster/post_add.js
```diff
@@ -0,0 +1,22 @@
+// @ts-ignore not typed
+const Hash = require('ipfs-only-hash')
```
alanshaw, line 2: We should probably use the importer directly, this just returns the hash of the first file...(which was good enough in nft.storage because we only had an upload blob test)

packages/api/src/upload.js
```diff
@@ -0,0 +1,92 @@
...
+    const { cid: cCId, size } = await env.cluster.add(blob, {
+      metadata: { size: blob.size.toString() },
+      // When >2.5MB, use local add, because waiting for blocks to be sent to
+      // other cluster nodes can take a long time. Replication to other nodes
+      // will be done async by bitswap instead.
+      local: blob.size > LOCAL_ADD_THRESHOLD
+    })
+
+    cid = cCId
+    dagSize = size
```
alanshaw, line 63:
```suggestion
    const entry = await env.cluster.add(blob, {
      metadata: { size: blob.size.toString() },
      // When >2.5MB, use local add, because waiting for blocks to be sent to
      // other cluster nodes can take a long time. Replication to other nodes
      // will be done async by bitswap instead.
      local: blob.size > LOCAL_ADD_THRESHOLD
    })

    cid = entry.cid
    dagSize = entry.size
```

alanshaw (changes requested):
🙌 Awesome!

packages/api/src/utils/multipart/utils.js
vasco-santos: Yes! I will open an issue and do this as a follow up to also PR NFT.storage

packages/api/src/upload.js
vasco-santos: https://github.com/web3-storage/web3.storage/issues/115

packages/api/src/utils/multipart/utils.js
vasco-santos: https://github.com/web3-storage/web3.storage/issues/114

vasco-santos:
@alanshaw review addressed 🚀  let me know if we should change the size stuff

alanshaw (approved):

PR #120, opened 2021-07-21
feat: mock deals script

alanshaw:
This adds a script for creating some mock deals in your local development database for content you've already added.

Make sure your database is up to date first by running `npm run import` in the `packages/db` directory.

It's safe to run multiple times to get more deals for your data.

Sample script output:

Screenshot 2021-07-21 at 15 45 09

After the script has run you should be able to see deals for your CIDs e.g.

Screenshot 2021-07-21 at 16 03 53

Screenshot 2021-07-21 at 16 03 43


vasco-santos (approved):
LGTM

PR #122, opened 2021-07-22
feat: update Content dagSize from addAggregateEntries

alanshaw:
If DAG size of each entry is sent as part of a call to `addAggregateEntries` and the `Content` the entry refers to does not yet have a `dagSize` set then set it.

This allows us to avoid having a separate cron job for obtaining the size of BIG DAGs since this information is sent from the Storage Broker.


vasco-santos (approved):
This is cool 👌🏼 🎉

olizilla (approved):

PR #123, opened 2021-07-22
feat!: add upload type property

### LINKED ISSUE #115: Identify upload method in the Upload collection
We should add the upload method to the Upload Collection and update CAR and upload mutations to set it.

alanshaw:
This PR adds a `type` property to `Upload` allowing us to track and produce metrics on the type of upload data received. It also renames `importCar` method to `createUpload` that takes a required `type` param.

Note: if you have existing uploads you'll need to add a `type: 'Car'` parameter to them in leu of a migration!

resolves #115

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "web3.storage",
       "version": "1.0.0",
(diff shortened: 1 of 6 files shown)


vasco-santos (approved):
LGTM

PR #125, opened 2021-07-22
chore: update the cluster domain

olizilla:
web3.storage.ipfscluster.io is the thing

License: (Apache-2.0 AND MIT)


olizilla:
@gmasgras is making it so. We can merge this when we see the signal flare ✨

alanshaw (approved):

PR #134, opened 2021-07-22
fix: add missing dep and add tests

olizilla:
and remove unused err-codes dep.

it is way too easy to screw up the deps for a package in a monorepo. ora is also a transitive dep of playwright-test which is why the command works locally despite not delcaring its own dependency on it.

License: (Apache-2.0 AND MIT)


alanshaw (approved):

olizilla:
its too easy to push changes wth missing deps. I am adding a test here. If it works I'll add it to the client too.

.github/workflows/w3.yml
```diff
@@ -0,0 +1,30 @@
...
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 16
+      # Only install the deps directly listed... dont do workspace magic.
+      # Running install from the root gets you all the deps for all the packages.
+      # which can easily mask missing deps if another modules also depends on it.
+      - run: npm install
+        working-directory: packages/w3
+      - run: npm test
+        working-directory: packages/w3
```
olizilla, line 30: @yusefnapora thanks for the nudge. You may be pleased to learn that we're now testing in CI to make sure that all the deps are declared!

yusefnapora: nice!

alanshaw (approved):

PR #135, opened 2021-07-23
refactor: change API module name

### LINKED ISSUE #86: Internal packages should use `@web3-storage` npm org prefix.
We're not planning on publishing the api, db, cron, website, etc as packages, but we should update the pkg.name fields to be prefixed with `@web3-storage` anyway. It's minor but someone could in theory squat/publish a module of that name, and npm-workspaces could weird out if run from a sub-package folder and install the bad other module instead.

alanshaw:
resolves #86

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2982,6 +2982,10 @@
         "web-streams-polyfill": "3.0.3"
       }
     },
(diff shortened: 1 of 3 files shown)


vasco-santos (approved):
LGTM

PR #142, opened 2021-07-23
docs: clarify onStoredChunk by specifying the default chunk size

alanshaw:


packages/client/src/lib/interface.ts
```diff
@@ -67,8 +67,9 @@ export type PutOptions = {
    */
   onRootCidReady?: (cid: CIDString) => void
   /**
-   * Callback called after each chunk of data has been uploaded. It is passed
-   * the chunk size in bytes.
+   * Callback called after each chunk of data has been uploaded. By default,
+   * data is split into chunks of around 10MB. It is passed the actual chunk
+   * size in bytes.
```
vasco-santos, line 72:
```suggestion
   * Callback called after each chunk of data has been uploaded. Data is split
   * into chunks of around 10MB. It is passed the actual chunk size in bytes.
```

nit: it is not really configurable at the moment to mention by default

vasco-santos (approved):

packages/client/src/lib/interface.ts
alanshaw: I know...it is the default though...you just can't change it.

PR #143, opened 2021-07-23
feat: user option for wrapWithDirectory

olizilla:
- add `wrapWithDirectory` option to client
- add `--no-wrap` flag to `w3
- add tests

License: (Apache-2.0 AND MIT)

### DIFF
--- a/packages/client/src/lib/interface.ts
+++ b/packages/client/src/lib/interface.ts
@@ -76,6 +76,18 @@ export type PutOptions = {
    */
   maxRetries?: number
   /**
+   * Should input files be wrapped with a directory. 
+   * Enabling it preserves the input filenames in DAG as they become
+   * directory entries in the generated wrapping dir. 
+   * 
+   * The trade off is your root CID will be that of the wrapping dir, 
+   * rather than the input file itself. 
+   * 
+   * so for a single file. e.g. `cat.png` its ipfs path would be 
+   * `<wrapping dir cid>/cat.png` rather than just `<cid for cat.png>`
+   */
+  wrapWithDirectory?: boolean
+  /**
    * Human readable name for this upload, for use in file listings.
    */
   name?: string
--- a/packages/client/src/lib/interface.ts
+++ b/packages/client/src/lib/interface.ts
@@ -72,18 +72,19 @@ export type PutOptions = {
    */
   onStoredChunk?: (size: number) => void
   /**
-   * Maximum times to retry a failed upload.
+   * Maximum times to retry a failed upload. Default: 5
    */
   maxRetries?: number
   /**
-   * Should input files be wrapped with a directory?
-   * Enabling it preserves the input filenames in DAG as they become
-   * directory entries in the generated wrapping dir. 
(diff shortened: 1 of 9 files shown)


olizilla:
This is one of those PRs where I am super happy to have a monorepo.

packages/client/src/lib/interface.ts
```diff
@@ -75,6 +75,18 @@ export type PutOptions = {
    * Maximum times to retry a failed upload.
    */
   maxRetries?: number
+  /**
+   * Should input files be wrapped with a directory. 
+   * Enabling it preserves the input filenames in DAG as they become
```
vasco-santos, line 80: It is enabled by default. Perhaps Disabling it?

I think we should mention the defaults used (same for maxRetries for example)

vasco-santos (approved):
LGTM
a nit regarding docs

packages/client/src/lib/interface.ts
```diff
@@ -75,6 +75,18 @@ export type PutOptions = {
    * Maximum times to retry a failed upload.
    */
   maxRetries?: number
+  /**
+   * Should input files be wrapped with a directory.
```
alanshaw, line 79:
```suggestion
   * Should input files be wrapped with a directory?
```

packages/client/src/lib/interface.ts
```diff
@@ -75,6 +75,18 @@ export type PutOptions = {
...
   maxRetries?: number
+  /**
+   * Should input files be wrapped with a directory. 
+   * Enabling it preserves the input filenames in DAG as they become
+   * directory entries in the generated wrapping dir. 
+   * 
+   * The trade off is your root CID will be that of the wrapping dir, 
+   * rather than the input file itself. 
+   * 
+   * so for a single file. e.g. `cat.png` its ipfs path would be
```
alanshaw, line 86:
```suggestion
   * For a single file e.g. `cat.png` it's IPFS path would be 
```

packages/client/src/lib/interface.ts
```diff
@@ -75,6 +75,18 @@ export type PutOptions = {
...
+  /**
+   * Should input files be wrapped with a directory. 
+   * Enabling it preserves the input filenames in DAG as they become
+   * directory entries in the generated wrapping dir. 
+   * 
+   * The trade off is your root CID will be that of the wrapping dir, 
+   * rather than the input file itself. 
+   * 
+   * so for a single file. e.g. `cat.png` its ipfs path would be 
+   * `<wrapping dir cid>/cat.png` rather than just `<cid for cat.png>`
```
alanshaw, line 87:
```suggestion
   * `<wrapping dir cid>/cat.png` rather than just `<cid for cat.png>`
   *
   * Wrapping with a directory is required when passing multiple files
   * that do not share the same root.
```

alanshaw (changes requested):

olizilla:
@vasco-santos you like now. @alanshaw Thanks, fixes applied.

vasco-santos (approved):

PR #158, opened 2021-07-26
chore: max width for code sample box

olizilla:
Constrain the code-sample max width, to match the grid above it.

**after**
Screenshot 2021-07-26 at 09 37 13

License: (Apache-2.0 AND MIT)

### DIFF
--- a/packages/website/pages/index.js
+++ b/packages/website/pages/index.js
@@ -133,7 +133,7 @@ retrieveFiles()`
           </p>
         </div>
       </div>
-      <div className="layout-margins pt-24">
+      <div className="md:layout-margins pt-24">
         <div className="relative flex flex-col xl:flex-row justify-between py-20 md:py-12 mt-20 bg-w3storage-blue-dark" style={{ borderTopLeftRadius: '6rem' }}>
           <Squares className="absolute top-14 left-16 hidden md:block" />
           <div className="text-white pl-12 pr-12 md:pl-32 md:pr-0 w-full max-w-none xl:max-w-lg mb-16 mr-0 xl:mr-10 xl:mb-0">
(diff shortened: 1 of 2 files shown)


packages/website/pages/index.js
```diff
@@ -133,7 +133,7 @@ retrieveFiles()`
           </p>
         </div>
       </div>
-      <div className="md:px-12 pt-24">
+      <div className="layout-margins pt-24">
```
alanshaw, line 136: I switched to `md:px-12` this so that on mobile the blue box reaches the edges of the screen to give this more space.

i.e. it was like this:



...and I switched to this:



We need a `md:layout-margins`, but it's a custom class so need to add to `global.css`.

olizilla:
I'm not focused on the website, I'm not going to add more changes to this PR this week. @jdiogopeixoto can you adopt this or submit an alternative fix please.

rafaelramalho19:
I'll assign this to me since João's on holidays this week

packages/website/pages/index.js
rafaelramalho19: Done

alanshaw (approved):
❤️

jnthnvctr:
Does this actually fix the max width? I'm still seeing this on the deployed PR: 

Think we want that entire blue area to not be wider than the four cards above it

image

rafaelramalho19:
> Does this actually fix the max width? I'm still seeing this on the IPFS Url below: https://bafybeihxaqa4ti5537emcqqzesg4e4peh4jz3cnblydbnqoyqgpychp5rq.ipfs.dweb.link/ )
> 
> Think we want that entire blue area to not be wider than the four cards above it
> 
> image

Good catch, I forgot to add a style above 1680px screens. Fixed now :)

PR #159, opened 2021-07-26
fix: throw on 404 from /get

olizilla:
- remove the spcial case to return undefined on a 404. We get from the gateway currently so any valid CID would either be returned succesfully or fail with a non 404 error code, so this 404 would not occur in the wild
- the plan is to have the api redirect to get from a dedicated gateway when we host the CID, and redirect to the public gateway where the user tries to get a CID we dont recognise.

License: (Apache-2.0 AND MIT)

### DIFF
--- a/packages/client/src/lib.js
+++ b/packages/client/src/lib.js
@@ -179,9 +179,6 @@ class Web3Storage {
       method: 'GET',
       headers: Web3Storage.headers(token)
     })
-    if (!res.ok) {
-      throw new Error(`${res.status} ${res.statusText}`)
-    }
     return toWeb3Response(res)
   }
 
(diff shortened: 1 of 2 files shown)


packages/client/src/lib.js
```diff
@@ -180,13 +180,7 @@ class Web3Storage {
...
     })
     if (!res.ok) {
-      // TODO: I'm assuming that an error for "CID isn't there (yet)" would be unergonomic. Need to verify.
-      // I'm thinking null means, nope, not yet, no can has. Anything else is _AN ERROR_
-      if (res.status === 404) {
-        return null
-      } else {
-        throw new Error(`${res.status} ${res.statusText}`)
-      }
+      throw new Error(`${res.status} ${res.statusText}`)
```
alanshaw, line 183: Ah I didn't realise it threw here for non-2xx status. This means everyone has to `try`/`catch` _as well as_ inspect `res.ok`. Conversely, `fetch` **never** throws so you only ever have to check `res.ok`. I vote we match that behaviour.

```suggestion
```

vasco-santos: I agree with @alanshaw 's point

olizilla:
Agreed, have changed so that `.files()` and `.unixFsIterator()` fail early with a helpful message if the user is not checking for `{ ok: false }` and calling them regardless.

alanshaw (approved):

olizilla:
cc: @johndmulhausen @yusefnapora note, this PR changes `client.get('cid')` so that it does not throw, not on 404, not on server error. The caller should check ther `ok` property of the response object to determine if it worked and it is safe to extract the `.files`

PR #160, opened 2021-07-26
chore: files page ui tweaks

rafaelramalho19:
Fixes: 
-  Files timestamp has different formats
- Files page's checkboxes were not keyboard accessible


alanshaw (approved):

jnthnvctr:
@rafaelramalho19  is this live on staging? i just tested and I think we might be parsing oddly still: https://github.com/web3-storage/web3.storage/issues/164

PR #170, opened 2021-07-26
fix: make unixfs importer on ipfs-car use same defaults as lotus

### LINKED ISSUE #173: Client chunking default params
We're considering a change to the default chunking parameters that will more closely reflect the defaults chosen by Lotus:

Current IPFS defaults:

* Chunk size: `262144`
* Max links: `174`

[Lotus defaults](https://github.com/filecoin-project/lotus/blob/bf468762b619c0cf018746a23d99f33a4f1026a9/build/params_shared_vals.go#L21-L22):

* Chunk size: `1048576`
* Max links: `1024`

Why?

* :green_circle: Flatter DAGs
* :green_circle: Fewer things to announce to the DHT
* 🟡  ~~Same CIDs for the same content vs Lotus~~ different hash functions (see below)
* 🟡 Different CIDs for the same content vs IPFS (although IPFS still stuck in v0 so we have that problem anyway since we use v1)
* :red_circle: Less de-duping

UPDATE:

Thanks to @vasco we know that [blake2js](https://www.npmjs.com/package/blakejs) is way slower than sha256. It is not unsurprising because sha256 uses native webcrypto in the browser and Node.js crypto module. There is an npm module with native bindings for Node.js and a module with a WASM version of blake as well. The work to integrate and test this would be huge and there's no benchmark data for either module so it might not even be faster than the native sha256 versions. Unless there are objections I vote we stick with sha256 hashing for now but switch chunk size and max children.

vasco-santos:
This makes `ipfs-car` packing with a `maxChunkSize` of  1048576  and a `maxChildrenPerNode` 1024.

We will keep using `sha256` hasher instead of [@multiformats/blake2](https://www.npmjs.com/package/@multiformats/blake2) given it is way slower (a minimal set of tests showed ~6x slower)

Needs:

- [x] https://github.com/web3-storage/ipfs-car/pull/67

resolves #173

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2185,9 +2185,9 @@
       "integrity": "sha1-p3c2C1s5oaLlEG+OhY8v0tBgxXA="
     },
     "node_modules/@rollup/plugin-commonjs": {
-      "version": "19.0.1",
-      "resolved": "https://registry.npmjs.org/@rollup/plugin-commonjs/-/plugin-commonjs-19.0.1.tgz",
-      "integrity": "sha512-bRrPTIAsWw2LmEspEMvV9f+7N7CEQgZCj2Zi1F0e0P3+/tbjQaSNNVVRSRWVhuDagp8yjK5kbIut8KTPsseRhg==",
+      "version": "19.0.2",
+      "resolved": "https://registry.npmjs.org/@rollup/plugin-commonjs/-/plugin-commonjs-19.0.2.tgz",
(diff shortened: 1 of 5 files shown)


alanshaw (approved):

olizilla:
@hugomrdias any idea why this PR was flagged as `release-please as a breaking change across multiple packages? i can't see why it's a breaking change at all.

PR #180, opened 2021-07-26
chore: fix pagination in the files page

rafaelramalho19:
- Fixes pagination for the files page (only shows next button if there's actually more items to show)

I feel like I cheated the system, but I'm fetching 25 items and only showing 24 in the UI, and only if we have 25+ items left, it shows the "next"

Related to https://github.com/web3-storage/web3.storage/issues/178

### DIFF
--- a/packages/website/pages/files.js
+++ b/packages/website/pages/files.js
@@ -136,7 +136,6 @@ export default function Files({ user }) {
   const initialFiles = [];
 
   const [selectedFiles, setSelectedFiles] = useState(/** @type string[] */ initialFiles)
-  const [isNextDisabled, setNextDisabled] = useState('')
   const [size] = useState(25)
   const [befores, setBefores] = useState([new Date().toISOString()])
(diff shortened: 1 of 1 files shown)


alanshaw:
You could fetch 26 and then still show a nice round 25 in the UI...

jnthnvctr:
I might be missing something - in the UI i thought I counted 25? 

I think what was happening is you show the button when theres 25 items, but the button doesn't take you anywhere (so its disabled).

jnthnvctr (approved):

rafaelramalho19:
> You could fetch 26 and then still show a nice round 25 in the UI...

My bad in the explanation, 

> You could fetch 26 and then still show a nice round 25 in the UI...

It does show 25 now, I changed the variable to be `25 + 1` so it's easier to realize in the future that the size requires 1 extra element to be fetched

alanshaw (approved):

Issue #182, opened 2021-07-26
About page UI / UX Update Requests


jnthnvctr:
1. Sentence case "Storage Term and Data Requests" -> "Storage term and data requests
2. Copy update for the about page (you can just copy and paste this wholesale): https://docs.google.com/document/d/1pqmmJwed6zfzG3FPjLGH99yrRlSFBZ8RS2N5xRY8dos/edit?disco=AAAANgiZMB4

Includes fixes to: 
a) A line to explain how this service is free
b) Miners -> Storage providers
c) Sentence casing in the above (1) 
d) A URL addition for cluster

jnthnvctr:
- [] Copy update on the About page using the linked google doc
- [] Sentence case for "Storage Term and Data Requests"

PR #185, opened 2021-07-26
chore: change about page's copy

rafaelramalho19:
Fixes. https://github.com/web3-storage/web3.storage/issues/182

### DIFF
--- a/packages/website/content/about.md
+++ b/packages/website/content/about.md
@@ -1,4 +1,4 @@
-### [What is Web3.Storage?](#what-is-web3storage)
+### What is Web3.Storage?
 
 Web3.Storage is a service to make building on top of Filecoin as simple as possible - giving the developers the power of open, distributed networks via a friendly JS client library. Behind the scenes, Web3.Storage is backed by Filecoin and makes content available via IPFS leveraging the unique properties of each network.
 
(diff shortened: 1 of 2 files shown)


jnthnvctr:
This looks good! Would we be able to make one other slight tweak? @terichadbourne  pointed out that there is a way we can enable Anchor links to each of the headers (e.g. Methodology) without having them actually be clickable links like Terms of Service. Would we be able to change the headers to do that? 

We definitely will want to be able to link to specific headers down the line, but we may not need to click on the link to just scroll down the page. 

Apologies if I'd described that ask poorly before!

jnthnvctr (commented):
Can we tweak the headers such that they aren't clickable, but are anchor links still?

alanshaw:
@jnthnvctr they should be clickable, people often click on headers and then copy the URL for sharing that part of the page. They should maybe not be underlined?

rafaelramalho19:
I remove the linking part so far, tell me if you're satisfied with the current solution in the PR or if you wanna go with Alan's advice

jnthnvctr:
Yeah thanks for this tweak - I just played with it - I think Mr. Shaw is
correct here

rafaelramalho19:
Done :)

alanshaw (approved):

PR #192, opened 2021-07-27
fix: query cluster until we get a preferable pin status

alanshaw:
After initial add, Cluster will usually report "Unpinned" when queried immediately. This PR polls cluster for status after adding until a preferable pin status is obtained from at least one of the nodes i.e. PinQueued, Pinning or Pinned.

The UI now shows the best pin status we have across all pins.

### DIFF
--- a/packages/api/src/car.js
+++ b/packages/api/src/car.js
@@ -28,10 +28,12 @@ const UPDATE_DAG_SIZE = gql`
   }
 `
 
-// Duration between status check poll.
+// Duration between status check polls in ms.
 const PIN_STATUS_CHECK_INTERVAL = 50
+// Max time in ms to spend polling for an OK status.
+const MAX_PIN_STATUS_CHECK_TIME = 5000
 // Pin statuses considered OK.
-const PIN_OK_STATUS = ['PinQueued', 'Pinning', 'Pinned']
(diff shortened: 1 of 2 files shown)


packages/api/src/car.js
```diff
@@ -109,16 +114,22 @@ export async function carPost (request, env, ctx) {
     local: blob.size > LOCAL_ADD_THRESHOLD
   })
 
+  /** @type {ReturnType<toPins>} */
+  let pins
   // Retrieve current pin status and info about the nodes pinning the content.
-  const { peerMap } = await env.cluster.status(cid)
+  // Keep querying Cluster until one of the nodes reports something other than
+  // Unpinned i.e. PinQueued or Pinning or Pinned.
+  while (true) {
```
vasco-santos, line 122: Can we become prone to Timeouts from CF with this?

I think we can probably add a `p-timeout` with a reasonable timeout for pooling. This will mean we might still present the state we are trying to fix to the user, but at least we will not throw a timeout error when the file is already stored properly

vasco-santos (commented):
Overall the solution is fine for me. But I have some concerns on the possible timeouts in the worker

vasco-santos (approved):
LGTM

PR #193, opened 2021-07-27
feat: show enqueuing uploads

alanshaw:
Displays something in the table when there is no deal information to see yet.

Screenshot 2021-07-27 at 12 05 36

### DIFF
--- a/packages/website/pages/files.js
+++ b/packages/website/pages/files.js
@@ -96,8 +96,16 @@ const UploadItem = ({ upload, index, toggle, selectedFiles }) => {
   const queuedDeals = upload.deals.filter(d => d.status === 'Queued')
   if (queuedDeals.length) {
     deals.push(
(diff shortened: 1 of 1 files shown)


rafaelramalho19 (approved):

rafaelramalho19:
LGTM, but in the future we should have a tooltip component that lets mobile users also see this information

olizilla:
`Enqueuing` is accurate but feels awkward. `Queuing` feels a little more natural here. In general I'm assuming we want to err on the side of less formal language.

alanshaw:
Quick survey says that folks do not know what this word is, like, literally no idea it existed. I don't like it but I can live with it 😂. https://github.com/web3-storage/web3.storage/commit/3180cac67c54ccae9e7a6f611ccc3c8c85d7b12c

PR #196, opened 2021-07-27
chore: query changes

rafaelramalho19:
- Disables refetch on windows focus
- Adds a button to refresh the files page


alanshaw:
As discussed, I think it'll not be great to see the loading spinner randomly after 1 minute on the page. @jnthnvctr had a suggestion to add a refresh button instead which I think is a good idea.

rafaelramalho19:
Added the refresh button 👍

alanshaw (approved):

PR #197, opened 2021-07-27
feat!: track user storage

vasco-santos:
This PR adds the possibility of tracking the quantity of data stored by a given user.

- The user collection now has a `usedStorage` column where each time a file is added (Car or Car chunk or Non Car) the used storage is increased.
  - we will likely add as a follow up the storage capacity available for the user in the same collection
- A `findStorageByUser` query was added where the relevant data to track the user storage is returned (now only `usedStorage`)

We still need to decide on what to do in reverse, i.e decrease used storage on file remove. There are several things we need to take into account:

- We can simply not decrease the used storage, given the content will still exist and be available (in a way, this can be exploited, i.e users "getting rid" of the limits available by deleting but keeping the CID in their registry)
- Delete content added immediately before will likely result on not having the size available to decrease in the user used storage
  - The most reliable solution would be to have a new collection with Pending Delete Items (when no dagSize yet?) with UserRef and ContentRef. And a cron job that would handle these entries has they appear and the size is known.
    - This means the "progress bar" of available storage would not always be immediately updated, which can also lead into users to delete a file to be able to add more, but then they still can't...
  - We could consider ignoring decreases when size not yet available, but this makes things not consistent across the board and does not seem the ideal compromise in this case

I would like feedback on these, I like the cron job, but I feel that some UX edge cases might become a problem. A possible follow up solution that could be useful for the admin to grant more storage to a given user would be to have a new property in the User, where we would use the cron job to detect the deletes behind the scenes and be able to understand if users are deleting files, and grant more storage on request.


packages/api/src/index.js
```diff
@@ -26,6 +26,7 @@ router.post('/user/tokens', withCorsHeaders(withAuth(userTokensPost)))
 router.delete('/user/tokens/:id', withCorsHeaders(withAuth(userTokensDelete)))
 router.get('/user/uploads', withCorsHeaders(withAuth(userUploadsGet)))
 router.delete('/user/uploads/:cid', withCorsHeaders(withAuth(userUploadsDelete)))
+router.get('/user/storage', withCorsHeaders(withAuth(userStorageGet)))
```
alanshaw, line 29: I'm tempted to just make this into a more generic user account info endpoint (that only returns storage usage for now). So like `/user/account`. We may have more user account things to retrieve in the future (IDK maybe we want to retrieve a user's avatar or something). WDYT?

packages/api/src/user.js
```diff
@@ -183,6 +183,24 @@ export async function userTokensPost (request, env) {
...
 
+/**
+ * Retrieve user storage data.
+ *
+ * @param {AuthenticatedRequest} request
+ * @param {import('./env').Env} env
+ */
+export async function userStorageGet (request, env) {
+  const res = await env.db.query(gql`
+    query findStorageByUser($user: ID!) {
```
alanshaw, line 194: Just use the existing `findUserByID` that fauna generates?

packages/db/fauna/schema.graphql
```diff
@@ -26,6 +26,11 @@ type User {
   Creation date.
   """
   created: Time!
+
+  """
+  Used storage.
```
alanshaw, line 31:
```suggestion
  Used storage in bytes.
```

alanshaw (changes requested):

alanshaw:
Is the simpliest solution for launch to simply abort a delete if we don't know the content `dagSize`?

packages/api/src/index.js
vasco-santos: `/user/account` seems a good plan

alanshaw:
> > Is the simpliest solution for launch to simply abort a delete if we don't know the content dagSize?
> 
> by abort you mean, just not decrease the size when we don't know the dagSize and decrease if we know? Or you mean adding support for aborting on going uploads?

I mean in `deleteUserUpload` we `Abort` if the `Content`'s `dagSize` is `null`, otherwise we decrement the user's `usedStorage`.

packages/api/src/user.js
vasco-santos: I looked into the generated functions yesterday and I couldn't find it lol 🤦🏼‍♂️  But now I see that the place to find them is in the docs tab in graphql section. So yes, of course let's just use that one

vasco-santos:
ahh, that sounds good! We require the dagSize to accept a delete, which means data will be consistent.

packages/db/fauna/schema.graphql
```diff
@@ -398,6 +404,10 @@ type Mutation {
   incrementPinRequestAttempts(pinRequest: ID!): PinRequest! @resolver
 }
 
+type UserStorage {
+  usedStorage: Long
+}
```
alanshaw, line 409: Not needed?

alanshaw (changes requested):

packages/db/fauna/resources/Function/deleteUserUpload.js
```diff
@@ -35,20 +38,38 @@ const body = Query(
...
+                  Select('ref', Var('content')),
+                  true
+                ),
+                upload: Select(['ref'], Get(Var('uploadMatch')))
+              },
+              If(
+                IsNonEmpty(Var('uploadMatch')),
+                Do(
+                  // Update user storage size
+                  Subtract(
```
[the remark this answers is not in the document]
vasco-santos: lol, no comments 😅

packages/db/fauna/resources/Function/deleteUserUpload.js
```diff
@@ -35,20 +38,42 @@ const body = Query(
...
+            IsNumber(Var('dagSize')),
+            Let(
+              {
+                uploadMatch: Match(
+                  Index('upload_by_user_and_content'),
+                  Var('userRef'),
+                  Select('ref', Var('content')),
+                  true
+                ),
+                upload: Select(['ref'], Get(Var('uploadMatch')))
```
alanshaw, line 53: You can't do this here because the match could be empty.

```suggestion
                )
```

packages/db/fauna/resources/Function/deleteUserUpload.js
```diff
@@ -35,20 +38,42 @@ const body = Query(
...
+                    data: {
+                      usedStorage: Subtract(
+                        Select(['data', 'usedStorage'], Get(Var('userRef'))),
+                        Var('dagSize')
+                      )
+                    }
+                  }),
+                  // Flag upload as deleted
+                  Update(
+                    Var('upload'),
```
alanshaw, line 69:
```suggestion
                    Select(['ref'], Get(Var('uploadMatch'))),
```

alanshaw (changes requested):

alanshaw (approved):

PR #198, opened 2021-07-27
feat: list user uploads in client and `w3`

olizilla:
- add `client.list` method
- add `w3 list` command with `--json` for ndjson output, or `--cid` for just the roots.

**The default output is nice for human eyes**

```console
$ w3 list
  Content ID                                                  Name
⁂ bafkreihs2lcpdsqowqpako6t4g3axyowb52r6k7hkl5mz35g7g3cyla5rq Upload at 2021-07-26T16:30:26.193Z
⁂ bafybeicwehe4cx74f5gf2duol7g5kgqmkh4z7ivnqikgqgspckqazz6ygi Upload at 2021-07-26T16:29:40.492Z
⁂ bafybeib6y5ih3q7dandsu3yu3ijinsreqklsvyydb2io3heussg5gptcxy dance.svg
```

**Just the CIDs for robot bashing**

```console
$ w3 list --cid
bafkreihs2lcpdsqowqpako6t4g3axyowb52r6k7hkl5mz35g7g3cyla5rq
bafybeicwehe4cx74f5gf2duol7g5kgqmkh4z7ivnqikgqgspckqazz6ygi
bafybeib6y5ih3q7dandsu3yu3ijinsreqklsvyydb2io3heussg5gptcxy
```

**ndJSON, for expert robots**

```console
$ w3 list --json
{"name":"Upload at 2021-07-26T16:30:26.193Z","cid":"bafkreihs2lcpdsqowqpako6t4g3axyowb52r6k7hkl5mz35g7g3cyla5rq","created":"2021-07-26T16:30:26.993827Z","dagSize":41980,"pins":[{"status":"Pinned"},{"status":"Pinned"},{"status":"Pinned"}],"deals":[]}
{"name":"Upload at 2021-07-26T16:29:40.492Z","cid":"bafybeicwehe4cx74f5gf2duol7g5kgqmkh4z7ivnqikgqgspckqazz6ygi","created":"2021-07-26T16:29:41.879866Z","dagSize":42037,"pins":[{"status":"Pinned"},{"status":"Pinned"},{"status":"Pinned"}],"deals":[]}
{"name":"dance.svg","cid":"bafybeib6y5ih3q7dandsu3yu3ijinsreqklsvyydb2io3heussg5gptcxy","created":"2021-07-26T15:03:18.760760Z","dagSize":127028,"pins":[{"status":"Pinned"},{"status":"Pinned"},{"status":"Pinned"}],"deals":[]}
```

**ndJSON, it pipe to jq!**

```console
$ w3 list --json | jq
{
  "name": "Upload at 2021-07-26T16:30:26.193Z",
  "cid": "bafkreihs2lcpdsqowqpako6t4g3axyowb52r6k7hkl5mz35g7g3cyla5rq",
  "created": "2021-07-26T16:30:26.993827Z",
  "dagSize": 41980,
  "pins": [
    {
      "status": "Pinned"
    },
    {
      "status": "Pinned"
    },
    {
      "status": "Pinned"
    }
  ],
  "deals": []
}
{
  "name": "Upload at 2021-07-26T16:29:40.492Z",
  "cid": "bafybeicwehe4cx74f5gf2duol7g5kgqmkh4z7ivnqikgqgspckqazz6ygi",
  "created": "2021-07-26T16:29:41.879866Z",
  "dagSize": 42037,
  "pins": [
    {
      "status": "Pinned"
    },
    {
      "status": "Pinned"
    },
    {
      "status": "Pinned"
    }
  ],
  "deals": []
}
{
  "name": "dance.svg",
  "cid": "bafybeib6y5ih3q7dandsu3yu3ijinsreqklsvyydb2io3heussg5gptcxy",
  "created": "2021-07-26T15:03:18.760760Z",
  "dagSize": 127028,
  "pins": [
    {
      "status": "Pinned"
    },
    {
      "status": "Pinned"
    },
    {
      "status": "Pinned"
    }
  ],
  "deals": []
}

```

TODO: we're iterating on this api, and will likely simplify it so the `client.list` function is an async generator.

License: (Apache-2.0 AND MIT)


olizilla:
**what the fork is happening here githoob? is your monospace font _not monospace!?_

Screenshot 2021-07-27 at 15 20 19

olizilla:
i think it depends on the font stack. They are using a "monospace-ui" for code, and it may fall back to a non-monospace font for the asterism...

packages/w3/index.js
```diff
@@ -13,12 +14,12 @@ const config = new Conf({
 })
 
 /**
- * Get a new API client configured either from opts or config
+ * Get a the API client config
```
vasco-santos, line 17:
```suggestion
 * Get the API client config
```

packages/w3/index.js
```diff
@@ -92,6 +103,55 @@ export async function get (cid, opts) {
   await writeFiles(res.unixFsIterator(), opts.output)
 }
 
+/**
+ * Print out all the uploads in your account by data created
+ *
+ * @param {object} opts
```
vasco-santos, line 109:
```suggestion
 * @param {object} [opts]
```

Let's also make opts optional?

packages/w3/index.js
```diff
@@ -92,6 +103,55 @@ export async function get (cid, opts) {
...
+ * @param {string} [opts.api]
+ * @param {string} [opts.token]
+ * @param {number} [opts.size] number of results to return per page
+ * @param {string} [opts.before] list items uploaded before this iso date string
+ */
+export async function list (opts) {
```
vasco-santos, line 115:
```suggestion
export async function list (opts = {}) {
```

vasco-santos (commented):
🎉 this is rad!

olizilla:
@vasco-santos we're thinking to have `client.list` be a generator instead, as it's a bit rough to use the current impl directly as it just returns the fetch Response directly, what do you think? I'd probably pull the `paginator` into the client lib, and then use it as the guts of `client.list` (but it could be reused for other paginatable end points in the future.

packages/client/src/lib.js
```diff
@@ -215,6 +215,20 @@ class Web3Storage {
     return res.json()
   }
 
+  /**
+   * @param {Service} service
+   * @param {{before: string, size: number}} opts
+   * @returns {Promise<Response>}
+   */
+  static async list ({ endpoint, token }, { before = new Date().toISOString(), size = 25 }) {
+    const search = new URLSearchParams({ before, size: size.toString() })
```
alanshaw, line 224: Nit: I think unnecessary to to to toString.

```suggestion
    const search = new URLSearchParams({ before, size })
```

Screenshot 2021-07-27 at 15 45 35

packages/w3/bin.js
```diff
@@ -21,6 +21,12 @@ cli.command('put <path>')
   .option('--no-wrap', 'Dont wrap input files with a directory.')
   .action(put)
 
+cli.command('list')
+  .describe('list all the uploads in your account')
+  .option('--json', 'Format as newline delimted JSON')
+  .option('--cid', 'Only print the root cid per upload')
```
alanshaw, line 27:
```suggestion
  .option('--cid', 'Only print the root CID per upload')
```

alanshaw (approved):
I love it.

vasco-santos:
@olizilla I am all into async generators, and for things like the CLI they are awesome. 

However, they tend to be less friendly for the users for what I have seen in IPFS+libp2p world. So, probably we should add `client.list()` and `client.getList()` ? where `getList()` would iterate the loop for them? Wdyt?

alanshaw:
Maybe `list` and `listIterator` to echo the `unixfsIterator` function we have on the `Web3Response`?

Or just `list()` but with a `res.paginator()` on the response?

olizilla:
> Or just `list()` but with a `res.paginator()` on the response?
I'm not clear on how this one would work, as you'd have a response with the first page of results on.

> Maybe `list` and `listIterator` to echo the unixfsIterator function we have on the Web3Response?
Yes, can has this.

packages/client/src/lib.js
olizilla: oh yes it is very uneccessary. but guess what. `tsc` hates it.

alanshaw: ohhhh

olizilla:
moved the pagination logic to the client. Added `client.list()` and `client.listIterator()`. Added totals to output

```console
$ w3 list
  Content ID                                                   Name
⁂ bafybeihuerbawpgbe5elho6tm6nkub42mj3fnrxzj4natrtcgvxmltywbq  web3-storage-web3.storage-203-4d56a318
⁂ bafybeicy66ruousdueiy43nqtr65ayagan4jd5f4ciepziojhvsam2yq7i  web3-storage-web3.storage-202-8c448172
⁂ bafybeiczvktm24qpmy3vwh7iapzy57mvx4gfshjwvoo7zfhi4jwmnfvbhe  web3-storage-web3.storage-201-537ad2ea
...trim...
⁂ bafybeihhjor3lyyckkdm4vjx55yx2cj2rsv4yw3mvtjuskjugvb5a2n5da  testing add-to-web3
⁂ bafybeifspyokz6jte5xue4smlt4jmd7jzzl2xhwsdq4z3x57wksjrlle5i  Upload at 2021-07-19T15:01:26.411Z
⁂ bafybeicwehe4cx74f5gf2duol7g5kgqmkh4z7ivnqikgqgspckqazz6ygi  Upload at 2021-07-15T10:03:02.236Z
⁂ bafybeig3zgiasqh7bsmvc5cudkkeeh3qborbscua3eqqky7m7fbzgrb4qu  not-distributed.jpg
⁂ bafybeif3l7iexiezaj65twojeqqbdkj3c43tyupb2ocayos5s6yz5duoxy  missingno.jpg
  112 items – 225.3MB strored
```

packages/client/src/lib.js
```diff
@@ -215,6 +217,50 @@ class Web3Storage {
     return res.json()
   }
 
+  /**
+   * @param {Service} service
+   * @param {{before: string, size: number}} opts
+   * @returns {AsyncIterable<Upload>}
+   */
+  static async * listIterator (service, opts) {
```
alanshaw, line 225: We were careful to keep the API surface area small for the rest of the API so adding 3 methods for this feels like too many.

I'm starting to wonder if `list(options): AsyncIterable` is fine.

First page is just `await client.list().next()` or `for await (const page of client.list())` if you need to paginate through. You can also call `client.list({ before }).next()` with a different date to do your own pagination...

packages/client/src/lib.js
```diff
@@ -215,6 +217,50 @@ class Web3Storage {
...
+      const page = await res.json()
+      for (const upload of page) {
+        yield upload
+      }
+    }
+  }
+
+  /**
+   * @param {Service} service
+   * @param {{before: string, size: number}} opts
```
alanshaw, line 239: Size doesn't really make sense here as an option. We should just set it to the max size.

alanshaw: ...or limit the results to the size that is passed. Which might be kinda cool - you can specify any size even though the API has a size restriction.

packages/client/src/lib.js
```diff
@@ -215,6 +217,50 @@ class Web3Storage {
     return res.json()
   }
 
+  /**
+   * @param {Service} service
+   * @param {{before: string, size: number}} opts
```
alanshaw, line 222: We should really have a type for this option so we can explain that `before` is a date as an ISO string and `size` is the max page size, not the max results to retrieve.

packages/client/src/lib/interface.ts
```diff
@@ -194,3 +194,7 @@ export interface Status {
    */
   deals: Array<Deal>
 }
+
+export interface Upload extends Status {
+  name: string
+}
```
alanshaw, line 200: 👍 we'll use this type in the UI also.

alanshaw (changes requested):

packages/client/src/lib.js
olizilla: yeah, it did feel chonky to have 3 list methods. I think having 1 way to do a thing is better. Folks are going to want a method that gives them an array, but i think if the docs give them the code to copy/paste like

```json
const uploads = []
for await (const item of client.list()) {
  uploads.push(item)
}
```

it should be fine.

packages/client/src/lib.js
olizilla: I think maybe i should change the name of the param to `maxResults` so not to have `size` mean either 'page size' or 'max results'

olizilla:
@alanshaw feedback actionnnnd. please to eyes.
I shall assume good and add tests.

alanshaw (approved):
🚢

PR #201, opened 2021-07-27
refactor: remove multipart parsing code

vasco-santos:


packages/api/package.json
```diff
@@ -36,6 +36,7 @@
     "webpack-cli": "^4.7.2"
   },
   "dependencies": {
+    "@web3-storage/multipart-parser": "^1.0.0",
```
alanshaw, line 39: Nit: should probably be placed after `@web3-storage/db` in this list.

alanshaw (approved):

PR #205, opened 2021-07-27
fix: metrics immutable headers

vasco-santos:
In production, GET /metrics was failing due to immutable headers. Let's just add them when creating the response

### DIFF
--- a/packages/api/src/cors.js
+++ b/packages/api/src/cors.js
@@ -57,6 +57,9 @@ export function withCorsHeaders (handler) {
  * @returns {Response}
  */
 export function addCorsHeaders (request, response) {
+  // Clone the response so that it's no longer immutable
+  response = new Response(response.body)
+
   const origin = request.headers.get('origin')
   if (origin) {
     response.headers.set('Access-Control-Allow-Origin', origin)
--- a/packages/api/src/cors.js
+++ b/packages/api/src/cors.js
@@ -57,9 +57,8 @@ export function withCorsHeaders (handler) {
  * @returns {Response}
  */
 export function addCorsHeaders (request, response) {
-  // Clone the response so that it's no longer immutable
-  response = new Response(response.body)
-
+  // Clone the response so that it's no longer immutable (like if it comes from cache or fetch)
+  response = new Response(response.body, response)
   const origin = request.headers.get('origin')
   if (origin) {
     response.headers.set('Access-Control-Allow-Origin', origin)
--- a/packages/api/src/metrics.js
+++ b/packages/api/src/metrics.js
@@ -70,10 +70,13 @@ export async function metricsGet (request, env, ctx) {
     `web3storage_pins_status_failed_total ${findMetrics.pinsFailedTotal}`
   ].join('\n')
 
-  res = new Response(metrics)
+  res = new Response(metrics, {
+    headers: {
+      // cache metrics response
+      'Cache-Control': `public, max-age=${METRICS_CACHE_MAX_AGE}`
+    }
+  })
 
-  // cache metrics response
-  res.headers.set('Cache-Control', `public, max-age=${METRICS_CACHE_MAX_AGE}`)
   ctx.waitUntil(cache.put(request, res.clone()))
 
   return res


alanshaw (approved):

olizilla:
Does this fix the issue? I was seeing behaviour where the first request succeeded, but then subsequent requests failed with the immutable headers error... my guess is that the response was served from the cache, but then we try and add CORS headers to it, which fails, becuase the cached response headers are immutable.

vasco-santos:
> Does this fix the issue?  I was seeing behaviour where the first request succeeded, but then subsequent requests failed with the immutable headers error.

I don't know, this does not happen to me locally. Does the same happen with `get`? 

I can also do a close of the request in the cors side of things

olizilla:
Yes, i have seen this error with `GET /car/:cid` too. It was on my list of things to explore when i get a moment.

packages/api/src/cors.js
```diff
@@ -57,6 +57,9 @@ export function withCorsHeaders (handler) {
  * @returns {Response}
  */
 export function addCorsHeaders (request, response) {
+  // Clone the response so that it's no longer immutable
+  response = new Response(response.body)
```
alanshaw, line 61: This throws away any headers set on the response already, including the status code.

alanshaw (changes requested):

olizilla (commented):
would be good to figure out what the cost of doing this is? This could be doubling the memory usage of every request!

I'd prefer to find another way to wire in cors headers to responses... ideally any responses we are going to cache should already have the cors headers set... maybe we need to pass through a `ctx.headers` so we can set cors headers first, and then merge them in when creating the response.

packages/api/src/cors.js
vasco-santos: yes, I am still testing this

vasco-santos:
> I'd prefer to find another way to wire in cors headers to responses... ideally any responses we are going to cache should already have the cors headers set... maybe we need to pass through a ctx.headers so we can set cors headers first, and then merge them in when creating the response.

Making `ctx.headers` does not seem the best solution to me. We would have logic replicated with the `addCorsHeaders` function. I think the cleanest solution by your suggestion is to simply leverage the `addCorsHeaders` function before caching. The `addCorsHeaders` will be called again in these functions, but it does not seem impactful.

Let me know your thouughts @olizilla @alanshaw just pushed this

olizilla:
My suggestion was broader... more like "lets refactor how we set CORS headers on things", and `ctx.headers` where we have 1 function that adds CORS headers to that property before any request, and contains the list of routes that we want CORS on inside it was where my head is at... like we'd replace the `addCorsHeaders` with a more middleware approach.

The other route that @alanshaw was in favour of was to have each handler import the `addCorsHeaders` function so it could apply it and be more self contained. I like that too, but mildly prefer having a middleware-ish implementation with an internal list of routes to apply to so you can see in one place all the routes that will have CORS headers. That said, if every handler would then be required to check for ctx.headers... then that's probably worse, as their is no guarantee that the handler code will remember to check it.... ok i am sold... i'm thinking that handlers should import an `addCorsHeaders` function that they choose when to apply.

vasco-santos:
I also thought on this, but then thought, should we require each handler to deal with this inside it, just because of the two handlers we need to cache responses? And I thought, it would fit better to keep the middleware as is.

I also prefer the middleware-ish implementation as this is contained in the routes. Special situation (like GET metrics GET car) can deal with this differently rather than making everyone need to handle what they could not need to, which means less code in the handlers logic

packages/api/src/metrics.js
```diff
@@ -2,6 +2,7 @@
 
 import { gql } from '@web3-storage/db'
 import { METRICS_CACHE_MAX_AGE } from './constants.js'
+import { addCorsHeaders } from './cors.js'
```
alanshaw, line 5: Need to remove the wrapping from `index.js`

alanshaw (changes requested):

alanshaw:
I'm currently worried about caching the CORS header for `Access-Control-Allow-Origin` since it gets set to the request origin and I don't know if there are any smarts in Cloudflare to change it automatically.

vasco-santos:
> I'm currently worried about caching the CORS header for Access-Control-Allow-Origin since it gets set to the request origin and I don't know if there are any smarts in Cloudflare to change it automatically.

@olizilla was saying:

> would be good to figure out what the cost of doing this is? This could be doubling the memory usage of every request!

I am not sure on the implications of re-creating the request and send it.

Maybe the compromise solution is:

```js
handler() {
  let body = await cache.match(request)

  if (res) {
    return return new Response(body, {
      headers: {
        // cache metrics response
        'Cache-Control': `public, max-age=${METRICS_CACHE_MAX_AGE}`
      }
    })
  }

  // ... Logic to get content from DB/Network

  ctx.waitUntil(cache.put(request, body))
  return new Response(body, {
    headers: {
      // cache metrics response
      'Cache-Control': `public, max-age=${METRICS_CACHE_MAX_AGE}`
    }
  })
}
```

This way, we can avoid the need to have handlers adding CORS and go back to the previous solution. In addition, we will not need to re-create the response as we are not storing it.

wdyt?

alanshaw:
You have to put a response object in the cache: https://developers.cloudflare.com/workers/runtime-apis/cache#put

vasco-santos:
So, either keep this solution, or reconstruct the responde when getting it from cache

olizilla:
Note: we need a sensible fix for how we apply CORS headers to cached requests this week, if we are launching press releaseses shortly. You currently get this error on metrics **and** `get` requests.

olizilla:
Not sure if this is official, but sounds plausible... `clone` is cheap as it just creates another reader on the same underlying data.
>Since a response.body readable-stream can only be read ONCE (its copy by ref, not copy by val) and then it is “disturbed” and can’t be read again (in new Response or whenever the promise is await-ed later on, explicitly by you or by the runtime). clone() allows you to create TWO body read streams that both start at “offset 0” from 1 origin response.
>
> “Resonse”/fetch object are constructed/resolve after the “\n\n” of the HTTP response header, the body aka readable-stream object is NOT yet delivered to CF/CFW instance, its still in transit, even though the headers are fully delivered. the body readable-stream theoretically resolves AFTER your CFW “exits” by calling respondWith() outside the V8 sandbox. clone lets you duplicate a future readable-stream body to be read twice.
– https://community.cloudflare.com/t/whats-the-point-of-response-clone/216456/2

olizilla:
I think the minimal change here would be to revert to pretty much how you had it before @vasco-santos... in https://github.com/web3-storage/web3.storage/pull/205/commits/f96d97083c847e425f78b3ca873707e4ae4336eb

but passing the response into the new Response so the old headers are copied over

```js
export function addCorsHeaders (request, response) {
  // Clone the response so that it's no longer immutable
  response = new Response(response.body, request)
  ...
```  

see: https://developers.cloudflare.com/workers/examples/alter-headers

olizilla:
@vasco-santos i'm going to adopt this PR and do the thing suggested above!

olizilla:
I'm not sure local testing will prove this works, but it hasn't obviously broken anything, and the tests still pass so :shipit: ?

olizilla:
@alanshaw plz to 👀

alanshaw (approved):

PR #216, opened 2021-07-28
chore: update account page UI

rafaelramalho19:
- Fixes tokens table's width
- Removes the Getting Started section when you already have tokens in your account


alanshaw (approved):

jnthnvctr:
@rafaelramalho19  - can we confirm that on the About page we've also swapped out where it references Profile (if you search its in the Storage term and data limits section). 

Also if you're in there coudl we do the terms of service underlining (or add that icon teri mentioned)

PR #218, opened 2021-07-28
chore: stop file deletion when its still uploading

rafaelramalho19:
- Disables the checkbox when a file is still being uploaded
image

### DIFF
--- a/packages/website/pages/files.js
+++ b/packages/website/pages/files.js
@@ -249,7 +249,9 @@ export default function Files({ user }) {
   }
 
   const toggleAll = () => {
-    selectedFiles.length >= 1 ? setSelectedFiles([]) : setSelectedFiles(uploads.map(u => u.cid))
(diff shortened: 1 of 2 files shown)


packages/website/pages/files.js
```diff
@@ -263,7 +263,7 @@ export default function Files({ user }) {
         <tr>
           { uploads.length > 0 && (
           <th className="w-8">
-            <Checkbox className="mr-2" checked={selectedFiles.length === uploads.length} onChange={toggleAll} />
+            <Checkbox className="mr-2" checked={selectedFiles.length === uploads.length} disabled={uploads.every(upload => !upload.dagSize)} onChange={toggleAll} />
```
alanshaw, line 266: We should still be able to toggle all the ones that can be deleted.

alanshaw (changes requested):

alanshaw (approved):

PR #219, opened 2021-07-28
fix: validate root car

### LINKED ISSUE #191: Client should verify root CID
The client should verify the CID returned from the service matches the root CID it generated.

vasco-santos:
Resolves #191


packages/client/src/lib.js
```diff
@@ -148,11 +149,15 @@ class Web3Storage {
...
 
-              if (request.ok) {
-                return res.cid
-              } else {
+              if (!request.ok) {
                 throw new Error(res.message)
               }
+
+              if (res.cid !== carRoot) {
+                throw new Error(`Service ${endpoint} is compromised and CID received is not the same as generated by the client`)
```
alanshaw, line 157: Wow haha I'd maybe tone this down a little so nobody panics. Maybe something more like "root CID mismatch, expected: x, received: y"

vasco-santos:
```suggestion
                throw new Error(`root CID mismatch, expected: ${carRoot}, received: ${res.cid}`)
```

packages/client/test/put.spec.js
```diff
@@ -41,6 +41,16 @@ describe('put', () => {
     }
   })
 
+  it('errors with a received CID that is not the same as generated in the client', async function () {
+    const client = new Web3Storage({ token, endpoint })
+    try {
+      await client.put([new File(['test-compromised-service'], 'file.txt')], { maxRetries: 1 })
+      assert.unreachable('should have thrown')
+    } catch (err) {
+      assert.match(err.message, /is compromised/)
```
vasco-santos, line 50:
```suggestion
      assert.match(err.message, /root CID mismatch/)
```

alanshaw (approved):

PR #222, opened 2021-07-28
docs: improve browser example code and add list demo

olizilla:
it look fun now! **and** it demo `put` and `list`!

Screenshot 2021-07-28 at 21 58 23

depends on #198

License: (Apache-2.0 AND MIT)


vasco-santos (approved):
LGTM ❤️

packages/client/examples/browser/main.js
```diff
@@ -1,17 +1,48 @@
...
+  showMessage(`> ✅ web3.storage now hosting ${cid}`)
+  showLink(`https://dweb.link/ipfs/${cid}`)
 
-  const text = document.createTextNode(`Content added with CID: ${cid}`)
-  cidP.appendChild(text)
+  showMessage('> 📡 fetching the list of all unique uploads on this account')
+  let totalBytes = 0
+  for await (const upload of client.list()) {
+    showMessage(`> 📄 ${upload.cid}  ${upload.name}`)
+    totalBytes += upload.dagSize
```
alanshaw, line 32:
```suggestion
    totalBytes += upload.dagSize || 0
```

packages/client/examples/browser/package.json
```diff
@@ -9,6 +10,6 @@
     "vite": "^2.3.7"
   },
   "dependencies": {
-    "web3.storage": "^1.0.0"
+    "web3.storage": "file:../.."
```
alanshaw, line 13: When published can it please depend on a real version so users don't have to npm install in the root as well as here?

packages/client/examples/browser/main.js
olizilla: not need, as dagSize is either a number or `null`
```console
> x = 0
0
> x += null
0
```

but
```console
> x += undefined
NaN
```
so will do for best safety

packages/client/examples/browser/package.json
olizilla: yes, this was just for testing until the latest module release, which now!

yusefnapora:
This is awesome! Nice work @olizilla

olizilla:
the thing to store things is now stored on the thing!
https://bafybeigxve4jcuc5o5cq3sjifgw5ukzteb23xsqalzjltavk5bzn4bi3w4.ipfs.dweb.link/

alanshaw (approved):

PR #234, opened 2021-07-29
fix: account page ui makeover

rafaelramalho19:
- Makes the account page a beautiful pixel-perfect page by matching the figma specs
image

- Account page now fetches storage capacity and shows it accordingly (with a sexy animation)

https://user-images.githubusercontent.com/24696635/127491981-a16b1b7e-8439-4a68-86b6-5bfc00969989.mov

- Fixes token cards with huge titles 
image

### DIFF
--- a/packages/api/src/user.js
+++ b/packages/api/src/user.js
@@ -218,7 +218,7 @@ export async function userTokensGet (request, env) {
           name
           secret
           created
-          uploads {
+          uploads(_size: 1) {
             data {
               _id
             }
--- a/packages/api/src/user.js
+++ b/packages/api/src/user.js
@@ -228,6 +228,12 @@ export async function userTokensGet (request, env) {
     }
   `, { user: request.auth.user._id })
 
+  res.findAuthTokensByUser.data = res.findAuthTokensByUser.data.map(t => {
+    t.hasUploads = Boolean(t.uploads.data.length)
+    delete t.uploads
+    return t
+  })
+
   return new JSONResponse(res.findAuthTokensByUser.data)
 }
 
--- a/packages/api/src/user.js
+++ b/packages/api/src/user.js
@@ -218,6 +218,11 @@ export async function userTokensGet (request, env) {
           name
           secret
           created
+          uploads {
+            data {
+              _id
+            }
+          }
         }
       }
     }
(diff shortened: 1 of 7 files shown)


jnthnvctr:
@rafaelramalho19 we should probably only hide the first card ("Create a New API Key") once a token is created. We should keep the "Start Buildling" one (if we're going to have a condition to hide that, it should be when someone has made an upload from one of the api keys)

rafaelramalho19:
> @rafaelramalho19 we should probably only hide the first card ("Create a New API Key") once a token is created. We should keep the "Start Buildling" one (if we're going to have a condition to hide that, it should be when someone has made an upload from one of the api keys)

@jnthnvctr Added that :)

alanshaw (changes requested):

packages/website/pages/account.js
```diff
@@ -40,6 +94,9 @@ export default function Account({ user }) {
     return () => clearTimeout(timer)
   }, [copied])
 
+  const isLoaded = !isLoading && !isFetching
+
+  const hasUsedTokensToUploadBefore = tokens.some(t => Object.entries(t.uploads?.data || {}).length > 0)
```
alanshaw, line 99: `t.uploads?.data` will always be an array
```suggestion
  const hasUsedTokensToUploadBefore = tokens.some(t => t.uploads.data.length > 0)
```

packages/api/src/user.js
```diff
@@ -218,6 +218,11 @@ export async function userTokensGet (request, env) {
           name
           secret
           created
+          uploads(_size: 1) {
+            data {
+              _id
+            }
+          }
```
alanshaw, line 225: Instead of returning this array of a single upload with an ID, why don't we transform it into a boolean property that's more useful to the UI?:

```js
res.findAuthTokensByUser.data = res.findAuthTokensByUser.data.map(t => {
  t.hasUploads = Boolean(t.uploads.data.length)
  delete t.uploads
  return t
})
```

alanshaw (approved):
This is good to merge as-is but it would be nice to return something from the API that is a bit easier for the UI to use (i.e. a boolean property rather than an array of 1 item). It'll also a bit less cryptic to devs.

packages/api/src/user.js
rafaelramalho19: Good suggestion, added :)

alanshaw (approved):

PR #236, opened 2021-07-29
fix: show vertical-lines illustration again

rafaelramalho19:
I screwed up the vertical-lines illustration a while ago 😢


alanshaw (changes requested):
Covers the text on about page:
Screenshot 2021-07-30 at 10 01 02
Not at edge of screen on new token and temrs page:
Screenshot 2021-07-30 at 10 03 05
Screenshot 2021-07-30 at 10 03 16

rafaelramalho19:
Fixed it, haven't changed in the account.js because that's done in https://github.com/web3-storage/web3.storage/pull/234

alanshaw (changes requested):
The new token page is the same problem for me (others fixed though)

alanshaw (approved):

Issue #231, opened 2021-07-29
API tokens should not permit admin operations


alanshaw:
Our API tokens currently allow access to the whole API. In leu of proper scoping for tokens we should lock down certain endpoints to be accessible only with a Magic.link token.

Proposal:

* 🤲 `GET /car/:cid`
* 🤲 `HEAD /car/:cid`
* 🔒 `POST /car`
* 🔒 `POST /upload`
* 🤲 `GET /status/:cid`
* 🤲 `POST /user/login`
* 👮 `GET /user/tokens`
* 👮 `POST /user/tokens`
* 👮 `DELETE /user/tokens/:id`
* 🔒 `GET /user/uploads`
* 👮 `DELETE /user/uploads/:cid`
* 👮 `GET /user/account`

🤲 = open 🔒 = authenticated 👮 = admin

The 🔒 endpoints will be accessible with an API token or Magic.link token. The 👮 endpoints are only accessible from the website using a magic.link token.

dchoi27:
Can this be closed by #237?

PR #237, opened 2021-07-29
feat: authmoji to limit the actions an api token can perfrom

olizilla:
new authmoji! it's cool! It's almost exactly what was asked for in #231

```js
const auth = {
  '👮': handler => withCorsHeaders(withMagicToken(handler))
}

router.post('/user/login',          auth['🤲'](userLoginPost))
router.get('/status/:cid',          auth['🤲'](statusGet))
router.get('/car/:cid',             auth['🤲'](carGet))
router.head('/car/:cid',            auth['🤲'](carHead))

router.post('/car',                 auth['🔒'](carPost))
router.put('/car/:cid',             auth['🔒'](carPut))
router.post('/upload',              auth['🔒'](uploadPost))
router.get('/user/uploads',         auth['🔒'](userUploadsGet))

router.delete('/user/uploads/:cid', auth['👮'](userUploadsDelete))
router.get('/user/tokens',          auth['👮'](userTokensGet))
router.post('/user/tokens',         auth['👮'](userTokensPost))
router.delete('/user/tokens/:id',   auth['👮'](userTokensDelete))
router.get('/user/account',         auth['👮'](userAccountGet))
```

less cool is that we need some mechanism to bypass magic.link in testing so this PR adds a `DANGEROUSLY_BYPASS_MAGIC_AUTH` env var that is only set in the worker-globals.js test code, and no where near prod. It's not great, but it it would be difficult to set accidentally.

License: (Apache-2.0 AND MIT)


packages/api/src/auth.js
```diff
@@ -0,0 +1,121 @@
...
+ */
+export function withMagicToken (handler) {
+  /**
+   * @param {Request} request
+   * @param {import('./env').Env}
+   * @returns {Response}
+   */
+  return async (request, env, ctx) => {
+    const token = getTokenFromRequest(request, env)
+    if (isMagicToken(token, env)) {
```
alanshaw, line 26:
```suggestion
    if (await isMagicToken(token, env)) {
```

packages/api/src/auth.js
```diff
@@ -0,0 +1,121 @@
...
+   * @param {Request} request
+   * @param {import('./env').Env}
+   * @returns {Response}
+   */
+  return async (request, env, ctx) => {
+    const token = getTokenFromRequest(request, env)
+    if (!token) {
+      throw new NoTokenError()
+    }
+    if (isWeb3ApiToken(token, env)) {
```
alanshaw, line 57:
```suggestion
    if (await isWeb3ApiToken(token, env)) {
```

packages/api/src/auth.js
```diff
@@ -0,0 +1,121 @@
...
+      const authToken = await findWeb3ApiToken(token, env)
+      if (!authToken) {
+        throw new TokenNotFoundError()
+      }
+      request.auth = { authToken, user: authToken.user }
+      env.sentry && env.sentry.setUser(authToken.user)
+      return handler(request, env, ctx)
+    }
+    if (isMagicToken(token, env)) {
```
alanshaw, line 66:
```suggestion
    if (await isMagicToken(token, env)) {
```

alanshaw (commented):
Oops, no review requested, stopping! Leaving feedback so far just incase it's useful.

packages/api/src/auth.js
olizilla: almost! turns out `magic.token.validate` is sync, and will throw if it doesn't recognise it. I'm going to catch that error and return false.

olizilla:
oh no! 
> Coming soon! Admin-sdk support for testMode!
– https://magic.link/docs/introduction/test-mode

olizilla:
@alanshaw please to 👀  now it has tests! that pass!

alanshaw (approved):
LGTM

PR #239, opened 2021-07-29
fix: homepage styles according to ui revision

rafaelramalho19:
- Fix homepage styles according to UI revision from https://www.figma.com/file/826Zurcs6kgWSzQ35it966/web3.storage?node-id=2062%3A565


packages/website/components/navbar.js
```diff
@@ -24,31 +24,31 @@ export default function Navbar({ bgColor = '', user, isLoadingUser }) {
...
             <>
               <Link href="/files">
-                <a className="text-w3storage-purple font-bold no-underline hover:underline align-middle mr-6 md:mr-12">
+                <a className="text-sm text-w3storage-purple font-bold no-underline hover:underline align-middle mr-6 md:mr-12">
                   Files
                 </a>
               </Link>
               <Link href="/account">
-                <a className="text-w3storage-purple font-bold no-underline hover:underline align-middle mr-6 md:mr-12">
+                <a className="text-sm text-w3storage-purple font-bold no-underline hover:underline align-middle mr-6 md:mr-12">
```
alanshaw, line 51: Since these are smaller now can we please add some padding to make the click area bigger?

alanshaw (changes requested):
Changes LGTM, just a request for a bit better accessibility 🙏

packages/website/components/navbar.js
rafaelramalho19: Good catch, done :)

alanshaw (approved):
LGTM. I added a tweak to ensure the spacing was consistent between the menu items and the login/logout button.

PR #242, opened 2021-07-30
feat: restrict block size

alanshaw:
Adds a restriction to CAR uploads to ensure we don't store a block that cannot be transferred over the IPFS network.


ribasushi (approved):
✔️

packages/api/src/car.js
```diff
@@ -104,6 +104,9 @@ export async function carPost (request, env, ctx) {
   }
 
   const blob = await request.blob()
+  const bytes = new Uint8Array(await blob.arrayBuffer())
+  const reader = await CarReader.fromBytes(bytes)
+  const chunkSize = await getBlocksSize(reader)
```
olizilla, line 109: We could use a CarBlockIterator here and wrap request.body as a AsyncIterator to only unpack a block into mem at once?

PR #243, opened 2021-07-30
refactor: return deleted uploads from findUploadsCreatedAfter

alanshaw:
Switch to an index that includes deleted records.


ribasushi (approved):
perfect

PR #249, opened 2021-07-30
fix: update usedStorage for chunked uploads

alanshaw:
If an existing upload object is found we currently return it, without updating the user's `usedStorage` with the `chunkSize`.

It's more likely that this is a chunked upload and we need to add on the chunk size to the user's `usedStorage`.


vasco-santos (approved):
LGTM
Good catch

PR #293, opened 2021-08-03
Expose API for uploading CAR files

### LINKED ISSUE #288: Expose API for uploading CAR files
We just need to abstract the current `put()` method a little. Probably turn this into an `_` method https://github.com/web3-storage/web3.storage/blob/main/packages/client/src/lib.js#L131 and write a small function on top like “`putCAR()`” to call the `_` method.

terichadbourne:
WIP - when ready will close #288


terichadbourne:
NOT READY FOR REVIEW; just making the tests run to debug

terichadbourne:
@mikeal & I are unable to run tests locally, apparently because the connection isn't loading. Lots of messages such as the following when running `npm run test:esm` from inside the client directory:  
image

terichadbourne:
@Gozala Could you please confirm we've made the right change in `lib/interface.ts` and update the comment? 

We still need help debugging this but seem to have squashed at least a few of the errors that were there.

terichadbourne:
Tests are now passing for me locally but failing here on GitHub. 
image

@mikeal Can you confirm these are the right next steps on this one? 
[ ] Fix Typescript docs
[ ] Check for other test failures once Typescript is resolved
[ ] Add a small function like “putCAR()” to call the new "_put" method

terichadbourne:
@Gozala and I just fixed the type errors. Currently getting no errors locally for me with either `npm run build:tsc` or `npm run test:esm`.  We solved this by adding a new `_PutOptions` extending the old `PutOptions` rather than just reusing the latter, since `_put` needs more data passed in than `put` does. 

We've only set the types of things, not added the descriptors / documentation to `lib/interface.ts`, since neither of us is familiar enough with this codebase to describe them accurately.

terichadbourne:
TODO: 
- [ ] Update definitions in `lib/interface.ts` (someone familiar with the codebase (@mikeal is that you or do we need @alanshaw?
- [ ] Add a small function like “putCAR()” to call the new "_put" method (Mikeal and I have time scheduled to do this bit tomorrow

terichadbourne:
@mikeal and I put our heads together to get this working today, in the sense that all the esm tests pass, including the new one I created for the new `putCar`.  

We have one error remaining from Typescript, and neither of us is super familiar with how to make it happy. Could @alanshaw or @Gozala please take a look at [this line](https://github.com/web3-storage/web3.storage/pull/293/files#diff-1d5466aa3a71f790007cfd7b5459f722d6e2c9e59ff53bdab52f3ac6e403b22aR146) and let me know what you'd suggest?

packages/client/src/lib.js
```diff
@@ -108,61 +140,52 @@ class Web3Storage {
...
-          content: f.stream()
-        })),
-        blockstore,
-        wrapWithDirectory,
-        maxChunkSize: 1048576,
-        maxChildrenPerNode: 1024
-      })
-      carRoot = root.toString()
+    const [ root ] = await car.getRoots()
+    const carRoot = root.toString()
```
alanshaw, line 146:
```suggestion
    const [ root ] = await car.getRoots()
    if (root == null) {
      throw new Error('missing root CID')
    }
    const carRoot = root.toString()
```

terichadbourne: @alanshaw Accepted your suggestion and now it has noticed that there's no test coverage for those lines. Can you think of a way to make a CID so poorly formed that it'll fail in a test? Not sure if your build process will allow those lines not to be covered in tests (and what I'm seeing is just a suggestion) or if it will put up a stink.

terichadbourne:
@mikeal @alanshaw I think we might be done(ish) with this! 

The only little loose ends I'm not sure how to tackle: 
- [ ] Missing test coverage for [this spot](https://github.com/web3-storage/web3.storage/pull/293/files#diff-1d5466aa3a71f790007cfd7b5459f722d6e2c9e59ff53bdab52f3ac6e403b22aR144-R147) where we test whether a CID is missing its root
- [ ] Tests for esm and tsc pass locally but not on GitHub 
image
- [ ] I'm unable to run the full test suite locally due to my NPM being out of whack (not a problem if tests come out successful on GH)
- [ ] Confirm it's okay that the request/response cycle runs many times when testing uploading big files (if not, you can revert the commit where I removed my console logs to quickly see this phenomenon for yourselves while testing) 

Please let me know what you recommend for next steps!

mikeal:
You should see multiple request/response cycles on large car files, that’s great! It means we can push arbitrarily large CAR files through Cloudflare by breaking them up the same way we do unixfs dags.

packages/client/src/lib.js
```diff
@@ -284,7 +302,53 @@ class Web3Storage {
   }
 
   /**
-   * Fetch the Content Addressed Archive by it's root CID.
+   * Uploads CAR files to web3.storage. Files are hashed in the client and uploaded as a single
+   * [Content Addressed Archive(CAR)](https://github.com/ipld/specs/blob/master/block-layer/content-addressable-archives.md).
```
alanshaw, line 306: Not hashed by the client here, the hashing and CAR construction needs to be done in the application.

```suggestion
   * Uploads a CAR ([Content Addressed Archive](https://github.com/ipld/specs/blob/master/block-layer/content-addressable-archives.md)) file to web3.storage.
```

packages/client/src/lib.js
```diff
@@ -284,7 +302,53 @@ class Web3Storage {
...
+
+   *    const inStream = fs.createReadStream('example.car')
+   *    // read and parse the entire stream in one go, this will cache the contents of
+   *    // the car in memory so is not suitable for large files.
+   *    const reader = await CarReader.fromIterable(inStream)
+   *    return reader
+   * }
+
+   * const car = await getCar()
+   * const cid = await putCar(car)
```
alanshaw, line 341:
```suggestion
   *
   * const car = await getCar()
   * const cid = await client.putCar(car)
```

packages/client/src/lib.js
```diff
@@ -198,7 +216,7 @@ class Web3Storage {
    */
   /* c8 ignore next 4 */
   static async delete ({ endpoint, token }, cid) {
-    console.log('Not deleteing', cid, endpoint, token)
+    console.log('Not deleting', cid, endpoint, token)
```
alanshaw, line 219: 👌

packages/client/src/lib.js
```diff
@@ -97,8 +99,38 @@ class Web3Storage {
...
+
+  /**
+   * @param {Service} service
+   * @param {_PutOptions} options
+   * @returns {Promise<CIDString>}
+   */
+  static async _put ({ endpoint, token }, {
+    onStoredChunk, car, name, blockstore,
+    maxRetries = MAX_PUT_RETRIES
+  }) {
```
alanshaw, line 130: Why not call this `putCar`? I'd also pull the car out of the options object and have it as a separate parameter since it's required - then the only difference with the instance method is the credentials, and this is consistent with other static methods.

```suggestion
  /**
   * @param {Service} service
   * @param {import('@ipld/car/api').CarReader} car
   * @param {PutCarOptions} options
   * @returns {Promise<CIDString>}
   */
  static async putCar ({ endpoint, token }, car, {
    onStoredChunk, name, blockstore,
    maxRetries = MAX_PUT_RETRIES
  }) {
```

packages/client/src/lib/interface.ts
```diff
@@ -112,6 +114,17 @@ export type PutOptions = {
...
+  /**
+   * CarReader interface from @ipld/car/api
+   */
+  car: CarReader
+  /**
+   * Blockstore
+   */
+  blockstore?: Blockstore
+}
+
```
alanshaw, line 127:
```suggestion
export type PutCarOptions = PutOptions & {
  /**
   * Blockstore
   */
  blockstore?: Blockstore
}

```

packages/client/src/lib/interface.ts
```diff
@@ -1,6 +1,8 @@
 import type { UnixFSEntry } from 'ipfs-car/unpack'
 import type { CID } from 'multiformats'
 export type { CID, UnixFSEntry }
+import type { MemoryBlockStore as Blockstore } from 'ipfs-car/blockstore/memory'
```
alanshaw, line 4: This doesn't have to be a `MemoryBlockStore` does it?

```suggestion
import type { Blockstore } from 'ipfs-car/blockstore'
```

alanshaw (changes requested):

packages/client/src/lib.js
```diff
@@ -284,7 +302,53 @@ class Web3Storage {
...
+
+   * const car = await getCar()
+   * const cid = await putCar(car)
+   * ```
+   * @param {CarReader} car
+   * @param {_PutOptions} [options]
+   */
+  putCar (car, options) {
+    return Web3Storage._put(this, { ...options, car })
+  }
```
alanshaw, line 348:
```suggestion
   * @param {import('@ipld/car/api').CarReader} car
   * @param {PutCarOptions} [options]
   */
  putCar (car, options) {
    return Web3Storage.putCar(this, car, options)
  }
```

packages/client/src/lib.js
```diff
@@ -39,6 +40,7 @@ const MAX_CHUNK_SIZE = 1024 * 1024 * 10 // chunk to ~10MB CARs
 /** @typedef { import('./lib/interface.js').Filelike } Filelike */
 /** @typedef { import('./lib/interface.js').CIDString} CIDString */
 /** @typedef { import('./lib/interface.js').PutOptions} PutOptions */
+/** @typedef { import('./lib/interface.js')._PutOptions} _PutOptions */
```
alanshaw, line 43:
```suggestion
/** @typedef { import('./lib/interface.js').PutCarOptions} PutCarOptions */
```

alanshaw (changes requested):
Apologies, I missed a couple of things.

packages/client/src/lib.js
terichadbourne: @alanshaw I understand why you're suggesting this name change, but could you help me understand why it would be okay to have both this renamed `status async putCar` function (defined here) and a `putCar` function (defined on line 346) in the same file? Do they not need different names from each other?

terichadbourne:
@alanshaw Thanks for the review. I committed your changes and then added a few extra little tweaks needed because of them. 

A couple of things still to be resolved that I'd appreciate your advice on: 
- [ ] [These lines](https://github.com/web3-storage/web3.storage/pull/293/files#diff-1d5466aa3a71f790007cfd7b5459f722d6e2c9e59ff53bdab52f3ac6e403b22aR146-R148)  that ensure a CID has root aren't covered in testing. Is there a way to set some sort of ignoring feature or do you want to address by creating a weird CID missing its root to fail the test? 
image
- [ ] Tests for esm and tsc pass locally but not on GitHub because "Readable.from is not available in the browser." I implemented a change of package from `stream` to `readable-stream` at @mikeal's suggestion yesterday (in [this commit](https://github.com/web3-storage/web3.storage/pull/293/commits/174b981eb83dfcf662fa64d5985e85429c9a4af7))  but it didn't help. What would you suggest there?

olizilla:
it would be great to have a test case for a CAR file with zero roots. We'd need to write a script with https://github.com/ipld/js-car to create such a thing. A CAR _could_ have zero or more roots, but it's currently unclear if we consider a zero root car a valid thing. see: https://ipld.io/specs/transport/car/carv1/#number-of-roots

**But** for the sake of this PR, we could just ignore that condition, and raise an issue to add a test case for zero root CAR files.

olizilla:
The `test:web` command is failling locally for me with the same error.
```
Error: Readable.from is not available in the browser
```

The client already depends on `ipfs-car` which has jumped through all the hoops to make a CAR file in the browser, so I recommend simplifying the test code to do what we do in the similar API tests:

`createCar` helper function for testing:
https://github.com/web3-storage/web3.storage/blob/4607156dcdc3b2d74c6f9a356664ef11c5e95ba1/packages/api/test/scripts/car.js#L1-L10

example usage:
https://github.com/web3-storage/web3.storage/blob/4607156dcdc3b2d74c6f9a356664ef11c5e95ba1/packages/api/test/car.spec.js#L24-L28

packages/client/src/lib.js
alanshaw: Static functions exist on the class (function) object:

```js
const cid = Web3Storage.putCar(...)
```

Instance methods exist on the prototype:

```js
const client = new Web3Storage()
const cid = client.putCar(...)
```

So there's no collision.

The `class` and `static` syntax is effectively sugar for:

```js
function Web3Storage {}
Web3Storage.putCar = async function () {}
Web3Storage.prototype.putCar = async function () {}
```

terichadbourne:
For the untested lines, I've set them up to be ignored here but opened a new issue #328 as suggested by @olizilla. 

I've [swapped out](https://github.com/web3-storage/web3.storage/pull/293/commits/2a2c689979b4efc7eebbef0e337cec679a4ec943) the code I was using to create a new readable stream CAR with the simpler code @olizilla suggested mimicking, but now I'm having trouble with [car.getRoots()](https://github.com/web3-storage/web3.storage/pull/293/files#diff-1d5466aa3a71f790007cfd7b5459f722d6e2c9e59ff53bdab52f3ac6e403b22aR145) in putCar because something about the formatting is now different. Tests for putting a file pass but tests for putting a CAR file do this: 
image

I thought Mikeal had encouraged me to use that readable stream to match the format that was being passed from the old `put` to the new `putCar` - is there something important that's different in the formatting when we do it this new way? 

I've tried both versions of pulling this in `const car = await createCar('random meaningless bytes')` and passing `car` or `const { root, car: carBody } = await createCar('random meaningless bytes')` and passing `carBody` but both error out. Would love some help!

alanshaw:
@terichadbourne I'm going to jump on this and resolve that last issue to pull it over the line. Hope that's ok! Thanks so much for sorting this for us.

alanshaw (approved):

PR #324, opened 2021-08-09
feat: save report to CSV

alanshaw:
Allows to save all data in the report to CSV for querying.

### DIFF
--- a/packages/report/index.js
+++ b/packages/report/index.js
@@ -73,7 +73,6 @@ export async function run (options) {
           spinner.text = renderText(`Fetching uploads page ${i}...`, { totalUsers, totalUploads, users, userUploads }, options)
           after = uploadsPage.after
           if (!after) break
-          if (i === 5) break
           i++
         }
       })(),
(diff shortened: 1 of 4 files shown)


packages/report/index.js
```diff
@@ -52,47 +56,61 @@ export async function run (options) {
...
+
+          spinner.text = renderText(`Fetching uploads page ${i}...`, { totalUsers, totalUploads, users, userUploads }, options)
+          after = uploadsPage.after
+          if (!after) break
+          i++
         }
+      })(),
+      (async () => {
+        let after = null
+        let i = 1
```
vasco-santos, line 81: Why this and iterate 5 times?

vasco-santos (changes requested):
This looks good! What I missed in one week that we did not have any report tool 😅 
But, I think there is some extra code in the loops?

packages/report/index.js
vasco-santos: Ok, you move faster than me https://github.com/web3-storage/web3.storage/pull/324/commits/b01dece4a72807db682893df08cdd276d19d4c46 😂

vasco-santos (approved):
LGTM 🎉

PR #327, opened 2021-08-10
feat: `w3 put --no-retry`

olizilla:
- Pass `--no-retry` to `w3 put` to have it fail on the first error. Useful for debugging.
- If you want to get really specific you can pass `--retry <number>` to assert how many times you want to try.

```console
$ w3 put pinpie.jpg --no-retry
⁂ maxRetries: 0
...
```

```console
$ w3 put -h

  Description
    Upload a file or directory to web3.storage

  Usage
    $ w3 put <path> [options]

  Options
    --no-wrap     Dont wrap input files with a directory.
    --no-retry    Dont try the upload again if it errors.
    -h, --help    Displays this message
```

License: (Apache-2.0 AND MIT)

### DIFF
--- a/packages/w3/bin.js
+++ b/packages/w3/bin.js
@@ -13,12 +13,13 @@ cli
 cli.command('token')
   .option('--api', 'URL for the Web3 Storage API. Default: https://api.web3.storage')
   .option('--delete', 'Delete your saved token')
-  .describe('Save an API token to use for all requests.')
+  .describe('Save an API token to use for all requests')
   .action(token)
 
 cli.command('put <path>')
   .describe('Upload a file or directory to web3.storage')
-  .option('--no-wrap', 'Dont wrap input files with a directory.')
+  .option('--no-wrap', 'Dont wrap input files with a directory')
+  .option('--no-retry', 'Dont try the upload again if it fails')
   .action(put)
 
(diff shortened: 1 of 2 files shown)


packages/w3/bin.js
```diff
@@ -13,12 +13,13 @@ cli
...
   .option('--delete', 'Delete your saved token')
-  .describe('Save an API token to use for all requests.')
+  .describe('Save an API token to use for all requests')
   .action(token)
 
 cli.command('put <path>')
   .describe('Upload a file or directory to web3.storage')
-  .option('--no-wrap', 'Dont wrap input files with a directory.')
+  .option('--no-wrap', 'Dont wrap input files with a directory')
+  .option('--no-retry', 'Dont try the upload again if it fails')
```
vasco-santos, line 22:
```suggestion
  .option('--no-retry', 'Dont try the upload again if it fails')
  .option('--retry', 'Set maximum number of retries for client put. Default: 5')
```

olizilla: Ah, that was intentional. I decided not to include the "try n times" option explicitly in the help as I think folks should either stick with the default, or turn it off if they are debugging.

vasco-santos (approved):

Issue #273, opened 2021-08-02
Faster pin cron


alanshaw:
@andrew is added content faster than we can process the queue. Lets parallelize this as much as possible.

PR #338, opened 2021-08-11
perf: faster pin status sync

alanshaw:
This reduces the time taken to update pin statuses for 1,000 pins from ~15m to ~5s.

Status requests from cluster are super quick, but DB requests are slower. This switches to using a bulk update UDF, called once per 1,000 pins.

resolves https://github.com/web3-storage/web3.storage/issues/273

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12858,6 +12858,11 @@
         "node": ">=0.10.0"
       }
     },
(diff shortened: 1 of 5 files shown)


vasco-santos (approved):
LGTM! This is great

PR #344, opened 2021-08-12
fix: disallow CAR of single block with links

alanshaw:
If a CAR file is sent to us that has a single block (the root block) and that block has links, then it is an invalid partial DAG.

### DIFF
--- a/packages/api/src/car.js
+++ b/packages/api/src/car.js
@@ -1,6 +1,6 @@
 /* eslint-env serviceworker */
 import { gql } from '@web3-storage/db'
-import { CarReader } from '@ipld/car'
+import { CarReader, CarBlockIterator } from '@ipld/car'
 import { Block } from 'multiformats/block'
 import * as raw from 'multiformats/codecs/raw'
 import * as cbor from '@ipld/dag-cbor'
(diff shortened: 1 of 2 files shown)


packages/api/src/car.js
```diff
@@ -263,41 +261,77 @@ export async function sizeOf (response) {
 }
 
 /**
- * Returns the sum of all block sizes in the received CAR. Throws if any block
+ * @param {CarReader} reader
+ * @param {CarStat} stat
+ */
+async function validateCar (reader, stat) {
```
vasco-santos, line 267: What is the case that leads to the creation of such CAR files? I am not sure if this is the best place for such validation?

olizilla: go-ipfs can create them where it's using a url-store and the url can no longer be reached. it's an edge case that we hit pretty hard, but is otherwise pretty uncommon.

packages/api/src/car.js
```diff
@@ -137,12 +139,7 @@ export async function carPost (request, env, ctx) {
 
   const blob = await request.blob()
   const bytes = new Uint8Array(await blob.arrayBuffer())
-  const reader = await CarReader.fromBytes(bytes)
-
-  const chunkSize = await getBlocksSize(reader)
-  if (chunkSize === 0) {
-    throw new Error('empty CAR')
-  }
+  const stat = await carStat(bytes)
```
olizilla, line 142: CAR could be LARGE at this point. We could pass a clone of the request to stat and have it convert the request body into an AsyncIterable so that we can use CarBlockIterator.fromIterable inside stat, to avoid creating a second copy of the CAR in mem as `bytes`, until after we scan it to figure out if it is large or invalid or both.

olizilla (approved):
It GOOD!

PR #349, opened 2021-08-13
fix: incremental metrics

alanshaw:
Allow specify the time the metric was updated. The "to date" is chosen when the cron starts and when it finishes it saves a metric with an updated date that is much further in the future than the chosen "to date". This PR changes the mutation and UDF to accept an `updated` date so next time the cron runs, it's "from date" is the previous runs "to date" and we don't miss any data in the counts.


olizilla (approved):

vasco-santos:
nice 👌🏼

PR #352, opened 2021-08-16
fix: remove unused user metadata request

alanshaw:
This PR removes the call to `magic.user.getMetadata()`. We do no use the result of this call anywhere at the moment and it has been observed to add 0.5s onto the time it takes the site to determine if the user is logged in.


jdiogopeixoto (approved):
LGTM! 🎉

PR #358, opened 2021-08-17
feat: cron to update status for PinError pins

alanshaw:
Adds a cron job to run once a day and check in with cluster on pins that have the status PinError.

If a pin gets the status `PinError` then we currently do not check again on it. Howveer Cluster does not stop trying to retrieve the CID and in the case of a failed upload the user may successfully re-upload it.

I ran this locally and it updated the status of 4,424 pins that were in the `PinError` state.

### DIFF
--- a/packages/cron/src/jobs/pins.js
+++ b/packages/cron/src/jobs/pins.js
@@ -44,7 +44,7 @@ const UPDATE_CONTENT_DAG_SIZE = gql`
 `
 
 /**
- * @param {import('@web3-storage/api/src/utils/pin.js').PinStatus} statuses Update pins with these statuses.
+ * @param {import('@web3-storage/api/src/utils/pin.js').PinStatus[]} statuses Update pins with these statuses.
  * @param {{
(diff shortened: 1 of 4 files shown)


packages/cron/src/jobs/pins.js
```diff
@@ -44,17 +44,20 @@ const UPDATE_CONTENT_DAG_SIZE = gql`
 `
 
 /**
+ * @param {import('@web3-storage/api/src/utils/pin.js').PinStatus} statuses Update pins with these statuses.
```
vasco-santos, line 47: this should be an array

vasco-santos (changes requested):
Looks good, just an issue in the typedef

packages/cron/src/jobs/pins.js
alanshaw: ah boo

```suggestion
 * @param {import('@web3-storage/api/src/utils/pin.js').PinStatus[]} statuses Update pins with these statuses.
```

vasco-santos (approved):

PR #361, opened 2021-08-17
fix: logo text style

### LINKED ISSUE #221: Lingering Design issues flagged by Martin
@rafaelramalho19  would we be able to fix some of the items listed in teh figma here: https://www.figma.com/file/826Zurcs6kgWSzQ35it966?node-id=2062:565#97496011

There's a design element that's also apparently broken here: https://www.figma.com/file/826Zurcs6kgWSzQ35it966?node-id=1396:565#97499406

rafaelramalho19:
Closes #221

### DIFF
--- a/packages/website/components/navbar.js
+++ b/packages/website/components/navbar.js
@@ -39,7 +39,7 @@ export default function Navbar({ bgColor = '', isLoggedIn, isLoadingUser }) {
             onClick={onLinkClick}
           >
             <img src="/w3storage-logo.svg" style={{ height: '1.8rem' }} />
-            <span className="space-grotesk ml-2 text-w3storage-purple font-medium text-md hidden xl:inline-block">Web3.Storage</span>
+            <span className="space-grotesk ml-2 text-w3storage-purple font-medium typography-logo hidden xl:inline-block">Web3.Storage</span>
           </a>
         </Link>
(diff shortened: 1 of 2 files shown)


packages/website/components/navbar.js
```diff
@@ -39,7 +39,7 @@ export default function Navbar({ bgColor = '', isLoggedIn, isLoadingUser }) {
             onClick={onLinkClick}
           >
             <img src="/w3storage-logo.svg" style={{ height: '1.8rem' }} />
-            <span className="space-grotesk ml-2 text-w3storage-purple font-medium text-md hidden xl:inline-block">Web3.Storage</span>
+            <span className="space-grotesk ml-2 text-w3storage-purple font-medium typography-logo hidden xl:inline-block">Web3.Storage</span>
```
jdiogopeixoto, line 42: You can just use `text-base` instead of adding a new typography class

jdiogopeixoto (changes requested):

packages/website/components/navbar.js
rafaelramalho19: You're right, didn't see the text-base, my bad :)

jdiogopeixoto (approved):

alanshaw (approved):

PR #365, opened 2021-08-18
feat: support any IPLD decoder

alanshaw:
This PR adds an optional parameter to `putCar` called `decoders`. It's an array of IPLD `BlockDecoder`.

It's needed when the DAG in the CAR file was not encoded using the default encoders `dag-pb`, `dag-cbor` and `raw`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1834,6 +1834,16 @@
         "multiformats": "^9.0.0"
       }
     },
+    "node_modules/@ipld/dag-json": {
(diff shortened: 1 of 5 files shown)


vasco-santos (approved):
LGTM

It would be good to have an example, can you create an issue to have it created later on?

alanshaw:
> It would be good to have an example, can you create an issue to have it created later on?

Good idea!

https://github.com/web3-storage/web3.storage/issues/367

PR #373, opened 2021-08-19
feat: api supports upload list sorting

vasco-santos:
This PR adds support for sortBy and sortOrder in the API and DB Function. 

The support for PinStatus and Size needs to be added as a follow up, as there are a few open questions that we should move to an issue:  https://github.com/web3-storage/web3.storage/pull/379#discussion_r693995972

Meanwhile, we can unblock UI work with these

Query Example in fauna Dashbord (needs to have userId updated to a valid one for you):

```
query findUploadsByUser {
  findUploadsByUser (where:{ user: "303388817558077963", createdBefore: "2021-09-09T19:07:16.161Z"}, sortBy: Name, sortOrder: Asc, _size: 10) {
    data {
      _id
      name
      content {
        cid
        dagSize
      }
      created
    }
  }
}
```

### DIFF
--- a/packages/db/fauna/resources/Function/findUploadsByUser.js
+++ b/packages/db/fauna/resources/Function/findUploadsByUser.js
@@ -42,30 +42,11 @@ const body = Query(
                 Index('upload_by_user_sort_by_name_asc'),
                 Index('upload_by_user_sort_by_name_desc')
               ),
-              // Size Sort
+              // Default to Date Sort
               If(
-                Equals(Var('sortBy'), 'Size'),
-                // TODO: DO to obtain all uploads
-                If(
-                  Equals(Var('sortOrder'), 'Asc'),
-                  Index('upload_by_user_sort_by_size_asc'),
-                  Index('upload_by_user_sort_by_size_desc')
-                ),
-                // PinStatus Sort
-                If(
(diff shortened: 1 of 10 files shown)


packages/db/fauna/schema.graphql
```diff
@@ -367,6 +367,18 @@ enum DealStatus {
   Terminated
 }
 
+enum Sort {
```
alanshaw, line 370: ?

```suggestion
enum SortDirection {
```

rafaelramalho19 (approved):

packages/db/fauna/resources/Function/findUploadsByUser.js
```diff
@@ -26,12 +26,30 @@ const {
...
+                Index('upload_by_user_sort_by_name_desc')
+              ),
+              // Default to Date Sort
+              If(
+                Equals(Var('sortOrder'), 'Asc'),
+                Index('upload_by_user_sort_by_created_asc'),
+                Index('upload_by_user_sort_by_created_desc')
+              )
+            ),
+            // Index('upload_by_user_sort_by_created_desc'),
```
vasco-santos, line 52:
```suggestion
```

PR #374, opened 2021-08-19
feat: add repl.it code sample to the tokens page

### LINKED ISSUE #317: Add web sandbox to account page
To reduce the time between creating a token and uploading your first file:

Add a repl.it instance to the account page,  so that people can execute an example script that adds a file with their own API key right after creating it.

rafaelramalho19:
Closes #317


packages/website/pages/tokens.js
```diff
@@ -203,6 +203,14 @@ export default function Tokens({ isLoggedIn }) {
                 </div>
               </div>
             </When>
+            <When condition={tokens.length === 1}>
+              <div className="flex flex-col items-center">
+                <p className="font-black mt-10">
+                  Want to test the token quickly? Upload a file in 
+                  {" "}<a className="underline" href={`https://web3storage-tokens-srv.rafaelramalho1.repl.co/?token=${tokens[0].secret}`} target="_blank" rel="noreferrer">repl.it</a>
```
olizilla, line 210: We need to set up a web3.stroage account for repl.co rather than relying on `rafaelramalho1`

rafaelramalho19: Done :)

packages/website/pages/tokens.js
```diff
@@ -203,6 +203,14 @@ export default function Tokens({ isLoggedIn }) {
                 </div>
               </div>
             </When>
+            <When condition={tokens.length === 1}>
+              <div className="flex flex-col items-center">
+                <p className="font-black mt-10">
+                  Want to test the token quickly? Upload a file in 
+                  {" "}<a className="underline" href={`https://Upload-with-Token-Node-demo.web3storage.repl.co/?token=${tokens[0].secret}`} target="_blank" rel="noreferrer">repl.it</a>
```
olizilla, line 210: nit: the url gets redirected to the lowercase equivalent, so let's just lowercase it here. 

Also let's not pass in the first token to the repl. The user should copy the token they want and paste it into the repl window manually. If i have 3 tokens listed, i'd like to know which token i am testing.

Also also, passing the token to repl.co in as a search param will have it appear in repl.co's http request logs which is bad, and the users brower history, which is less bad, but still undesirable.

olizilla (changes requested):

olizilla:
Screenshot 2021-09-06 at 15 46 06

Also, please update the page title in the repl to read `Web3.Storage token tester`.

packages/website/pages/tokens.js
rafaelramalho19: Alright, I'll also change it so that the demo can show with any number of tokens, right now it's only if you have 1 token only.

olizilla:
The repl.it sample always takes a while to warm up for me, which feels kinda rough. What about if we direct folks to our client upload example at https://bafybeic5r5yxjh5xpmeczfp34ysrjcoa66pllnjgffahopzrl5yhex7d7i.ipfs.dweb.link/ which also lets them paste in a token, and upload a file of their choosing, and lists the uploads on the account.

atopal:
Yes, let's land this as Oli suggested, and we then we might want to iterate.

rafaelramalho19:
Done, please review @olizilla 🙏

PR #379, opened 2021-08-20
Feat/sort uploads list

### LINKED ISSUE #276: Add ability to sort/filter files in the web UI
Users should be able to sort and filter files in the Web UI. Currently the table is static with fixed pagination.

rafaelramalho19:
- Closes #276


packages/db/fauna/resources/Function/findUploadsByUser.js
```diff
@@ -26,12 +26,49 @@ const {
...
+                  Index('upload_by_user_sort_by_size_desc')
+                ),
+                // PinStatus Sort
+                If(
+                  Equals(Var('sortBy'), 'PinStatus'),
+                  If(
+                    Equals(Var('sortOrder'), 'Asc'),
+                    Index('upload_by_user_sort_by_pin_status_asc'),
+                    Index('upload_by_user_sort_by_pin_status_desc')
+                  ),
```
vasco-santos, line 61: Let's remove this until we figure out how to properly do this?

Particularly in PinStatus:
- Would it be a sort by Pin status alphabetical order?
- What is the Pin Status when an upload has multiple Pins and they have different states?

For size itself, We need to list all uploads a user have sorted by size. So, we have a userId and need to get all the uploads for the user (which needs to be paginated…), and with that we need to get the Content Ref of them, get the content doc and sort by the content size, in order to than return the actual Uploads.
Sadly, we can’t do an index with two collections where we could get uploads from user and values for dagSize to sort.
So, I am not seeing any other solution except to replicate the dagSize of Content also in upload?

vasco-santos (commented):
Any ideas why CI is now failing?

packages/db/fauna/resources/Function/findUploadsByUser.js
rafaelramalho19: That should be tracked in #373 I believe

vasco-santos: ok, I updated the branch to unblock this

vasco-santos: I merged it in, you can get this updated to get this shipped 🚢

PR #385, opened 2021-08-23
fix: pin sync request history

alanshaw:
Stops history accumulating on `PinSyncRequest` collection.

`deleteWithHistory` provided by Fauna support.

Currently deployed to prod.

### DIFF
--- a/packages/db/fauna/resources/Function/deletePinSyncRequests.js
+++ b/packages/db/fauna/resources/Function/deletePinSyncRequests.js
@@ -12,7 +12,7 @@ const {
   Map,
   Collection,
   Ref,
-  Delete
+  Call
 } = fauna
 
 const name = 'deletePinSyncRequests'
@@ -21,7 +21,7 @@ const body = Query(
     ['requests'],
     Map(
       Var('requests'),
-      Lambda(['id'], Delete(Ref(Collection('PinSyncRequest'), Var('id'))))
+      Lambda('id', Call('deleteWithHistory', Ref(Collection('PinSyncRequest'), Var('id'))))
     )
   )
 )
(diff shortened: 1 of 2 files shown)


olizilla:
From Fauna, so we don't forget:

> We have good evidence now that this is an issue with history.  I thought it was unlikely that there was an issue with history from Updates before.  But there are many create/delete events that are being read along with the current data you are requesting

> We've cleaned up about 1 million events from the PinSyncRequest Collection.  If there are just the 1 create + 1 delete events per document, that means about 500k documents gone from the history.  We expect you should be seeing performance improvements by now.

> We have some UDFs to use as replacements for the Update and Delete functions.  They internally use the Remove function to permanently delete the history as they go. It is important to note that each call will cost at least 2 Write Ops, rather than one.  It can also cost more compute to iterate through more history.  It is only impactful if that single document has a lot of history.  For example, to use the offered `DeleteWithHistory` function on a document that had 64 history events, it cost 11 Write Ops and 11 Compute Ops.

>  history_days=0 tells the garbage collector it's okay to delete any events more than 0 days old.  Delete only adds a "delete" event and all of that document's history will remain until the GC get's to it.

> The DeleteWithHistory function should be deployed ideally before the cron job starts running again.  If the cron job runs successfully through completion there could be 100's of thousand of more garbage docs created.  We will be unable to keep up sustainably, manually, running a targeted GC. verify Paginate(Documents(Collection("PinSyncRequest")), { size: 64 }) does not abort.  We found that requesting a larger page actually more reliable comes back with data.  size = 64 is the default, so the explicit option here is redundant. assuming that paging data directly from the Documents index is working, apply a max page size of 64 to findPinSyncRequests.  Running some performance checks, we found today that scanning the whole PinSyncRequest Collection by pages of 64 takes as much time as scanning through with pages of 1000.

olizilla (approved):
Already in prod.

PR #389, opened 2021-08-24
fix: delete fauna history when deleting PinRequest

olizilla:
Suggested by the fauna team, a similar fix to https://github.com/web3-storage/web3.storage/pull/385. The code was calling the auto-generated `deletePinRequest` function, which does not delete the document snapshots, and leads to a build of of data that slows down queries until they fail to complete within the time budget.

License: (Apache-2.0 AND MIT)


vasco-santos (approved):
LGTM

Issue #386, opened 2021-08-23
Allow deleting from the CLI


andrew:
I'd like to be able to delete things using the [`w3`](https://github.com/web3-storage/web3.storage/tree/main/packages/w3) cli, something like:

```shell
w3 delete bafybeibibb5g656femgrpkkixmke2lzp33i5sf5oaivyc2rv6olc3nst7u
```

olizilla:
Blocked on #314

Issue #314, opened 2021-08-05
Token scoping


dchoi27:
Issues related to how tokens are scoped (permissions, how users interact with multiple tokens, etc.).

dchoi27:
@alanshaw know you had some thoughts related to token permissions, feel free to make a new issue and add to this epic!

alanshaw:
## Problem

At the moment, developers building web applications are likely to bundle a Web3.Storage token together with the app. This allows users to interact directly with the Web3.Storage service from their browser. However it has some issues:

* Token is openly available in the application source code. Anyone can take the token and use it to upload their own content.
* The token when used to list uploads does so on an account level so you see uploads that were created using other tokens.

The alternatives to bundling a token currently are:

1. Applications ask users to sign up at web3.storage and obtain a token for use. This is terrible UX.
2. Application developers create their own API and proxy uploads from their users. This is a lot of work and means that content is funneled through a centralized point that is maybe not as well prepared for the amount of traffic it may receive.

## Proposal: Token Scoping and Token Management API

Developers simply have to implement an API in their app that creates "user" scoped tokens. It uses an "admin" scoped token to do this. The "user" token may be stored in `local storage` for use by the web app. It is a centralized component of the user's app, but avoids token bundling, asking the user for a token and allows uploads listings that are filtered to the token that performed the upload.

### Scopes

Public permissions are implied.

1. **Admin**
    * Create _user_ token
    * Delete _user_ token
    * List _user_ tokens
    * Upload
    * List _all_ uploads
    * Delete _any_ upload
2. **User**
    * Upload
    * List _owned_ uploads
    * Delete _owned_ upload
3. **Legacy** (existing tokens already created)
   * Upload
   * List _all_ uploads
   * Delete _owned_ upload

### Work Involved

* The list uploads endpoint should fetch uploads belonging to a token if it has the "user" scope, else everything for the account
* The delete endpoint should allow deletion of any content from a token with "admin" scope
* The create token endpoint should...
    * Accept a `scope` param that allows magic.link tokens to create "admin" or "user" scoped tokens
    * Allow a token with "admin" scope to create a "user" scoped token
* The delete token endpoint should allow a token with "admin" scope to delete (revoke) a "user" scoped token
* The list tokens endpoint should allow a token with "admin" scope to list all "user" scoped tokens
* UI work for selecting scope of token to create from the website and showing scopes in token list

### Alternative

[...]

PR #390, opened 2021-08-24
feat: add delete upload to w3 and API

rafaelramalho19:
Wanted to learn a bit more about the w3 & API integration, so I made it possible to delete uploads through the API & W3 CLI

- Closes https://github.com/web3-storage/web3.storage/issues/386


packages/w3/index.js
```diff
@@ -195,6 +195,24 @@ export async function put (path, opts) {
   console.log(`⁂ https://dweb.link/ipfs/${root}`)
 }
 
+/**
+ * Deletes 1 or more uploads
+ *
+ * @param {string} cid the root CID to fetch
```
vasco-santos, line 201:
```suggestion
 * @param {string} cid the root CID to delete
```

packages/w3/index.js
```diff
@@ -195,6 +195,24 @@ export async function put (path, opts) {
...
+ * @param {object} opts
```
vasco-santos, line 202:
```suggestion
 * @param {object} [opts]
```

let's make opts optional

packages/w3/index.js
```diff
@@ -195,6 +195,24 @@ export async function put (path, opts) {
...
+ * @param {string} [opts.api]
+ * @param {string} [opts.token]
+ * @param {string} [opts.output] the path to write the files to
+ */
+export async function deleteUpload (cid, opts) {
```
vasco-santos, line 207:
```suggestion
export async function deleteUpload (cid, opts = {}) {
```

packages/client/src/lib.js
```diff
@@ -208,12 +208,33 @@ class Web3Storage {
...
+      const response = await res.json()
+      if (response.response.errors) {
+        console.error(response.response.errors)
+        return undefined
+      }
+
+      return response
+    } catch (e) {
+      console.error(e)
+      return undefined
```
vasco-santos, line 236: humm, wondering if we should fail in case of no delete found. what do you think?

vasco-santos (changes requested):

packages/client/src/lib.js
rafaelramalho19: It gracefully fails, as it says:
image

vasco-santos: yes, if fails gracefully in the CLI, but I am concerned on the client UX. What if we go with an approach like JS built in Map [return value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/delete#return_value)? We would return `true` if deleted, `false` otherwise. This would fit well with our `client.get` behaviour.

rafaelramalho19: Gotcha, changing it

packages/w3/index.js
rafaelramalho19: Any reason for that? If so, we should change it on all the commands, no?

packages/api/src/index.js
```diff
@@ -31,8 +31,8 @@ router.post('/car',                 auth['🔒'](carPost))
 router.put('/car/:cid',             auth['🔒'](carPut))
 router.post('/upload',              auth['🔒'](uploadPost))
 router.get('/user/uploads',         auth['🔒'](userUploadsGet))
+router.delete('/user/uploads/:cid', auth['🔒'](userUploadsDelete))
```
olizilla, line 34: This is a blocker here. Until we get scoped api tokens as per #314 we deliberately made it so tokens can only be used for non-destructive actions like uploading and listing.

olizilla (changes requested):
Nice PR! Alas, it is blocked on #314 as we do not want api tokens to be able to perform destructive actions until users are able to specify the powers a token should have when creating them.

You can't delete with an api token until we land that feature.

packages/w3/index.js
vasco-santos: well, given no one will use this as a lib, I will not expect problems. But typescript compiler would fail if this function is used as `x.deleteUpload(cid)` because `opts` is not documented as optional. However, none of the Object properties are required. Not a big deal here

PR #391, opened 2021-08-24
feat: add a new step to the getting started

### LINKED ISSUE #277: Add a "upload a file" getting started option
Add a "upload a file" getting started option to the onboarding screen of web3.storage
image

 that immediately takes you to the files page (before creating any API token) so you can add your first file

image

jdiogopeixoto:
- Adds a new "upload your first file" step to the getting started section inside the `/account` page. It only shows up when the user hasn't uploaded any file.

image

- Adds a new information block to the `/upload` page.

image

- Adds (myself as) a new entry to the `wrangler.toml` file.

Closes #277


olizilla (changes requested):
The caution box here is a little too scary. Can we make it a bit less intense

- Move the info below the upload box.
- Remove the red box.
- Remove the Caution header
- Split the content into 2 sections with headers (see https://car.ipfs.io/ for a layout example)
  - *🌍 Public data*
  - *♾ Permenant data*
- Give the text a narrower max-width to make it easier to read

> “Anything from 45 to 75 characters is widely regarded as a satisfactory length of line for a single-column page… the 66-character line (counting both letters and spaces) is widely regarded as ideal. For multiple-column work, a better average is 40-50 characters.” 
— Robert Bringhurst, The Elements of Typographic Style [via tachyons.io](http://tachyons.io/docs/typography/measure/)

olizilla:
The empty / new account view should clearly encourage a single next step... upload a file! That action should be the clearest thing on this page. 

Then as secondary items with less visual weight we should offer "Create a token" or "Read the docs" as other options. "Start building" does not tell the user that we are offering them docs.

cc @atopal for verification.

jdiogopeixoto:
@olizilla @atopal, I just updated the PR description with a new screenshot, please let me know your thoughts.

atopal:
Thanks @jdiogopeixoto that works for me!

olizilla (changes requested):
there is a change to the root package-lock.json. that seems unintended, can it be removed from this PR.

there is a package-lock.json in the website procect that i didn't know we had. Ideally we all run our npm commands from the root of the workspace and just use the package-lock.json at the root to pin dependences...so installing deps and running the website would be done from the root of the projcet by

```
npm install
npm start -w packages/website
```

this would avoid creating a sparete node_modules folder and package-lock in the website folder.

olizilla (approved):

Issue #395, opened 2021-08-24
backups: write CAR uploads to S3


mikeal:
We have a lot of guarantees in our backend about replication across the IPFS cluster, and then asynchronously we have backups into Pinata, but there’s still the potential for us to lose data if there is a cluster wide issue in data that hasn’t made it to a Pinata backup.

On upload, we could write the CAR file to S3, perhaps keyed by a hash of the entire CAR file, and then we would just need to record that hash in the database for that upload.

We can go back and remove these once data is archived in enough Filecoin deals that we’re confident we don’t need them, but we can worry about that later, the first version of this can just leave them around until we decide to go through them.

mikeal:
we need to make sure any schema changes for this make it into the Postgres refactoring.

dchoi27:
Thanks for making this issue! Let's have a broader discussion with the team about whether this is the right backup solution.

PR #417, opened 2021-08-29
feat: storage backup

vasco-santos:
This PR adds storage backup per https://github.com/web3-storage/web3.storage/issues/395

In short, it sends each received file to S3 to work around possible failures, where we will keep it until we do not need a backup. This includes the following changes in the codebase:
- `POST /car`, `POST /upload` add the received file(s) to S3, keyed by its content hash (parallel with adding to cluster)
- added database collection to keep track of existing backups per upload
  - with support for adding deleted timestamp
  - though about a relation with content instead, but if we want to eventually return success if we can add the files to backup but not in cluster, we should not have the content created in the DB
- `createUpload` UDF updated to add received backupKeys to the upload
  - note that we support uploading directories, so this must receive an array

Assumptions:
- The implementation currently assumes we do not require backup configured for development. I find this to be the best alternative, given creating a personal account in AWS takes days to get it verified.
- Given we only use this for PUT (catastrophic events), and putting in a zone is also faster than the parallel ops (cluster.add + cluster.status) we don't currently need to put this in different zones to enhance faster GETs.
  - Either way, we will be storing the name of the bucket so that we can move this solution into a map of [cloudflare country region](https://developers.cloudflare.com/workers/runtime-apis/request#incomingrequestcfproperties) and S3 regions.
  - We will need to implement a custom solution for this, given there is not such map on npm as far as I can tell.

Possible improvements we can do in follow up PRs:
- Setup S3 bucket [cross replication](https://aws.amazon.com/s3/features/replication/?nc1=h_ls) for other zones
- Do not fail if we can add to backup, but not cluster
  - We need to have a job to try to make errored data in first
- ...

Production and staging credentials were added to wrangler, and infrastructure was setup

Closes https://github.com/web3-storage/web3.storage/issues/395


packages/api/package.json
```diff
@@ -48,12 +50,13 @@
...
     "multiformats": "^9.0.4",
-    "p-retry": "^4.6.1"
+    "p-retry": "^4.6.1",
+    "uint8arrays": "^3.0.0"
   },
   "bundlesize": [
     {
       "path": "./dist/main.js",
-      "maxSize": "1 MB",
+      "maxSize": "1.4 MB",
```
vasco-santos, line 59: This makes a huge difference given https://bundlephobia.com/package/@aws-sdk/client-s3@3.28.0 

As this is the worker, it is not problematic, even though it is a lot

packages/api/src/env.js
```diff
@@ -49,4 +58,22 @@ export function envAll (_, env, event) {
   const clusterAuthToken = env.CLUSTER_BASIC_AUTH_TOKEN || (typeof CLUSTER_BASIC_AUTH_TOKEN === 'undefined' ? undefined : CLUSTER_BASIC_AUTH_TOKEN)
   const headers = clusterAuthToken ? { Authorization: `Basic ${clusterAuthToken}` } : {}
   env.cluster = new Cluster(env.CLUSTER_API_URL || CLUSTER_API_URL, { headers })
+
+  try {
+    if ((env.S3_ACCESS_KEY_ID || S3_ACCESS_KEY_ID) && (env.S3_SECRET_ACCESS_KEY_ID || S3_SECRET_ACCESS_KEY_ID)) {
+      const s3Endpoint = env.S3_BUCKET_ENDPOINT || (typeof S3_BUCKET_ENDPOINT === 'undefined' ? undefined : S3_BUCKET_ENDPOINT)
+      env.s3Client = new S3Client({
+        endpoint: s3Endpoint,
+        forcePathStyle: !!s3Endpoint, // Force path if endpoint provided
```
vasco-santos, line 67: only way we can enforce a path for testing. In alternative, given backup is not required, we can remove this and not test with it, as we should be able to trust `@aws-sdk/client-s3` module

packages/api/package.json
olizilla: bundlephobia says it tree-shakes, so may not be so bad. is our build process taking advantage of that?

packages/api/README.md
```diff
@@ -181,3 +185,7 @@ SENTRY_UPLOAD=false # toggle for sentry source/sourcemaps upload (capture will s
 ```
 
 Production vars should be set in Github Actions secrets.
+
+## S3 Setup
+
+We use [S3](https://aws.amazon.com/s3/) for backup storage. For production an accounton AWS needs to be created.
```
olizilla, line 191:
```suggestion
We use [S3](https://aws.amazon.com/s3/) for backup storage. For production an account on AWS needs to be created.
```

packages/db/fauna/schema.graphql
```diff
@@ -215,6 +215,36 @@ type PinSyncRequest {
...
+
+  """
+  Backup key.
+  """
+  key: String!
+
+  """
+  Backup bucket name.
+  """
+  name: String!
```
olizilla, line 235: `name` `is ambigious here, it should be `bucketName`.

olizilla (changes requested):
This is decent, and we could deploy as is, but i think we should:

- Rename things to be more specific. ~Backup~ -> S3Object (or similar). We also have code to back our pins up to Pinata, so we should be clearer in the code and schema that these new code paths are about staging user uploads on S3 while we wait for the filecoin deals to become active.

- Have seperate S3 object key prefixes e.g `car/` and `files/` (or buckets) for CAR uploads vs raw file uploads. They will have to be handled seperately to recreate a cluster from a bucket, so it would he helpful to be able to query S3 for a list of each type. I am imagining a recovery job that woulc have to either slurp in car files with format=car or format=unixfs to a new ipfs-cluster. That could be determined from our db, but it'd operationally comforting if we could also see the difference in s3.

-  _nice to have_ preserve the filenames and paths for raw file uploads. we could name them `files/<hash>/<file path>` or similar... or we could wait for the CID to come back from cluster and key raw file uploads with a CID prefix, so a recovery script could determine which raw files were uploaded together. Again, this is all available in the db, but it would be rad if we didn't have to trust and rely on the db in a disaster scenario.

packages/db/fauna/schema.graphql
vasco-santos: I named backup and name here taking into consideration that we are using S3, but it could be any other solution. I would prefer to use agnostic names that are not tied to the service we use. What do you think?

vasco-santos:
> nice to have preserve the filenames and paths for raw file uploads. we could name them files/<hash>/<file path> or similar... or we could wait for the CID to come back from cluster and key raw file uploads with a CID prefix,

👍🏼  this is a good call

I will add this, waiting on CID from cluster will mean longer response times in uploads (also more likely timeouts from worker). I think you meant upload name here? For `/files` we could use file name/path, but for CAR this would require unixfs exporting. My proposal is to use backup keys as follows:

- `/files/<upload_name>/car/<hash>`
- `/files/<upload_name>/file/<hash>`
- `/directory/<upload_name>/file/<hash>`

Let me know your thoughts

olizilla:
my point about preserving file names was motivated by the difference where a CAR of a file tree would preserve all the file names internally, so it would be ok to simply key those by the hash of the CAR, but we would lose all the filenames of and uploaded directory if they were only keyed by hash. I had not considered the user "upload name" property. I'm kinda ok with that remaining in the db for now.

it would be worth spiking out what a recovery script would look like. At it's simplest we's just want to be able to quickly stand up a new cluster and re-add all the content from S3, producting the same CIDs. Simple enough for the CAR files, but needs more care for the raw files.

packages/db/fauna/schema.graphql
```diff
@@ -514,11 +544,17 @@ input CreateUploadInput {
   cid: String!
   type: UploadType!
   name: String
+  backupData: [createBackupInput]!
```
alanshaw, line 547: Minor: `backupData` is not actually the data that is being backed up, just information about where it can be found.

Maybe `backupLocations` or `backupUrls` if you switch to using a URL instead.

packages/db/fauna/resources/Function/createUpload.js
```diff
@@ -103,6 +103,20 @@ const body = Query(
                         )
                       )
                     ),
+                    Foreach(
+                      Select('backupData', Var('data')),
+                      Lambda(
+                        ['data'],
+                        Create('Backup', {
```
alanshaw, line 110: If a user uploads the same file twice then we'll get multiple backup objects for the same CAR pointing to the same bucket+key?

alanshaw (commented):
I don't want to stand in the way of this, but it would be awesome if these went into s3 by root CID, so we could just serve them from a gateway easily. I feel that would be quite difficult to do as this PR stands but I appreciate that this is not the motivation behind the work and just a nice to have.

Maybe partials could go in as directories:

```
root_cid/sha256(car0)
root_cid/sha256(car1)
root_cid/sha256(car2)
```

...and then we stream them out by listing the directory and trimming the CAR header from all but the first.

Hard bits:

* detect partials (maybe everything gets a directory even if not partial)
* undeterministic CARs
* it would be nice not to do a directory listing for non-partials (I'd like to just be able to just get `root_cid.car` from the bucket)

The cool bit here is that if we can add by root CID then we don't _need_ to store the bucket key of each partial CAR along with the upload...

vasco-santos:
> the (probably safe) assumption is that it will be quicker to rename a key in s3 than it is to upload the file.

@olizilla Sadly, s3 JS sdk does not seem to support rename/move in an atomic operations. We would need to use copyObject with deleteObject to achieve this without having data replicated.

We can make pin and backup in parallel, but given this is a disaster recovery, I think we are better with just getting the information from the DB to perform the recovery.

@alanshaw I like your suggestion, but as you mention this might leave the scope of the backup a bit. If we want this to be more than a backup, and actually be able to use S3 as a backend of a gateway, we likely need to go with computing rootCID first. When we receive a CAR, we can likely trust the received car and use its root to backup (if we receive a different root from IPFS cluster later on, we can revert the backup). But, with generic files, we will not have the rootCID beforehand.

packages/api/src/car.js
```diff
@@ -169,6 +161,12 @@ export async function carPost (request, env, ctx) {
         cid,
         name,
         type: 'Car',
+        backupData: backupKey
+          ? [{
+              key: backupKey,
+              name: env.s3BucketName
+            }]
```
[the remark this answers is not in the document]
vasco-santos: I think this is a good idea! I will re-check it once I get back to the PR

olizilla:
@alanshaw It's unfortunate, but I think we have to avoid treating what goes in to S3 as something we could serve back to the world as content-addressed data, without some post-processing.

The problem is these are user uploads from the wild. They could send us a deliberately malformed CAR that contained blocks that are not part of the DAG for the declared root CID. Such a CAR could could be added to cluster, and we'd get back the root CID as declared, but it would not be correct to store and use that file as **the** CAR file for that CID.

vasco-santos:
Given #480 is on the ways to unify both `/upload` and `/car`, we will wait for it to be merged here.
This allows us to get to know the root cid in advance and act accordingly to save data into S3 with the following namespace:

`/${rootCid}/${userId}/${carHash}`

As a next step, we will also look into a script to boot a new IPFS cluster with all the data stored in S3, as well as S3 data normalization

packages/api/package.json
vasco-santos: oh, I was not aware of it. Per the docs:

> A Workers script can be up to 1MB in size after compression.

our bundlesize check is done pre-compression though. I will look into that too

vasco-santos: ok, making the bundle size job test compressed made the bundle size change from `1.06MB` to `287.12KB`

packages/db/fauna/resources/Function/createUpload.js
vasco-santos: We will get a single S3 object stored considering it is stored on a key value fashion, but multiple entries in Fauna for the same file.

I thought about adding an index at first, but it would run in every single upload to check if it already exists. Using `@unique` would also fail, which we don't want.

So, given we should just iterate on the list of objects in S3 for backup I think the best solution is to simply keep record of everything as is. With the record, we can easily access specific data to prfioritize backups as needed.

What do you think?

packages/api/package.json
```diff
@@ -46,15 +47,16 @@
...
     "multiformats": "^9.0.4",
-    "p-retry": "^4.6.1"
+    "p-retry": "^4.6.1",
+    "uint8arrays": "^3.0.0"
   },
   "bundlesize": [
     {
       "path": "./dist/main.js",
-      "maxSize": "1 MB",
-      "compression": "none"
```
vasco-santos, line 57: https://developers.cloudflare.com/workers/platform/limits#script-size

> A Workers script can be up to 1MB in size after compression.

So let's just test this as the limits are defined by Cloudflare

packages/api/README.md
```diff
@@ -46,6 +46,10 @@ One time set up of your cloudflare worker subdomain for dev:
     wrangler secret put FAUNA_KEY --env $(whoami) # Get from fauna.com after creating a dev Classic DB
     wrangler secret put CLUSTER_BASIC_AUTH_TOKEN --env $(whoami) # Get from web3.storage vault in 1password (not required for dev)
     wrangler secret put SENTRY_DSN --env $(whoami) # Get from Sentry (not required for dev)
+    wrangler secret put S3_BUCKET_REGION --env $(whoami) # Get from Amazon S3 (not required for dev)
```
olizilla, line 49: what's the transition plan from setting a bucket region to picking dynamically? 

Should this be a list of available bucket regions available for production?
Should we just tell devs to create one in a specific region, and in dev mode always default to that region?

olizilla, line 49:
```suggestion
    wrangler secret put S3_BUCKET_REGION --env $(whoami) # e.g us-east-2 (not required for dev)
```

packages/api/README.md
```diff
@@ -46,6 +46,10 @@ One time set up of your cloudflare worker subdomain for dev:
...
+    wrangler secret put S3_ACCESS_KEY_ID --env $(whoami) # Get from Amazon S3 (not required for dev)
+    wrangler secret put S3_SECRET_ACCESS_KEY_ID --env $(whoami) # Get from Amazon S3 (not required for dev)
+    wrangler secret put S3_BUCKET_NAME --env $(whoami) # Get from Amazon S3 (not required for dev)
```
olizilla, line 52:
```suggestion
    wrangler secret put S3_BUCKET_NAME --env $(whoami) # e.g web3.storage-staging-us-east-2 (not required for dev)
```

packages/api/README.md
vasco-santos: Good question. For having multiple regions (from web3.storage put data perspective) we would need multiple S3_BUCKET_REGION and multiple S3_BUCKET_NAME.

What if we move this into wrangler config env vars instead (keeping just the access key as secrets), making them an array? We would only support one for now, but it would be easier to just add more later on.

Either way, with our current approach of parallel cluster.add, we will not have gains from multiple region put. So, the best approach at this point for a next step would be to set up [Amazon S3 Cross-Region Replication (CRR)](https://aws.amazon.com/s3/features/replication). This means we would put it in one region (sometimes not the ideal region for users, but faster than cluster to resolve) and than content would be replicated in other regions to allow enhanced GETs.

olizilla: > What if we move this into wrangler config env vars instead (keeping just the access key as secrets), making them an array

agree, something like:
```
S3_BUCKETS = [{ s3Region: "xxx", bucketName: }]
```

> this means we would put it in one region.

I think the future enhancement is more like we'd want to put to the single bucket nearest to the cf worker that is receiving the upload, rather than put to many at once.

It'd be neat to have it so that we can have the option in the future of using a set of buckets, 1 per [dvd region](https://meta.wikimedia.org/wiki/List_of_countries_by_regional_classification#/media/File:DVD-Regions_with_key-2.svg) or [whatever](https://meta.wikimedia.org/wiki/List_of_countries_by_regional_classification), and then map cf regions to aws regions, so the put doesn't have to go so far.

I think cross-region replication might be interesting for a cdn, but less so for backing up our backup.

All this said tho, let's leave this PR as is, and then see what we actually need when we get there.

packages/website/pages/files.js
```diff
@@ -142,6 +142,7 @@ const UploadItem = ({ upload, index, toggle, selectedFiles, showCopiedMessage })
       setError('')
     } catch(e) {
       console.error(e);
+      // @ts-ignore Catch clause variable type annotation must be 'any' or 'unknown' if specified.ts(1196)
```
vasco-santos, line 145: This is a side effect of updating the ts compiler dep, which was making cloudflare pages CI job to fail. We seem to assume it will always be an Error we know about, and probably this is a safe assumption, but I am flagging it.

packages/api/README.md
vasco-santos: Yeah, that sounds good. It is an easy change later on

packages/api/package.json
olizilla: i'm getting an error on launching the api in dev mode now about 50% of the time, so we may need to find a lighter s3 client or figure out why cf is less happy

vasco-santos: it is not related to s3, but to wrangler itself. I will share with you the details

packages/db/fauna/resources/Function/createUpload.js
vasco-santos: Regarding this, If a same upload is created, we will not create the Upload entry in Fauna, which means we will not create the backups for this in Fauna as well.

olizilla (approved):
🚀

packages/db/fauna/resources/Function/createUpload.js
vasco-santos: The above is only true because there is a bug 🤦🏼  We are not adding the following chunks to Fauna... Working on a PR

PR #418, opened 2021-08-30
chore: extract pinpin package for pinning our pins on pinata

olizilla:
- extract and simplify pinata backup process to pin to pinata as fast as possible
- add Procfile to run the process continuously as a worker on heroku.

The previous code blocked on each request to pinata, so would not make full use of the 3 req/s rate limit.

Github cron schedules are best effort, and we were seeing > 30 mins between runs, which would let the PinRequest queue fill up. This PR gets the backup process off of gh cron, and off of my machine, and on to a dyno, where it can run more reliably and continually. A web3-storage team has been set up on heroku to manage access. The dyno runs as a single worker thanks to the the `dyno scale web=0 pinpin=1` command, and uses node 16 & npm 7 to run due to setting them in the root `engines` field. This branch was deployed to heroky via `git push heroku pinpin:main`. 

Screenshot 2021-08-30 at 14 39 09

License: (Apache-2.0 AND MIT)


packages/pinpin/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+```sh
+npm run start
+```
+
+## Running on Heroku
+
+A Procfile is added to the root of the monorepo to define the `pinpin` process.
+- we have a web3-storage team, so run it there.
+- set
```
yusefnapora, line 35: kind of trails off here :)

packages/pinpin/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+It's a task queue! Fetch the oldest 600 PinRequests from the DB and pin them on Piñata, then delete the PinRequests from the db when successfully submitted.
+- At the time of writing the Fauna query to fetch PinRequests hits errors if a batch size other than 600 is used.
+- We can only upload to Piñata at 3req/s max, as they have rate-limiting in place.
+
+## Getting started
+
+Ensure you have all the dependencies, by running `npm i` in the parent project.
+
+To run this locally you will need the following in your `packages/cron/.env` file:
```
vasco-santos, line 14: Can we use `.env.local`?

packages/pinpin/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "version": "1.0.0",
+  "description": "pin pins on pinata",
+  "license": "(Apache-2.0 AND MIT)",
+  "type": "module",
+  "main": "index.js",
+  "bin": {
+    "pinpin": "./src/bin.js"
+  },
+  "scripts": {
+    "start": "DEBUG_HIDE_DATE=true DEBUG=pinpin ./src/bin.js"
```
vasco-santos, line 12: Do we want debug mode in prod?

packages/pinpin/src/bin.js
```diff
@@ -0,0 +1,52 @@
...
+    throw new Error('missing FAUNA_KEY environment var')
+  }
+  return new DBClient({ token: env.FAUNA_KEY })
+}
+
+/**
+ * Create a new Piñata client instance from the passed environment variables.
+ * @param {Record<string, string|undefined>} env
+ */
+function getPinata (env, { reqsPerSec = 2 }) {
```
vasco-santos, line 43: can't we default to 3?
```suggestion
function getPinata (env, { reqsPerSec = 3 }) {
```

packages/pinpin/package.json
olizilla: Currently all the logging is done via debug, and it's flakey so it's useful to see what's going on. There should really be some log levels here, so you can see the job is running (info), vs seeing every request (debug), vs something failed (error), but it's really not critical at the moment.

packages/pinpin/src/bin.js
olizilla: yes. this was carried over from when i thought the rate limit was 2. I intend to still pass in the limit explicilty from the calling code so it's obvious from the top level.

packages/pinpin/README.md
olizilla: sleepless rambling

packages/pinpin/README.md
olizilla: where does the `.env.local` pattern come from? we're not using the .env file for production, just as a convenience for developers... also notable https://github.com/motdotla/dotenv#should-i-have-multiple-env-files

vasco-santos: We already use the `.env.local` pattern in our codebase. The pattern was probably doing like nft.storage:

- https://github.com/web3-storage/web3.storage/blob/6cf8571c79505db4769a40bf42b36d484d315550/packages/api/webpack.config.js#L9
- https://github.com/web3-storage/web3.storage/blob/dfeffd6f4d529fe5765ee0ea12d0a813f2e951af/packages/db/tools/import-graphql-schema.js#L5

And we ignore it to be commited, while a .env file is whitelisted 
- https://github.com/web3-storage/web3.storage/blob/main/.gitignore#L26

andrew (approved):

packages/pinpin/README.md
olizilla: aye, i think my point was that we're not using a .env file in production, so we don't need the destinction between local and prod here. the env file is just to make it easier for a dev to set the env vars.

packages/pinpin/README.md
olizilla: Fixed!

packages/pinpin/README.md
```diff
@@ -30,6 +31,22 @@ npm run start
...
+Deploying from scratch looks like this:
+
+```console
+# create the app as a resource under the web3-storage team, and add `heroku` as git remote
+$ heroku create --team=web3-storage web3-storage
+
+# set the env (or better, use the webui for secrets)
+$ heroku config.set FAUNA_KEY=*** PINATA_JWT=***
+
+# deploy the local pinpin branch to heroku via git (you might push from a local main branch instead.)
```
andrew, line 45: once this PR is merged, the pinpin branch will be no more?

olizilla: i was documenting what i did, as I'm not expecting to do anything else on this or even set up CI on it for the foresable future, so i'll probably want to deploy a local branch name again if there is a next time.

packages/pinpin/README.md
vasco-santos: My concern is that adding this can easily leak keys into github repo, given `.env` file is not in .gitginore and can be unintentionally committed quite easily. But this is a nit

vasco-santos (approved):

olizilla:
@vasco-santos i understand the concern, but the .env file is added to the .gitignore in the pinpin folder for that reason

Issue #416, opened 2021-08-28
Add support for upload rename


vasco-santos:
Anecdotally, renaming a file is something common for an user. Given there is support for naming the upload, this name should be easily renamed.

A few examples:
- when adding an upload via the website the upload name will be the file name, but the user might not want the actual name of its local file system
- when uploading a large file via CLI and the user forgets to add the optional name flag, the upload will be named by default as `Upload at <DATE>`. This is good, but how likely will this date resonate with the user after several other uploads?

PR #420, opened 2021-08-30
feat: allow uploads to be renamed

rafaelramalho19:
Closes https://github.com/web3-storage/web3.storage/issues/416

- This PR allows for awesome and quick edit of a upload's name 🎉 

https://user-images.githubusercontent.com/24696635/131395673-88d5eb45-64e7-4ec7-b6d1-518e817a72f9.mov

### DIFF
--- a/packages/db/fauna/resources/Function/renameUserUpload.js
+++ b/packages/db/fauna/resources/Function/renameUserUpload.js
@@ -1,0 +1,67 @@
+import fauna from 'faunadb'
+
+const {
+  Abort,
+  Collection,
+  CreateFunction,
+  Exists,
+  Function,
+  Get,
+  If,
+  Index,
+  IsNonEmpty,
+  Lambda,
+  Let,
+  Match,
+  Query,
+  Ref,
+  Select,
+  Update,
+  Var
+} = fauna
+
+const name = 'renameUserUpload'
+const body = Query(
+  Lambda(
+    ['userId', 'cid', 'name'],
+    Let(
+      {
+        userRef: Ref(Collection('User'), Var('userId')),
+        contentMatch: Match(Index('unique_Content_cid'), Var('cid'))
+      },
+      If(
+        IsNonEmpty(Var('contentMatch')),
+        Let(
+          {
+            content: Get(Var('contentMatch')),
+            dagSize: Select(['data', 'dagSize'], Var('content'))
+          },
+          Let(
+            {
+              uploadMatch: Match(
+                Index('upload_by_user_and_content'),
+                Var('userRef'),
+                Select('ref', Var('content')),
+                true
+              )
+            },
+            If(IsNonEmpty(Var('uploadMatch')),
+              Update(
+                Select(['ref'], Get(Var('uploadMatch'))),
+                { data: { name: Var('name') } }
+              ),
+              Abort('upload not found')
+            )
+          )
+        ),
+        Abort('not found')
+      )
+    )
+  )
+)
+
+export default If(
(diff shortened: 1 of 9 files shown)


packages/db/fauna/resources/Function/renameUserUpload.js
```diff
@@ -0,0 +1,67 @@
...
+      {
+        userRef: Ref(Collection('User'), Var('userId')),
+        contentMatch: Match(Index('unique_Content_cid'), Var('cid'))
+      },
+      If(
+        IsNonEmpty(Var('contentMatch')),
+        Let(
+          {
+            content: Get(Var('contentMatch')),
+            dagSize: Select(['data', 'dagSize'], Var('content'))
```
vasco-santos, line 37:
```suggestion
```

packages/db/fauna/resources/Function/renameUserUpload.js
```diff
@@ -0,0 +1,67 @@
...
+    Let(
+      {
+        userRef: Ref(Collection('User'), Var('userId')),
+        contentMatch: Match(Index('unique_Content_cid'), Var('cid'))
+      },
+      If(
+        IsNonEmpty(Var('contentMatch')),
+        Let(
+          {
+            content: Get(Var('contentMatch')),
```
vasco-santos, line 36:
```suggestion
            content: Get(Var('contentMatch'))
```

vasco-santos (changes requested):
Looking good, just some remainings on Fauna UDF

vasco-santos (approved):
LGTM

packages/db/fauna/resources/Function/renameUserUpload.js
```diff
@@ -0,0 +1,66 @@
...
+                Index('upload_by_user_and_content'),
+                Var('userRef'),
+                Select('ref', Var('content')),
+                true
+              )
+            },
+            If(IsNonEmpty(Var('uploadMatch')),
+              Update(
+                Select(['ref'], Get(Var('uploadMatch'))),
+                { data: { name: Var('name') } }
```
alanshaw, line 50: Would be great if this set `updated` - maybe add to schema if not already?

packages/db/fauna/resources/Function/renameUserUpload.js
```diff
@@ -0,0 +1,66 @@
...
+            {
+              uploadMatch: Match(
+                Index('upload_by_user_and_content'),
+                Var('userRef'),
+                Select('ref', Var('content')),
+                true
+              )
+            },
+            If(IsNonEmpty(Var('uploadMatch')),
+              Update(
```
alanshaw, line 48: Check for non-empty name?

vasco-santos: hummm, shouldnt' upload name be required instead? I thought it was, but just confirmed it is not required in the upload collection schema

alanshaw: I think that's ok. I was just thinking about the UX of seeing blank as the upload name, but I guess it could show "No Name" or something.

I think this is not required. Folks may want to remove the name...lets not stop them from doing that.

packages/db/fauna/resources/Function/renameUserUpload.js
rafaelramalho19: @alanshaw I avoided doing too much changes in the schema side of things because we're currently migrating to postgres, but I could do it now if needed

PR #422, opened 2021-08-31
chore: update client browser examples

olizilla:
- depend on latest web3.storage
- get rid of package-lock as we always want to use the latest
- switch jwt input to be type password to hide yer secrets

Screenshot 2021-08-31 at 14 44 12

License: (Apache-2.0 AND MIT)


vasco-santos (approved):
LGTM

PR #423, opened 2021-08-31
fix: make refresh fetch recent uploads

### LINKED ISSUE #265: Refresh button not working
I believe this may be because the `before` data is not updated to be "now".

rafaelramalho19:
- Closes #265

### DIFF
--- a/packages/website/pages/files.js
+++ b/packages/website/pages/files.js
@@ -266,6 +266,11 @@ export default function Files({ isLoggedIn }) {
     }
   )
 
+  const refreshUploads = async () => {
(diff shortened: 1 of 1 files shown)


alanshaw:
It would be good if this didn't take you back to the first page if you're on page 2 for example.

olizilla:
@rafaelramalho19 i think the problem you identify in https://github.com/web3-storage/web3.storage/pull/423#issuecomment-912603377 is ok. I can live with that.

Is this PR otherwise ready?

olizilla (approved):

PR #427, opened 2021-09-01
Add putCar examples for node.js

### LINKED ISSUE #367: Add example for putCar
Create an example for `putCar` including using a non-default codec (`dag-json`/`dag-jose` for example) to encode data.

refs https://github.com/web3-storage/web3.storage/pull/365#pullrequestreview-732706918

yusefnapora:
This adds two new examples in `packages/client/examples/node.js`

- `put-car-from-fs.js` adds a CAR from a small fixture file
- `put-car-dag-cbor.js` makes CARs with dag-cbor data and adds them

closes #367


packages/client/examples/node.js/put-car-dag-cbor.js
```diff
@@ -0,0 +1,106 @@
...
+  // add the blocks to the CAR and close it
+  for (const block of ipldBlocks) {
+    writer.put(block)
+  }
+  writer.close()
+
+  // create a new CarReader we can hand to Web3.Storage.putCar
+  const reader = await CarReader.fromIterable(out)
+  return reader
+}
```
alanshaw, line 102:
```suggestion
async function makeCar (rootCID, ipldBlocks) {
  return new CarReader(1, [rootCID], ipldBlocks)
}
```

...or since it's a 1 liner just inline this function?

I'm not 100% sure this is better since it doesn't create a CAR file...but it's certainly less code.

alanshaw (approved):

Issue #421, opened 2021-08-31
Add end-to-end tests for uploading content and run them automatically and often


olizilla:
Add end-to-end tests for
- [x] uploading via http
- [x] uploading via client
- [ ] uploading via website (P3)
- [ ] reading newly uploaded file from dweb.link

Run them every n minutes, and raise an alert on failure. For bonus points, time how long each run takes and push it to prometheus so we an chart upload perf over time.

olizilla:
`w3` has an e2e test already, but it's not run as part of ci yet https://github.com/web3-storage/web3.storage/blob/main/packages/w3/test/e2e.spec.js

olizilla:
We now have a cron job that runs regulary and uploads to the site via w3 -> js client -> http api, and we run a similar test against staging whenever a change to one of the components is merged.

olizilla:
The website smoke test requires running a browser in CI. @hugomrdias recommends [Playwright Test](https://playwright.dev/). This looks like it'll work as it supports both automated file upload via the browser, and has guidance for dealing with authentication. For this I think we need:

- beforeAll step that logs in to a bot account on GitHub. Let's create a bot account on github for `support@web3.storage`
- see: https://playwright.dev/docs/auth#automate-logging-in
- open the hompage, wait for and click the login button
- click the github login button... we'll need to login in manually the first time to approve the app in github, but subsequent runs _should_ log in automatically. This is the biggest unknown... we need to verify that we can make auth work in a browser test.
- click the Files link
- click the add files link
- click the file input and set the file upload to use a local test fixture... we should programtically generate that so that we test uploading a "new" file that web3.storage doesn't already have in it's db yet.
- sumbit the form and verify that the website takes you to the files list and that the new upload is in the list
- For completeness a second test should be added where we upload a known file and verify the CID matches what we expect... this would need us to add a hook in the upload code of the website, that provides the CID response from the site to test code.

PR #445, opened 2021-09-03
chore: test upload via w3 on cron

olizilla:
The pit canary. A minimally viable smoke test for uploading to web3.storage that can run frequently and alert us if it dies.

In it's current form it will test both staging and production with small files and random files sligtly larger than the local add threshold. I'd like to replace the random files with wikipedia pages or similar, but this will do for now.

it'll let us know when the latest published `w3` command, using the latest published `web3.storage` client is working, which is nice.

WIP on https://github.com/web3-storage/web3.storage/issues/421

License: (Apache-2.0 AND MIT)

### DIFF
--- a/.github/workflows/cron-test.yml
+++ b/.github/workflows/cron-test.yml
@@ -1,0 +1,32 @@
+name: Cron Pins
+
+on:
+  schedule:
+    - cron: '*/10 * * * *'
+
+jobs:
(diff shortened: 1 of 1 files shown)


alanshaw (approved):
Tiny fix but LGTM otherwise. It would be _really_ nice to store something useful.

Issue #433, opened 2021-09-02
Make delete button same size as copy button on tokens listing page


atopal:
Currently the button size is dictated by the text length, and it's uneven. @alanshaw for more context potentially.

alanshaw:
<img width="236" alt="Screenshot 2021-09-01 at 15 51 38" src="https://user-images.githubusercontent.com/152863/131868335-42c92daa-8f0b-49e1-8631-5aec34fd15d6.png">

PR #446, opened 2021-09-03
fix: button size on the tokens page

rafaelramalho19:
Closes https://github.com/web3-storage/web3.storage/issues/433

### DIFF
--- a/packages/website/pages/tokens.js
+++ b/packages/website/pages/tokens.js
@@ -74,14 +74,14 @@ const TokenRow = ({ token, index, copied, onCopy, deleting, onDelete }) => {
       </TableElement>
       <TableElement {...sharedArgs} breakAll={false}>
         <form onSubmit={e => { e.preventDefault(); onCopy(token) }}>
-          <Button type="submit" small className="w-28">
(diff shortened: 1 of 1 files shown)


packages/website/pages/tokens.js
```diff
@@ -74,14 +74,14 @@ const TokenRow = ({ token, index, copied, onCopy, deleting, onDelete }) => {
       </TableElement>
       <TableElement {...sharedArgs} breakAll={false}>
         <form onSubmit={e => { e.preventDefault(); onCopy(token) }}>
-          <Button type="submit" small className="w-28">
+          <Button type="submit" small className="w-24">
```
alanshaw, line 77: This is smaller than before. I believe it was this wide to allow "Copied!" to fit. Can you confirm that "Copied!" fits within this new smaller size, and the same for "Deleting..." below?

In general, please can we have screenshots with UI fixes like this? 🙏

rafaelramalho19: Of course, I'll add screenshots or videos to next UI fixes :)

Here's how it looks like now:
image

alanshaw (approved):

PR #447, opened 2021-09-03
chore: use prod api when adding to web3

olizilla:
Ensure that the add-to-web3 github action uses the prod api.

License: (Apache-2.0 AND MIT)


olizilla:
This should fix the CI job failures we see in https://github.com/web3-storage/web3.storage/pull/423/checks?check_run_id=3507335903

olizilla:
The check is green!

PR #449, opened 2021-09-06
chore: fix upload test cron

olizilla:
- Simplify and fix cron end-to-end test. It just tests against production now.
- Add an e2e upload test as the last step in CI for each of the relevant components [api,client,db,w3]

License: (Apache-2.0 AND MIT)


alanshaw (approved):

olizilla:
Ok, this PR now runs an end-to-end upload test against staging at the end of the existing CI jobs for client,api,db or w3 command. That simplifies the cron upload test, which is intended to be used just for telling us that prod is healthy.

hugomrdias (approved):

PR #460, opened 2021-09-09
Always show header links & login/out button

yusefnapora:
This makes the header less janky by always rendering the links to the Files and Account pages, which now redirect to `/login/` instead of `/` if you're not logged in.

It also renders a login button with a loading spinner while it's checking the auth state, instead of rendering nothing and popping in the button once the state resolves.

@olizilla you mentioned doing this yesterday, and I figured it would be a good stop-gap until we can make the actual auth flow faster.

### DIFF
--- a/packages/website/components/button.js
+++ b/packages/website/components/button.js
@@ -17,7 +17,7 @@ import countly from '../lib/countly'
  * @prop { import('react').MouseEventHandler<HTMLButtonElement> } [onClick]
  * @prop {string} [href]
  * @prop {import('react').ButtonHTMLAttributes<HTMLButtonElement>["type"]} [type]
- * @prop {import('react').ReactChildren | string} children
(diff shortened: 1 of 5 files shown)


olizilla:
Very excited to try this out. Minor, but please always include screenshots or recordings of visual changes to show off what you did!

yusefnapora:
@olizilla good call :)

zebateira (approved):
This is great! Good step forward to present the pending auth state to the user 🚀

olizilla:
@yusefnapora this is great and i love it. Also a nudge to use semantic commit messages for you PR titles, so that the changes get picked up by the release-please bot and added to change logs... and most important, so they can be released to production! In this case i just have to merge another website PR, and which will create a release PR that will let me trigger a prod deploy.

yusefnapora:
Thanks for the nudge @olizilla, will do.

PR #462, opened 2021-09-09
chore: track upload test results in prometheus

olizilla:
this will allow us to alert on failed tests from prometheus and track upload durations.

The metric:

```
web3storage_upload_duration_seconds{endpoint="/car",size="3MiB",cmd="w3"} 4.59
```

tells us how long in seconds it took to upload a 3MiB file to the /car endpoint via the w3 command.

it is imagined that we could also test via curl `cmd=curl` and with other sizes `size=100MiB` (careful with the carinality there tho) or other endpoints `endpoint=/files`.

License: (Apache-2.0 AND MIT)

### DIFF
--- a/.github/workflows/cron-test.yml
+++ b/.github/workflows/cron-test.yml
@@ -13,12 +13,22 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: 16
+      - run: npm install -g @web3-storage/w3
       - name: Add small file via w3
         run: |
           echo "$(date --utc --iso-8601=seconds) web3.storage upload test" > ./upload-test-small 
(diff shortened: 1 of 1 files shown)


olizilla:
`PUSHGATEWAY_BASIC_AUTH` is added as a repo secret.

olizilla:
note that i wasn't able to test this in `act` as it uses a light image as it's runner that doesn't include `/usr/bin/time` but i am confident that the github runner does as it's listed here https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#installed-apt-packages

Also you have to explicitly provide the full path to the `time` binary to be able to use flags, as bash provides a built-in time command that has precendence, that lacks those flags.

olizilla:
Merging, to see if it work live!

PR #467, opened 2021-09-10
chore: debug missing upload metric

olizilla:
everything works *except* the run duration metric value is not appering in prom.

License: (Apache-2.0 AND MIT)


olizilla:
it work!? i just added debug logging. wild.

PR #470, opened 2021-09-10
feat: add first page button to the files pagination

rafaelramalho19:


jdiogopeixoto (changes requested):
Looks good!
Only thing that looks odd to me are the buttons being inside the same container as the table (image below). I would move them to be after/outside the table container, that way the scroll only applies to the table.

Screenshot 2021-09-10 at 15 42 07

PR #480, opened 2021-09-17
chore: unify /upload and /car uploads

olizilla:
- Have the API pack uploaded files into a CAR, to make the two paths more similar and ideally simplify the upload to S3 path.
  -  We need to test this out on staging to see how it impacts the total file size folks can upload, as now we have to have the upload and the CAR in memory at the same time.
- Simplify dagSize calcualtion. If it's a dag-pb root, then just derive the cumulative size of the dag in the same way that ipfs does from the sum of the link sizes and the node size. if it's not dag-pb, just use the sum of the block sizes in the CAR. 
  - This means we dont have to walk the entire DAG and we can get a dagSize figure for dags that are split across mutiple CARs (assuming they stick to the treewalk strategy from carbites, and each contain the root block).

License: (Apache-2.0 AND MIT)

### DIFF
--- a/packages/api/src/car.js
+++ b/packages/api/src/car.js
@@ -135,6 +135,7 @@ export async function carPost (request, env, ctx) {
  * @param {import('./env').Env} env
  * @param {import('./index').Ctx} ctx
  * @param {Blob} car
+ * @param {string} [uploadType = 'Car']
  */
 export async function handleCarUpload (request, env, ctx, car, uploadType = 'Car') {
   const { user, authToken } = request.auth
--- a/packages/api/src/car.js
+++ b/packages/api/src/car.js
@@ -1,12 +1,12 @@
 /* eslint-env serviceworker */
 import { gql } from '@web3-storage/db'
-import { CarReader, CarBlockIterator } from '@ipld/car'
+import { CarBlockIterator } from '@ipld/car'
 import { Block } from 'multiformats/block'
 import * as raw from 'multiformats/codecs/raw'
 import * as cbor from '@ipld/dag-cbor'
 import * as pb from '@ipld/dag-pb'
 import retry from 'p-retry'
-import { GATEWAY, LOCAL_ADD_THRESHOLD, DAG_SIZE_CALC_LIMIT, MAX_BLOCK_SIZE } from './constants.js'
+import { GATEWAY, LOCAL_ADD_THRESHOLD, MAX_BLOCK_SIZE } from './constants.js'
 import { JSONResponse } from './utils/json-response.js'
 import { toPinStatusEnum } from './utils/pin.js'
 
@@ -19,14 +19,6 @@ const CREATE_UPLOAD = gql`
         _id
         dagSize
       }
-    }
-  }
-`
-
-const UPDATE_DAG_SIZE = gql`
-  mutation UpdateContentDagSize($content: ID!, $dagSize: Long!) {
-    updateContentDagSize(content: $content, dagSize: $dagSize) {
-      _id
     }
   }
 `
@@ -129,33 +121,57 @@ export async function carGet (request, env, ctx) {
  * @param {import('./index').Ctx} ctx
  */
 export async function carPost (request, env, ctx) {
+  let blob = await request.blob()
+  // Ensure car blob.type is set; it is used by the cluster client to set the foramt=car flag on the /add call.
+  blob = blob.slice(0, blob.size, 'application/car')
+
+  return handleCarUpload(request, env, ctx, blob)
+}
+
+/**
+ * Post a CAR file.
+ *
+ * @param {import('./user').AuthenticatedRequest} request
+ * @param {import('./env').Env} env
+ * @param {import('./index').Ctx} ctx
+ * @param {Blob} car
+ */
+export async function handleCarUpload (request, env, ctx, car) {
   const { user, authToken } = request.auth
   const { headers } = request
 
-  let name = headers.get('x-name')
-  if (!name || typeof name !== 'string') {
-    name = `Upload at ${new Date().toISOString()}`
-  }
+  const bytes = new Uint8Array(await car.arrayBuffer())
 
-  const blob = await request.blob()
-  const bytes = new Uint8Array(await blob.arrayBuffer())
-  const stat = await carStat(bytes)
+  // throw if CAR is invalid
+  await carStat(bytes)
 
-  // Ensure car blob.type is set; it is used by the cluster client to set the foramt=car flag on the /add call.
-  const content = blob.slice(0, blob.size, 'application/car')
-
-  const { cid } = await env.cluster.add(content, {
-    metadata: { size: content.size.toString() },
+  const {
+    cid,
+    bytes: sumOfBlocksInCarSize,
+    size: unixFsCumulativeSize
+  } = await env.cluster.add(car, {
+    metadata: { size: car.size.toString() },
     // When >2.5MB, use local add, because waiting for blocks to be sent to
     // other cluster nodes can take a long time. Replication to other nodes
     // will be done async by bitswap instead.
-    local: blob.size > LOCAL_ADD_THRESHOLD
+    local: car.size > LOCAL_ADD_THRESHOLD
   })
+
+  // If the CAR had a unixFS root, then use the cumlative size property from the root.
+  // Otherwise, cluster gives us the sum of the bytes of every block in the car.
+  // see: https://github.com/ipfs/ipfs-cluster/blob/396a348a6541177a61e91bfd8c1d50ef75822559/adder/adder.go#L296-L308
+  const dagSize = unixFsCumulativeSize || sumOfBlocksInCarSize
+  console.log({ dagSize, unixFsCumulativeSize, sumOfBlocksInCarSize })
 
   const { peerMap } = await env.cluster.status(cid)
   const pins = toPins(peerMap)
   if (!pins.length) { // should not happen
     throw new Error('not pinning on any node')
+  }
+
+  let name = headers.get('x-name')
+  if (!name || typeof name !== 'string') {
+    name = `Upload at ${new Date().toISOString()}`
   }
 
   // Store in DB
@@ -169,7 +185,8 @@ export async function carPost (request, env, ctx) {
         cid,
         name,
         type: 'Car',
-        pins
+        pins,
+        dagSize
       }
     })
   ), {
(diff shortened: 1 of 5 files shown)


olizilla:
Looks like it might be our webpack config trying to split the service worker bundle into chunks, but as it's for upload to the cloudflare that should just be disabled. Am investigating.

olizilla:
Nice! Forcing webpack to not split the bundle fixes the importScripts error.

olizilla:
Of note this pushes the api bundlesize up from [686.35KB](https://github.com/web3-storage/web3.storage/runs/3525715154) to 799.28KB

```
> @web3-storage/api@3.4.0 test:size
> bundlesize

 PASS  ./dist/main.js: 799.28KB < maxSize 1MB (no compression) 
 ```

File: packages/api/src/upload.js [hunk 1]
```diff
@@ -29,90 +11,33 @@ const INCREMENT_USER_USED_STORAGE = gql`
...
-        })
-      } catch (err) {
-        console.error(`failed to update user used storage: ${err.stack}`)
-      }
-    })())
+    input = [blob]
   }
-
-  return new JSONResponse({ cid })
+  // TODO: do we want to wrap file uploads like we do car uploads from the client?
```
vasco-santos, line 39: when receiving a single file, it won't have a path, while directories might have in the name. For directories with several files with no path like name it would likely make sense. I suggest that we should discuss this externally within an issue and keep the current behaviour in this PR context.

packages/api/src/car.js
```diff
@@ -303,48 +300,24 @@ async function carStat (carBytes) {
...
-    const block = await getBlock(cid)
-    let size = block.bytes.byteLength
-    for (const [, cid] of block.links()) {
-      size += await getSize(cid)
-    }
-    return size
-  }
-
-  return getSize(rootCid)
+function cumulativeSize (block, links) {
```
vasco-santos, line 319: Can we add jsdoc here? Can you point me where Tsize is? I could not find it in multiformats

packages/api/src/car.js
```diff
@@ -129,35 +121,51 @@ export async function carGet (request, env, ctx) {
...
+  const { cid } = await env.cluster.add(car, {
+    metadata: { size: car.size.toString() },
     // When >2.5MB, use local add, because waiting for blocks to be sent to
     // other cluster nodes can take a long time. Replication to other nodes
     // will be done async by bitswap instead.
-    local: blob.size > LOCAL_ADD_THRESHOLD
+    local: car.size > LOCAL_ADD_THRESHOLD
   })
 
+  console.log({ cid, dagSize })
```
vasco-santos, line 156: I know this is still draft, but do not forget to take it out

packages/api/src/car.js
```diff
@@ -180,32 +189,19 @@ export async function carPost (request, env, ctx) {
   /** @type {(() => Promise<any>)[]} */
   const tasks = []
 
+  // TODO: this should be handled in the db code as part of the CREATE_UPLOAD
```
vasco-santos, line 192: with fauna was problematic. I guess we can do this with postgres ?

packages/api/src/car.js
olizilla: I pulled this from `ipfs-core` here: https://github.com/ipfs/js-ipfs/blob/b5488cb3ef5ab7ea24b8b2275ffa8cfa4b7b7ed1/packages/ipfs-core/src/components/files/stat.js#L121

vasco-santos: Ok, I could not find where this type is documented but I found that it is added in unixfs https://github.com/ipfs/js-ipfs-unixfs/search?q=Tsize

olizilla: this is the most canonical thing I could find https://ipld.io/specs/codecs/dag-pb/spec/#serial-format

vasco-santos: thanks, I would suggest we add a comment with that link to be more clear

olizilla: also https://github.com/ipld/js-dag-pb/blob/998e0b2a501de6e844d1b0e9a86034b474427080/src/interface.ts#L8-L12

olizilla: yeah, this isn't doing what i think it's doing! Thanks for the nudge.

packages/api/src/car.js
```diff
@@ -270,10 +264,11 @@ export async function sizeOf (response) {
  * - Missing non-root blocks (when root block has links)
  *
  * @typedef {{ size: number, blocks: number }} CarStat
- * @param {Uint8Array} carBytes
+ * @param {Blob} carBlob
  * @returns {Promise<CarStat>}
```
vasco-santos, line 268: nit: Given we only use the dag size, should we now return the unnecessary blocks?

packages/api/src/car.js
```diff
@@ -129,27 +121,36 @@ export async function carGet (request, env, ctx) {
...
+  return handleCarUpload(request, env, ctx, blob)
+}
+
+/**
+ * Post a CAR file.
+ *
+ * @param {import('./user').AuthenticatedRequest} request
+ * @param {import('./env').Env} env
+ * @param {import('./index').Ctx} ctx
+ * @param {Blob} car
```
vasco-santos, line 137:
```suggestion
 * @param {Blob} car
 * @param {string} [uploadType = 'Car']
```

packages/api/src/car.js
```diff
@@ -129,27 +121,36 @@ export async function carGet (request, env, ctx) {
...
-  const bytes = new Uint8Array(await blob.arrayBuffer())
-  const stat = await carStat(bytes)
-
-  // Ensure car blob.type is set; it is used by the cluster client to set the foramt=car flag on the /add call.
-  const content = blob.slice(0, blob.size, 'application/car')
-
-  const { cid } = await env.cluster.add(content, {
-    metadata: { size: content.size.toString() },
+  // We can't use the `bytes` from cluster where it's a unixfs root partial, as it'll be a shard of the total dag size.
+  // We can't use the `size` from cluster as it's the unixfs FileSize which is 0 for directories, and is not set for raw encoded files, only dag-pb ones.
```
vasco-santos, line 147: Maybe we can add this knowledge to the function where we compute the dag size. What do you think?
Reading through the code, it seems strange this placement as it is not clear what it relates to

vasco-santos (changes requested):

packages/api/src/car.js
```diff
@@ -129,27 +121,36 @@ export async function carGet (request, env, ctx) {
  * @param {import('./index').Ctx} ctx
  */
 export async function carPost (request, env, ctx) {
+  let blob = await request.blob()
+  // Ensure car blob.type is set; it is used by the cluster client to set the foramt=car flag on the /add call.
+  blob = blob.slice(0, blob.size, 'application/car')
```
vasco-santos, line 126: Given we will send CAR files to cluster either way, shouldn't this live in `handleCarUpload`

packages/api/src/car.js
olizilla: I think once the db migration is done, I'd like to track more metrics about CARs we see, like the number of blocks and the sum of the block sizes vs the dag size, things like that. I'm inclined to keep carStat returning more info for now.

vasco-santos (approved):
LGTM

packages/api/src/car.js
```diff
@@ -129,27 +121,39 @@ export async function carGet (request, env, ctx) {
  * @param {import('./index').Ctx} ctx
  */
 export async function carPost (request, env, ctx) {
+  let blob = await request.blob()
+  // Ensure car blob.type is set; it is used by the cluster client to set the foramt=car flag on the /add call.
```
alanshaw, line 125:
```suggestion
  // Ensure car blob.type is set; it is used by the cluster client to set the format=car flag on the /add call.
```

packages/api/src/car.js
alanshaw: I _think_ you can just do:

```js
cluster.add(blob, { format: 'car' })
```

...and avoid this dance.

https://github.com/nftstorage/ipfs-cluster/blob/0458e04c0b0331fab4f08f0c092b774c71c0ead2/index.js#L41

packages/api/src/car.js
```diff
@@ -129,27 +121,39 @@ export async function carGet (request, env, ctx) {
...
+  blob = blob.slice(0, blob.size, 'application/car')
+
+  return handleCarUpload(request, env, ctx, blob)
+}
+
+/**
+ * Post a CAR file.
```
alanshaw, line 132:
```suggestion
 * Request handler for POST a CAR file.
```

alanshaw (commented):
Not finished review, submitting what I have...

packages/api/src/upload.js
```diff
@@ -29,90 +11,36 @@ const INCREMENT_USER_USED_STORAGE = gql`
...
-    cid = dir.cid
-    dagSize = dir.size
-    type = 'Multipart'
+    const files = form.getAll('file')
+    input = files.map((f) => ({
+      path: f.name,
+      content: f.stream()
+    }))
+  } else if (contentType.includes('application/car')) {
+    throw new Error('Please POST Content-addressed Archives to /car')
```
alanshaw, line 31:
```suggestion
    throw new HTTPError('Please POST Content-addressed Archives to /car', 400)
```

alanshaw (approved):

packages/api/src/car.js
```diff
@@ -303,48 +301,32 @@ async function carStat (carBytes) {
...
-    }
-    return size
-  }
-
-  return getSize(rootCid)
+function cumulativeSize (pbNodeBytes, pbNode) {
+  // NOTE: Tsize is optional, but all ipfs implementations we know of set it.
+  // It's metadata, that could be missing or deliberately set to an incorrect value.
+  // This logic is the same as used by go/js-ipfs to display the cumulative size of a dag-pb dag.
+  return pbNodeBytes.length + pbNode.Links.reduce((acc, curr) => acc + (curr.Tsize || 0), 0)
```
olizilla, line 329: @vasco-santos uh oh. i think this should be:
```diff
-  return pbNodeBytes.length + pbNode.Links.reduce((acc, curr) => acc + (curr.Tsize || 0), 0)
+  return pbNodeBytes.byteLength + pbNode.Links.reduce((acc, curr) => acc + (curr.Tsize || 0), 0)
```

vasco-santos: Any suggestions to reliably test this?

PR #482, opened 2021-09-17
Checkbox error handling

rafaelramalho19:
- Adds Alert component
- Shows extra details to the user when he tries to delete/select an upload that's still geting the dagsize calculated

Demo:

https://user-images.githubusercontent.com/24696635/133822963-9c6d572e-2b00-410b-85c4-a4fd76ec1c42.mov

### DIFF
--- a/packages/website/components/alert.js
+++ b/packages/website/components/alert.js
@@ -1,0 +1,58 @@
+import React, { useState } from 'react'
+import clsx from 'clsx'
+
+/**
+ * @typedef {Object} AlertProps
+ * @prop {string} [className]
+ * @prop {('top'|'bottom')} [position]
+ * @prop {('error'|'success'|'info')} [type]
+ * @prop {Number} [timer] How much seconds before onTimerEnd should be called
+ * @prop {Function} [onTimerEnd]
+ * @prop {JSX.Element} [children]
+ */
+
+const ANIMATION_DURATION = 400
+
+/**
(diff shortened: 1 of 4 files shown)


olizilla:
A user should be able to delete an upload even if the dagSize has not been calcuated yet. I don't think we want to prevent that. Is there an issue connected to this PR where I can read the backstory?

olizilla:
@rafaelramalho19 the work here is good! my concern is just that I dont think we should need it. Also of note we are about to land changes that mean partial uploads will have their full dagSize set even while the other parts are still uploading, so we might need to rethink this.

@vasco-santos can you explain why we have this limitation? Why can't I delete before dagSize is calculated? Do we need some other flag to tell us when we have multiple partial dags still being processed.

vasco-santos:
With the changes from unifying /upload and /car routes, we calculate dag size differently now https://github.com/web3-storage/web3.storage/pull/480

With this, the limitation to wait for dagSize to be computed should not exist anymore. Can you try and confirm @rafaelramalho19 ?

rafaelramalho19:
> With the changes from unifying /upload and /car routes, we calculate dag size differently now #480
> 
> With this, the limitation to wait for dagSize to be computed should not exist anymore. Can you try and confirm @rafaelramalho19 ?

Good call @vasco-santos! Now it's not needed, closing this PR 😄

PR #490, opened 2021-09-23
fix: upload type in upload should be blob or directory

vasco-santos:
The PR to unify /car and /upload https://github.com/web3-storage/web3.storage/pull/480 was merged and changed the Upload type provided to the Database from `Directory` / `Blob` to `Upload`, keeping `Car` to when using the /car endpoint.

This is causing problems in production because `enum type 'UploadType'. Known values are: Car, Blob, Multipart` per https://github.com/web3-storage/web3.storage/blob/website-v1.4.0/packages/db/fauna/schema.graphql#L262-L274

This PR reverts this back to the original behaviour. However, given everything is going to be stored as a Car, maybe the best solution is to either drop this property, or really change its type to `Upload` in the Database schema.

### DIFF
--- a/packages/db/fauna/schema.graphql
+++ b/packages/db/fauna/schema.graphql
@@ -265,6 +265,10 @@ enum UploadType {
   """
   Car
   """
+  Files uploaded and converteded into a CAR file.
+  """
+  Upload
+  """
   A raw blob upload in the request body.
   """
   Blob
--- a/packages/db/fauna/schema.graphql
+++ b/packages/db/fauna/schema.graphql
@@ -265,7 +265,7 @@ enum UploadType {
   """
   Car
   """
-  Files uploaded and converteded into a CAR file.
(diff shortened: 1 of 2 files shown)


olizilla:
Let's introduce a new `Upload` type to the db schema, so we can distinguish between things that were uploaded as files but we turned them into a CAR, from older uploads that were raw files.

packages/db/fauna/schema.graphql
```diff
@@ -265,6 +265,10 @@ enum UploadType {
   """
   Car
   """
+  Files uploaded and converteded into a CAR file.
```
olizilla, line 268:
```suggestion
  Files uploaded and converted into a CAR file.
```

olizilla (changes requested):

olizilla (approved):

PR #501, opened 2021-09-28
fix: increment used storage

vasco-santos:


olizilla (changes requested):
please restore `incrementUserUsedStorage.js` as in theory we'd want to be able to deploy the db and have the current api version still work against it, and in practice our fauna import task only adds and updates things, it does not delete, so removing it from our code base will mean the prod db has an extra function that a dev db may not.

PR #504, opened 2021-09-29
feat: postgres schema

vasco-santos:
This PR adds the postgres schema based on:
- https://github.com/web3-storage/web3-schema/blob/main/schema.sql
- https://github.com/web3-storage/web3.storage/blob/main/packages/db/fauna/schema.graphq
- https://github.com/ipfs-shipyard/nft.storage/blob/main/packages/api/db/tables.sql

image

### DIFF
--- a/packages/db/postgres/tables.sql
+++ b/packages/db/postgres/tables.sql
@@ -122,10 +122,9 @@ CREATE TABLE IF NOT EXISTS upload
   -- User authentication token that was used to upload this content.
   -- Note: nullable, because the user may have used a Magic.link token.
   auth_key_id     BIGINT REFERENCES auth_key (id),
-  -- The root of the uploaded content.
+  -- The root of the uploaded content (base32 CIDv1 normalised).
   content_cid     TEXT                                                          NOT NULL REFERENCES content (cid),
-  -- CID for the content as provided by the user.
-  -- TODO: do we need this? Should we fail if they do not match?
+  -- CID in the from we found in the received file.
   source_cid      TEXT                                                          NOT NULL,
   -- Type of received upload data.
   type            upload_type                                                   NOT NULL,
--- a/packages/db/postgres/tables.sql
+++ b/packages/db/postgres/tables.sql
@@ -1,0 +1,154 @@
+
+-- A user of web3.storage.
+-- TODO: Should we call account?
+CREATE TABLE IF NOT EXISTS user
+(
+  id              BIGSERIAL PRIMARY KEY,
+  name            TEXT                                                          NOT NULL,
+  picture         TEXT,
+  email           TEXT                                                          NOT NULL,
+  -- The Decentralized ID of the Magic User who generated the DID Token.
+  issuer          TEXT                                                          NOT NULL UNIQUE,
+  -- GitHub user handle, may be null if user logged in via email.
+  github          TEXT,
+  -- Cryptographic public address of the Magic User.
+  public_address  TEXT                                                          NOT NULL UNIQUE,
+  -- Used storage in bytes.
+  used_storage    BIGINT DEFAULT 0,
+  inserted_at     TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL,
+  updated_at      TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL
+);
+
+-- User authentication keys.
+CREATE TABLE IF NOT EXISTS auth_key
+(
+  id              BIGSERIAL PRIMARY KEY,
+  -- User assigned name.
+  name            TEXT                                                          NOT NULL,
+  -- Secret that corresponds to this token.
+  secret          TEXT                                                          NOT NULL,
+  -- User this token belongs to.
+  user_id         BIGINT                                                        NOT NULL REFERENCES user (id),
+  inserted_at     TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL,
+  deleted_at      TIMESTAMP WITH TIME ZONE
+);
+
+-- Details of the root of a file/directory stored on web3.storage.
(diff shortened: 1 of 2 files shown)


packages/db/postgres/tables.sql
```diff
@@ -0,0 +1,154 @@
+
+-- A user of web3.storage.
+-- TODO: Should we call account?
```
vasco-santos, line 3: we have /user/account route. nft.storage changed to account

packages/db/postgres/tables.sql
```diff
@@ -0,0 +1,154 @@
...
+  deleted_at      TIMESTAMP WITH TIME ZONE
+);
+
+-- Details of the root of a file/directory stored on web3.storage.
+CREATE TABLE IF NOT EXISTS content
+(
+  -- Root CID for this content. Normalized as v1 base32.
+  cid             TEXT PRIMARY KEY,
+  -- Size of the DAG in bytes. Set if known on upload or for partials is set
+  -- when content is fully pinned in at least one location.
```
olizilla, line 42:
```suggestion
  -- Size of the DAG in bytes. Either the cumulativeSize for dag-pb or the sum of block sizes in the CAR.
```

packages/db/postgres/tables.sql
```diff
@@ -0,0 +1,154 @@
...
+  -- User that uploaded this content.
+  user_id         BIGINT                                                        NOT NULL REFERENCES user (id),
+  -- User authentication token that was used to upload this content.
+  -- Note: nullable, because the user may have used a Magic.link token.
+  auth_key_id     BIGINT REFERENCES auth_key (id),
+  -- The root of the uploaded content.
+  content_cid     TEXT                                                          NOT NULL REFERENCES content (cid),
+  -- CID for the content as provided by the user.
+  -- TODO: do we need this? Should we fail if they do not match?
+  source_cid      TEXT                                                          NOT NULL,
```
olizilla, line 131: is `content_cid` the base32 cidv1 normalised CID, while `source_cid` is the CID in the form we found in the CAR file, could be cidv0 etc?

packages/db/postgres/tables.sql
```diff
@@ -0,0 +1,154 @@
...
+  id              BIGSERIAL PRIMARY KEY,
+  -- Upload that resulted in this backup.
+  upload_id       BIGINT                                                        NOT NULL REFERENCES upload (id),
+  -- Backup url location.
+  url             TEXT                                                          NOT NULL UNIQUE,
+  inserted_at     TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL,
+  deleted_at      TIMESTAMP WITH TIME ZONE,
+)
+
+-- TODO: we need https://github.com/web3-storage/web3.storage/blob/main/packages/db/fauna/schema.graphql#L174-L216
```
vasco-santos, line 154: Only PinSyncRequest should be needed

packages/db/postgres/tables.sql
vasco-santos: That is my assumption, this was grabbed from both nft.storage and https://github.com/web3-storage/web3-schema/blob/main/schema.sql#L54 . Not sure why to keep it though

vasco-santos:
Added DB diagram with proposed schema at https://dbdiagram.io/d/61546519825b5b014618caf6

packages/db/postgres/tables.sql
alanshaw: > is `content_cid` the base32 cidv1 normalised CID, while `source_cid` is the CID in the form we found in the CAR file, could be cidv0 etc?

yes, and content_cid is a FK into the content table

olizilla: @vasco-santos @alanshaw nft.storage is using `cid`  as the primary key for the content table... do we want to match that here? https://github.com/ipfs-shipyard/nft.storage/blob/7102ad5fc06d90a85027e40fa0aece3429b833e7/packages/api/db/tables.sql#L49

packages/db/postgres/tables.sql
```diff
@@ -0,0 +1,159 @@
...
+  deleted_at      TIMESTAMP WITH TIME ZONE
+);
+
+-- A request to keep a Pin in sync with the nodes that are pinning it.
+CREATE TABLE IF NOT EXISTS pin_request
+(
+  id              BIGSERIAL PRIMARY KEY,
+  -- Identifier for the pin to keep in sync.
+  pin_id          BIGINT                                                        NOT NULL REFERENCES pin (id),
+  inserted_at     TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL
```
olizilla, line 158: we dont delete pin_requests?

vasco-santos: Good question!! I think we should delete, but apparently we do not have it in our fauna schema: https://github.com/web3-storage/web3.storage/blob/main/packages/db/fauna/schema.graphql#L176 😱 

Ideas why?

olizilla (approved):
This is great and many thanks to @andrew for the shipping much of this.

I'm up for using db generated primary keys for things, but we should aim to match nft.storage as much as possible so just one question on should we adopt `cid` as the primary key for `content`

Happy to land this asap and we can iterate on it.

packages/db/postgres/tables.sql
vasco-santos: Added as I think this make sense!

packages/db/postgres/tables.sql
vasco-santos: Ok, I updated the comments to make them more clear. I think we should move with `cid` as the primary key yes! and we are already doing it

vasco-santos:
> should we adopt cid as the primary key for content

Yes, we are doing exactly it here and I think we should

vasco-santos:
Getting this merged in. Feel free to still comment and I can do updates as follow up 🚢

PR #505, opened 2021-09-29
fix: source mapping and commit info in sentry

### LINKED ISSUE #499: API Publish with Intermittent script timeout
Sometimes when publishing to CF workers, or starting a CF Worker in Dev mode, the script timeouts after the build finishes.

It is likely related to the S3 client dependency added to the bundle in https://github.com/web3-storage/web3.storage/pull/417 

I could not replicate this locally, but @olizilla can. On latest releases, it happened once and it might become annoying if it is recurrent. Let's track this and if it is frequent, we should probably just rely on [S3 HTTP PutObject API](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html) instead, given all the other most used s3 client are not well suited for a browser environment.

olizilla:
Match up the release name reported during deploy with the release name reported at run time. This PR lines things up so releases appear in sentry like `web3-api@<semver>` which it automagically parses nicely. We currently we deploy as `<sort git sha>` but we report as `api-<semver>`.

see: https://sentry.io/organizations/protocol-labs-it/releases/web3-api%403.5.1/?project=5866408

- `SENTRY_RELEASE` env var is set during the build, and used for both creating the release at build time and on error reports so they get matched up and our sourcemap is used to unravel the stacktrace.
- Rename the bundle to be `worker.js` as that is what cloudflare renames (!?) it to on deploy, and errors are reported as coming from `worker.js` in production, so that has to match or else the source map is not applied.
- remove the `//# sourceMappingURL=...` link from the end of the js bundle by switching to `devtool: hidden-source-map` in the webpack config. I have no idea whey this is necessary. see: https://github.com/robertcepa/toucan-js/issues/54#issuecomment-930416972
- add commit info to releases in a way that sentry understands, and can match up to our repo with `setCommits: { auto: true }` see: https://github.com/getsentry/sentry-webpack-plugin#optionssetcommits

This fixes our double release counting in sentry and make source maps work for _much error readability_!

Screenshot 2021-09-29 at 16 24 53

I have also connected up the repo in Sentry so we can see the commits that go in to the release.

Screenshot 2021-09-29 at 21 09 37

### Follow on work

- We can spare ourselves from cf worker upload errors by just not uploading our 4MB source map as part of deployment. Use the webpack remove files plugin to delete the source map after it has been uploaded to sentry, as we're just not using in cloudflare! would fix #499 
- Add COMMITSHA and BRANCH env vars, and a `/version` endpoint to report it as nft.storage does. https://github.com/ipfs-shipyard/nft.storage/blob/7102ad5fc06d90a85027e40fa0aece3429b833e7/packages/api/src/index.js#L58-L64

License: (Apache-2.0 AND MIT)

### DIFF
--- a/packages/api/src/env.js
+++ b/packages/api/src/env.js
@@ -1,4 +1,4 @@
-/* global MAGIC_SECRET_KEY FAUNA_ENDPOINT FAUNA_KEY SALT CLUSTER_BASIC_AUTH_TOKEN CLUSTER_API_URL SENTRY_DSN, VERSION DANGEROUSLY_BYPASS_MAGIC_AUTH S3_BUCKET_ENDPOINT S3_BUCKET_NAME S3_BUCKET_REGION S3_ACCESS_KEY_ID S3_SECRET_ACCESS_KEY_ID ENV */
+/* global MAGIC_SECRET_KEY FAUNA_ENDPOINT FAUNA_KEY SALT CLUSTER_BASIC_AUTH_TOKEN CLUSTER_API_URL SENTRY_DSN SENTRY_RELEASE DANGEROUSLY_BYPASS_MAGIC_AUTH S3_BUCKET_ENDPOINT S3_BUCKET_NAME S3_BUCKET_REGION S3_ACCESS_KEY_ID S3_SECRET_ACCESS_KEY_ID ENV */
 import Toucan from 'toucan-js'
 import { S3Client } from '@aws-sdk/client-s3'
 import { Magic } from '@magic-sdk/admin'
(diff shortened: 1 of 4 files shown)


olizilla:
Trying to understand why removing the `sourceMappingURL` fixed this... https://github.com/robertcepa/toucan-js/issues/54#issuecomment-930416972

packages/api/src/env.js
```diff
@@ -26,19 +26,16 @@ import pkg from '../package.json'
 export function envAll (_, env, event) {
   env.sentry = (env.SENTRY_DSN || typeof SENTRY_DSN !== 'undefined') && new Toucan({
     dsn: env.SENTRY_DSN || SENTRY_DSN,
-    event,
+    context: event,
```
olizilla, line 29: the`event` property is deprecated in toucan-js https://github.com/robertcepa/toucan-js/releases/tag/v2.5.0

packages/api/src/env.js
```diff
@@ -26,19 +26,16 @@ import pkg from '../package.json'
...
     },
-    version: env.VERSION || VERSION,
-    pkg: {
-      ...pkg,
-      // sentry cannot deal with "/" in version
-      name: pkg.name.replace('@web3-storage/', '')
-    }
+    environment: env.ENV || ENV,
+    release: env.SENTRY_RELEASE || SENTRY_RELEASE,
+    pkg
```
olizilla, line 38: No need to replace the package.name now as sentry will use the value of the `release` option now instead of the package name.

alanshaw (approved):

vasco-santos (approved):
LGTM

Thanks for the detailed process to get to this solution @olizilla  ❤️

PR #506, opened 2021-09-30
chore: dont send source-map to cloudflare

olizilla:
- delete the source map after it has been sent to sentry but before it would be uploaded to cloudflare
- skip building a sourece map in dev
- skip the extracting comments to a license file

builds on #505
may help with #499

License: (Apache-2.0 AND MIT)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20373,7 +20373,7 @@
     },
     "packages/api": {
       "name": "@web3-storage/api",
-      "version": "3.4.0",
+      "version": "3.5.1",
       "license": "(Apache-2.0 AND MIT)",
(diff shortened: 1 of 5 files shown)


olizilla:
hmm, i still occasionally get

```
✨  Build completed successfully!
Script startup timed out.
 [API code: 10021]
```

even with this change in place.

vasco-santos (approved):
Nice! LGTM

Do you know the difference of bundle we were sending before vs now?

olizilla:
worker.js is identical after this change (~1.1MB). The source map fille is ~4MB.

PR #508, opened 2021-09-30
feat: postgres setup

vasco-santos:
This PR adds the local setup for postgres database using docker compose (partially grabbed from nft.storage PRs)


.github/workflows/api.yml
```diff
@@ -7,8 +7,6 @@ on:
       - 'packages/api/**'
       - '.github/workflows/api.yml'
   pull_request:
```
vasco-santos, line 9: We should be able to run CI tests for other branch targets, like here where we target the DB migration branch

olizilla:
Tests passing locally for me! 🎉

vasco-santos:
> Tests passing locally for me! 🎉

Yeah, CF pages job is failing because https://github.com/web3-storage/web3.storage/issues/529

olizilla:
I'm ok with the "dbClient" being a wrapper wround both fauna and postgres currently. I agree that fauna queries should live in their own file. Ideally the postgres implmentation would also live in it's own file and then the wrapper pulls in both implementations for now, so we could more easily compare the two implementations side by side.

olizilla:
>  Probably better in #524 though? which is where I move fauna code around.
yep, sounds good.

PR #524, opened 2021-10-08
chore: api rewire

### LINKED ISSUE #441: Rewire API endpoints to use Postgres
Once #440 is complete we'd need to rewire API endpoints to read / write data into postgres. This may require intermediary step where endpoints target both DBs to ensure smooth migration.

vasco-santos:
Closes #441 

This PR rewires the API to be able to use either Postgres or Fauna DBs when switching an ENV var. It still uses Fauna by default, but has tests and mocks for both of them. I tested it out with wrangler for all the flows I could remember and it is looking good!

Fauna specific logic was abstracted into a temporary fauna client (abstraction from DB main client), so that the DB client is fully used in the API agnostically to the DB used behind the scenes. Per discussion in previous PR, postgres client is also in its own file
- ⚠️ Postgres client is structured in a way that when we want to do the DB breaking change and not support it, we can just move the postgres client to the main folder as `index.js` and we are good to go! I am not a big fan of this design, and I would much rather default export a factory and we could just create an instance of the needed client... But given this is a temporary state of the codebase, I avoided doing a breaking change and just have the main export class ready to be replaced.

Metrics were also added to the DB Client.

Needs:
- [x] https://github.com/web3-storage/web3.storage/pull/508

### DIFF
--- a/packages/api/README.md
+++ b/packages/api/README.md
@@ -17,7 +17,7 @@ One time set up of your cloudflare worker subdomain for dev:
 
     ```sh
     npm install -g localtunnel
-    lt --port 9094 --subdomain USER-cluster-api-web3-storage
+    npm run lt:cluster
     ```
 
     - There is an npm script you can use to quickly establish these tunnels during development:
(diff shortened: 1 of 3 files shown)


vasco-santos:
merging this to unblock follow up PR!

@hugomrdias and @alanshaw please feel free to still look at this. Or more specifically, the most helpful thing for you to have a look once you have a moment is the `functions.sql` file and the db client:

- https://github.com/web3-storage/web3.storage/blob/chore/db-migration/packages/db/postgres/functions.sql
-

PR #538, opened 2021-10-15
feat: maintenance mode

vasco-santos:
This PR adds middleware to prevent writes or prevent reads and writes to the API when maintenance mode is enabled. It allows us to perform migrations where we need to prevent writes to the database/cluster for the duration and as well as giving us a way of preventing writes to cluster/the db when outages occur.

The mode can be enabled by setting an environment variable MAINTENANCE_MODE to one of the following values:

-- = no reading or writing
r- = read only mode
rw = read and write (normal operation)

A similar approach to the authmoji was created to distinct routes needs:

image

For clients (like the website) to consult the mode state of the API, a `/version` endpoint was created to return the current maintenance mode so that a banner can be shown in the UI. It also returns some other nice information like nft.storage does.


vasco-santos:
> I think the maintenance mode middleware needs to be applied after the CORS middleware.

We set CORS headers on error https://github.com/web3-storage/web3.storage/blob/main/packages/api/src/index.js#L75 so I think we do not need to.

PR #539, opened 2021-10-15
fix: encode filenames

### LINKED ISSUE #532: Files with names that contain non ISO-8859-1 code point throw exception
image

Example file name : Scarlett Rose – Medium PWA Firefox.pdf

Would be nice to get a warning that filename contains characters that cannot be 
alternatively change filename and issue warning that the finame has been changed

image

zebateira:
This PR fixes upload not working on the website when the filename includes a non ISO-8859-1 character.
We were missing an encode/decodeURIComponent. This only affects the files because the filename is sent via header `X-NAME`.
I made the fix in the `lib/api.js` file on the website code. Should this done on the js client instead?

Fixes #532

### DIFF
--- a/packages/api/README.md
+++ b/packages/api/README.md
@@ -94,6 +94,8 @@ curl -X POST --data-binary @x.car -H 'Authorization: Bearer YOUR_API_KEY' http:/
 }
 ```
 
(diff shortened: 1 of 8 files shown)


zebateira:
Docs added!
Will PR the docs site in a sec.

PR #547, opened 2021-10-20
feat: add user email to the account page

### LINKED ISSUE #443: Show who you are logged in as in the website
I have multiple web3.storage accounts across multiple instances. I have to derive which account I am looking at from the files that are in the account. Some indication of which account I am logged in to would help!

### LINKED ISSUE #534: Add user info to account page
When a user is logged in, there is no way to tell what the user's account email is (or if they logged in through github).
We should at least show the user's email.

Codigo-Fuentes:
- Now displays user's email address and login method on account page
via-gh
email

⚠️ Awaiting DB migration

Closes #443
Closes #534


zebateira:
image

I logged in with email but it doesn't show my email though 😕

PR #575, opened 2021-10-28
feat(client): add blockchain auth for solana keys

yusefnapora:
This is an early draft of adding wallet auth to the Web3.Storage client, using a signature created by a blockchain private key to authenticate the user and authorize the upload of a specific CID.

I haven't tested anything yet (local env issues 😞) and am just putting this up for feedback on the API design and so on.

The interface is fairly generic, in that there's a `WalletAuthProvider` interface that includes a string tag for the blockchain and type of key used to create the signatures, and you can plug in your own `signMessage` function to create the signatures. If this is too generic we can walk it back and define a concrete `SolanaAuthProvider` or something instead.

The signature is made by creating a payload object that contains the CID being uploaded, the wallet public key (encoded as a CID using the identity hasher), and a chain-specific context object. For solana it contains a recent blockhash, plus the blockchain and network identifiers.

The payload is serialized as a dag-cbor object for deterministic encoding, and we prefix the serialized bytes with a domain separation tag before signing.

The serialized payload (without the separation prefix) and the signature are encoded to base64 with multibase prefix and attached to the request in headers. The signature goes in the `Authorization` header as essentially a `Bearer` token, but with the word `Bearer` replaced with `X-Wallet-Signature-${blockchain}-${network}`, so e.g. `X-Wallet-Signature-solana-mainnet`. The payload goes in an `X-Wallet-Signature-Payload` header.

The user supplies two callbacks in the `WalletAuthProvider`:

- `signMessage` accepts a `Uint8Array` and returns a promise of the signature `Uint8Array`
- `getChainContext` returns (a promise for) the chain-specific context info

These are only invoked if there's no `token` and the endpoint is CID-specific (it will just fail for `list`, etc).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2893,6 +2893,62 @@
         "node": "^10.12.0 || >=12.0.0"
       }
     },
+    "node_modules/@ethersproject/bytes": {
+      "version": "5.5.0",
+      "resolved": "https://registry.npmjs.org/@ethersproject/bytes/-/bytes-5.5.0.tgz",
+      "integrity": "sha512-ABvc7BHWhZU9PNM/tANm/Qx4ostPGadAuQzWTr3doklZOhDlmcBqclrQe/ZXUIj3K8wC28oYeuRa+A37tX9kog==",
+      "dev": true,
+      "funding": [
+        {
+          "type": "individual",
+          "url": "https://gitcoin.co/grants/13/ethersjs-complete-simple-and-tiny-2"
+        },
+        {
+          "type": "individual",
+          "url": "https://www.buymeacoffee.com/ricmoo"
+        }
+      ],
+      "dependencies": {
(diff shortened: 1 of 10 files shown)


yusefnapora:
@mikeal last commit tries to bring this closer to the design in #577 (without the tags at the moment, so there's still a `ChainContext` object). 

Let me know if I'm barking up the wrong tree - overall though I think it makes the design simpler and feels better than what I was using.

We can implement it this way without needing to build out the management side of things (linking pubkeys to an developer's account). We can just hard-code that if your request has the `chain:solana` tag or whatever, then we'll accept `put:cid:*` requests as long as the sig is correct and is a real solana key.

alanshaw:
Is this stale @yusefnapora? Can be closed?

yusefnapora:
yes, thanks for the reminder

PR #585, opened 2021-11-01
fix: add v2 metrics

vasco-santos:


vasco-santos:
I'll add this in https://github.com/web3-storage/db-migration-pipeline/pull/15

PR #604, opened 2021-11-11
fix: bigserial id cast to string

vasco-santos:
Postgres REST returns a number from a Bigserial, which for really large Bigserials gets a different number. This PR casts all the IDs into text

### DIFF
--- a/packages/db/postgres/functions.sql
+++ b/packages/db/postgres/functions.sql
@@ -13,7 +13,34 @@ CREATE OR REPLACE FUNCTION json_arr_to_json_element_array(_json json)
   RETURNS json[] LANGUAGE sql IMMUTABLE PARALLEL SAFE AS
 'SELECT ARRAY(SELECT * FROM json_array_elements(_json))';
 
-CREATE OR REPLACE FUNCTION create_upload(data json) RETURNS BIGINT
+CREATE OR REPLACE FUNCTION create_key(data json) RETURNS TEXT
+    LANGUAGE plpgsql
+    volatile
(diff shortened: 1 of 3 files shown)


packages/db/postgres/functions.sql
```diff
@@ -13,7 +13,34 @@ CREATE OR REPLACE FUNCTION json_arr_to_json_element_array(_json json)
...
+          (data ->> 'user_id')::BIGINT,
+          (data ->> 'inserted_at')::timestamptz,
+          (data ->> 'updated_at')::timestamptz);
+
+  inserted_key_id := (select id
+  from auth_key
+  where name = data ->> 'name'
+        AND user_id = (data ->> 'user_id')::BIGINT
+        AND secret = data ->> 'secret'
+  limit 1);
```
alanshaw, line 37: Is it possible to use the `RETURNING` clause to the `INSERT` instead of issuing a separate `SELECT`...like in `create_upload`?

alanshaw (approved):
LGTM, just one suggestion.

packages/db/postgres/functions.sql
vasco-santos: It seems it does https://www.postgresql.org/docs/9.5/dml-returning.html

I thought it was only possible when we use the on conflict clause for some previous attempts. But no, the problem is when we do "ON CONFLICT () DO NOTHING". If we use ON CONFLICT we need to update to have return, but simple insert works 😉 

Thanks, just running tests and will commit

PR #611, opened 2021-11-12
Chore/fix types

flea89:


vasco-santos (commented):
thank you so much for getting these updated

packages/db/postgres/client.js
```diff
@@ -120,7 +121,7 @@ export class PostgresClient {
    */
   async createUpload (data) {
     const now = new Date().toISOString()
-    /** @type {{ data: number, error: Error }} */
+    /** @type {{ data: number, error: PostgrestError }} */
```
alexandrastoica, line 124: Is data a string (instead of number)?

alexandrastoica (approved):
Looks ok to me

packages/db/postgres/client.js
vasco-santos: good catch! Probably the same error in several other client rpc calls

packages/db/postgres/client.js
```diff
@@ -20,7 +20,8 @@ const uploadQuery = `
         content(cid, dagSize:dag_size, pins:pin(status, updated:updated_at, location:pin_location(_id:id, peerId:peer_id, peerName:peer_name, region)))
       `
 /**
- * @typedef {import('./postgres/pg-rest-api-types').definitions} definitions
+ * @typedef {import('./pg-rest-api-types.js').definitions} definitions
+ * @typedef {import('@supabase/postgrest-js').PostgrestError} PostgrestError
```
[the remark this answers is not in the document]
flea89: It's for typecasting as described [here](https://docs.joshuatz.com/cheatsheets/js/jsdoc/#simple-casting)

Actually in this particular instance is not really needed since the problem is that `data` is not in the typedef,  but  I've made a similar change it in 2 different instances in the file ([here](https://github.com/web3-storage/web3.storage/pull/611/files#diff-04117e09fe44d5643c6812b42b94359638abe1e2c2e1e2ba64065c6ca3228819R107) and [here](https://github.com/web3-storage/web3.storage/pull/611/files#diff-04117e09fe44d5643c6812b42b94359638abe1e2c2e1e2ba64065c6ca3228819R356))

And I've realized, instead of casting we should be calling 
`.single()` since in all those places we're actually expecting a single result, isn't it?

`.rpc()` return `PostgrestFilterBuilder<any>` while we'd be expecting is PostgrestSingleResponse<any>

What do you think?

packages/db/postgres/client.js
```diff
@@ -616,8 +617,8 @@ export class PostgresClient {
    * @return {Promise<import('../db-client-types').AuthKey>}
    */
   async getKey (issuer, secret) {
-    /** @type {{ error: Error } */
-    const { data, error } = await this._client
+    /** @type {{ error: PostgrestError } */
+    const { data, error } = (await this._client
```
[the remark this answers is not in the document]
flea89: I added the comment in the wrong place, moving it here.

In this particular instance that's actually not required, the problem is our type definition is missing the `data` field, I'll change that.

However I've done a similar change in 2 other instances and the aim was typecasting as described [here]((https://docs.joshuatz.com/cheatsheets/js/jsdoc/#simple-casting).)

But by reviewing the other 2 changes ( [this one](https://github.com/web3-storage/web3.storage/pull/611/files#diff-04117e09fe44d5643c6812b42b94359638abe1e2c2e1e2ba64065c6ca3228819R107) and [this one](https://github.com/web3-storage/web3.storage/pull/611/files#diff-04117e09fe44d5643c6812b42b94359638abe1e2c2e1e2ba64065c6ca3228819R356)) instead of casting we should change the client call and add `.single()` instead.

`.rpc()` returns a `PostgrestFilterBuilder<any>` while we should expect a `PostgrestSingleResponse<any>` instead given we're assuming a single result. Isn't it? (Which also means we wouldn't need to cast)

What you think?

packages/db/postgres/client.js
flea89: Good catch, I'll update it.

packages/db/postgres/client.js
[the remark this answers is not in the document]
vasco-santos: I was expecting the `.rpc` call to return a single value instead. But I just confirmed that `createUpload` calls single.

So, this will not change the actual value that is returned, but the types will work. Correct? If so, I think that makes sense to change this and add `.single()`

packages/db/db-client-types.ts
```diff
@@ -236,6 +212,13 @@ export type UploadOutput = definitions['upload'] & {
   deals: Deal[]
 }
 
+export type Location = {
+  id?: definitions['pin_location']['id']
+  peerId: definitions['pin_location']['peer_id']
+  peerName: definitions['pin_location']['peer_name']
```
alexandrastoica, line 218: I think `peerName` can also be null

alexandrastoica (approved):

packages/db/postgres/client.js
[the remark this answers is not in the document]
flea89: indeed

packages/db/index.js
```diff
@@ -36,11 +37,11 @@ export class DBClient {
   /**
    * Upsert user.
    *
-   * @param {import('../db-client-types').UpsertUserInput} user
-   * @return {import('../db-client-types').UpsertUserOutput}
+   * @param {import('./db-client-types').UpsertUserInput} user
+   * @return {Promise.<import('./db-client-types').UpsertUserOutput>}
```
vasco-santos, line 41:
```suggestion
   * @return {Promise<import('./db-client-types').UpsertUserOutput>}
```

packages/db/index.js
```diff
@@ -102,8 +103,8 @@ export class DBClient {
    * @returns {Promise<number>}
    */
   async getUsedStorage (userId) {
-    /** @type {{ data: number, error: Error }} */
-    const { data, error } = await this._client.rpc('user_used_storage', { query_user_id: userId })
+    /** @type {{ data: number, error: PostgrestError }} */
```
vasco-santos, line 106:
```suggestion
    /** @type {{ data: string, error: PostgrestError }} */
```

vasco-santos (changes requested):

flea89 (commented):
Accepted the suggestions, thanks

vasco-santos (approved):
LGTM

PR #612, opened 2021-11-12
fix: create new types only if they do not exist

flea89:
Running `node scripts/cli.js db-sql` without `--reset` is currently failing because we're trying to recreate types that already exist, we should check first.

I'm not sure if there are better implementation to check if a type already exists 🤔

### DIFF
--- a/packages/db/postgres/tables.sql
+++ b/packages/db/postgres/tables.sql
@@ -45,12 +45,13 @@ CREATE TABLE IF NOT EXISTS content
 
 CREATE INDEX IF NOT EXISTS content_updated_at_idx ON content (updated_at);
 
--- IPFS Cluster tracker status values.
--- https://github.com/ipfs/ipfs-cluster/blob/54c3608899754412861e69ee81ca8f676f7e294b/api/types.go#L52-L83
--- TODO: nft.storage only using a subset of these: https://github.com/ipfs-shipyard/nft.storage/blob/main/packages/api/db/tables.sql#L2-L7
 DO $$
 BEGIN
   IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = 'pin_status_type') THEN
+    
+    -- IPFS Cluster tracker status values.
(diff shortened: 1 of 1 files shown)


packages/db/postgres/tables.sql
```diff
@@ -48,34 +48,53 @@ CREATE INDEX IF NOT EXISTS content_updated_at_idx ON content (updated_at);
...
-  -- The item has been queued for pinning on the IPFS daemon.
-  'PinQueued',
-  -- The item has been queued for unpinning on the IPFS daemon.
-  'UnpinQueued',
-  -- The IPFS daemon is not pinning the item through this CID but it is tracked
-  -- in a cluster dag
-  'Sharded'
-);
+DO $$
+BEGIN
```
alanshaw, line 52: Do we need this outer `DO` block?

flea89: As far as I understand it, IF statements are part of pgSQL.
So in order to be used they need to be part of a function or block statement, hence the use of a `DO` block.

I'm not an expert in pgSQL though, so really happy to be proven wrong and be advised on better ways to achieve this.
Do you have anything in mind?

vasco-santos: I tried other alternatives but I think that is as @flea89 says and we must use the DO block

packages/db/postgres/tables.sql
```diff
@@ -48,34 +48,53 @@ CREATE INDEX IF NOT EXISTS content_updated_at_idx ON content (updated_at);
 -- IPFS Cluster tracker status values.
 -- https://github.com/ipfs/ipfs-cluster/blob/54c3608899754412861e69ee81ca8f676f7e294b/api/types.go#L52-L83
 -- TODO: nft.storage only using a subset of these: https://github.com/ipfs-shipyard/nft.storage/blob/main/packages/api/db/tables.sql#L2-L7
```
vasco-santos, line 50: Can we move this to right above the pin_status_type create statement?

vasco-santos (approved):
LGTM

PR #618, opened 2021-11-15
fix: add indexes and optimise token list function

alanshaw:


vasco-santos (approved):
LGTM

ribasushi (approved):

ribasushi:
@vasco-santos perhaps a TODO to remove the `...deleted IS NULL` stuff, and also return the CASE / END

alanshaw:
Here's an issue for that TODO: https://github.com/web3-storage/web3.storage/issues/619

vasco-santos (approved):

PR #625, opened 2021-11-15
fix: performance metrics queries

vasco-santos:
Improves metrics related queries to get the API endpoint working:

- supabase count was not properly doing the count, but getting all the rows instead, which was too slow.
- added an unique index to cid including `dag_size` to speed up `pin_dag_size_total` rpc function.

All thanks to @ribasushi 

New function already created into production DB


ribasushi (approved):
@vasco-santos perhaps the problem was that the `'*'` was supposed to be `'count(*)'` who knows...

PR #632, opened 2021-11-16
chore: add missing indexes for fk

vasco-santos:
This PR adds missing indexes for all foreign keys. They were already added in the DB

### DIFF
--- a/packages/db/postgres/tables.sql
+++ b/packages/db/postgres/tables.sql
@@ -108,6 +108,8 @@ CREATE TABLE IF NOT EXISTS pin
   UNIQUE (content_cid, pin_location_id)
 );
 
+CREATE INDEX IF NOT EXISTS pin_content_cid_idx ON pin (content_cid);
+CREATE INDEX IF NOT EXISTS pin_location_id_idx ON pin (pin_location_id);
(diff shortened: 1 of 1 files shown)


packages/db/postgres/tables.sql
```diff
@@ -108,6 +108,8 @@ CREATE TABLE IF NOT EXISTS pin
   UNIQUE (content_cid, pin_location_id)
 );
 
+CREATE INDEX IF NOT EXISTS pin_content_cid_idx ON pin (content_cid);
```
vasco-santos, line 111: already existing before, but was not added here

packages/db/postgres/tables.sql
```diff
@@ -147,8 +149,9 @@ CREATE TABLE IF NOT EXISTS upload
   UNIQUE (user_id, source_cid)
 );
 
-CREATE INDEX IF NOT EXISTS upload_updated_at_idx ON upload (updated_at);
 CREATE INDEX IF NOT EXISTS upload_auth_key_id_idx ON upload (auth_key_id);
+CREATE INDEX IF NOT EXISTS upload_content_cid_idx ON upload (content_cid);
```
vasco-santos, line 153: already existing before, but was not added here

ribasushi (approved):

PR #639, opened 2021-11-16
feat: add website version and commit to meta tags

### LINKED ISSUE #448: Add a build git commit sha and release version to the website
To help us diagnose deployment issues, let's add the git commit sha and website release version number to the website somewhere. Setting it just as metadata in the head or an html comment would be fine or we could render it in the page too if folks want to see it.

vasco-santos:
This PR adds api version and commit to website meta tags

image

Closes #448

### DIFF
--- a/packages/website/components/layout.js
+++ b/packages/website/components/layout.js
@@ -11,7 +11,7 @@ import { useQuery } from 'react-query'
 /**
  * @param {any} highlightMessage
  */
-const MessageBanner = ({ highlightMessage }) => {
+const MessageBanner = ({ highlightMessage, apiVersionData }) => {
   let maintenanceMessage = ''
 
   const { data: statusPageData, error: statusPageError } = useQuery(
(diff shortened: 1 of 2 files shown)


alanshaw:
I think there's been some miscommunication here. I think it's a good idea to put the git commit in meta tags, but the commit we want is the commit that the website was built from not the commit the API was built from. The website and API can be deployed independently so these commits are unlikely to be the same.

vasco-santos:
Yeah, I did not understand we wanted the website version/hash. Thanks

Just updated the PR:

image

packages/website/components/layout.js
```diff
@@ -8,6 +10,11 @@ import { getVersion } from '../lib/api'
 import { getStatusPageSummary } from '../lib/statuspage-api'
 import { useQuery } from 'react-query'
 
+/**
+ * @global
+ * @typedef {string} COMMITHASH
+ */
+
```
vasco-santos, line 17:
```suggestion
```

alanshaw (approved):

PR #643, opened 2021-11-17
fix: list uploads fetch deals for correct cid

vasco-santos:
The status API was working fine to get the deals, but the same was not true for the list of uploads. When the query was simplified, the `content_cid` was removed from the query and we can rely on the content relation to ask the deals for the correct cid


alanshaw (approved):
Good catch.

PR #649, opened 2021-11-17
Add skeleton for pinning API endpoints

alexandrastoica:
Adds endpoints:

- POST /pins
- GET /pins/:requestId
- GET /pins
- DELETE /pins/:requestId

Adds payload validation and tests

### DIFF
--- a/packages/api/src/pins.js
+++ b/packages/api/src/pins.js
@@ -110,7 +110,7 @@ export async function pinGet (request, env, ctx) {
  * @param {import('./index').Ctx} ctx
  */
 export async function pinsGet (request, env, ctx) {
-  const { cid, name, match, status } = await request.json()
+  const { cid, name, match, status } = request.params
 
   // Normalize cid
   if (cid) {
@@ -148,7 +148,7 @@ export async function pinsGet (request, env, ctx) {
   }
 
   // Validate status
-  if (status && typeof !Array.isArray(status)) {
+  if (status && !Array.isArray(status)) {
     const isValidStatus = status.every(v => STATUS_OPTIONS.includes(v))
     if (!isValidStatus) {
       return new JSONResponse(
--- a/packages/api/src/pins.js
+++ b/packages/api/src/pins.js
@@ -192,3 +192,12 @@ export async function pinDelete (request, env, ctx) {
   // TODO: write logic for deleting pin request
   return new JSONResponse('OK')
 }
+
+/**
+ * @param {import('./user').AuthenticatedRequest} request
+ * @param {import('./env').Env} env
+ * @param {import('./index').Ctx} ctx
+ */
+export async function pinReplace (request, env, ctx) {
+  throw new Error('Not implemented')
+}
--- a/packages/api/src/pins.js
+++ b/packages/api/src/pins.js
@@ -16,7 +16,11 @@ export async function pinPost (request, env, ctx) {
     )
   }
 
-  const normalizedCid = normalizeCid(cid)
+  try {
+    const normalizedCid = normalizeCid(cid)
+  } catch (err) {
+    return new JSONResponse(err)
+  }
 
   // Validate name
   if (name && typeof name !== 'string') {
@@ -35,12 +39,23 @@ export async function pinPost (request, env, ctx) {
   }
 
   // Validate meta
-  if (meta && (typeof meta !== 'object' || Array.isArray(meta))) {
-    return new JSONResponse(
+  if (meta) {
+    const res = new JSONResponse(
       { error: { reason: 'INVALID_PIN_DATA', details: 'Invalid metadata' } },
       { status: 400 }
     )
+    if (typeof meta !== 'object' || Array.isArray(meta)) {
+      return res
+    }
+
+    const hasStringValues = Object.entries(meta).some(([, v]) => typeof v === 'string')
+    if (!hasStringValues) {
+      return res
+    }
   }
+
+  // @todo: write logic for pinning cid
+  return new JSONResponse('OK')
 }
 
 /**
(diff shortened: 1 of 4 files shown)


packages/api/src/pins.js
```diff
@@ -0,0 +1,86 @@
...
+ * @param {import('./user').AuthenticatedRequest} request
+ * @param {import('./env').Env} env
+ * @param {import('./index').Ctx} ctx
+ */
+export async function pinPost (request, env, ctx) {
+  const { cid, name, origins, meta } = await request.json()
+
+  if (!cid) {
+    return new JSONResponse(
+      { error: { reason: `Invalid request id: ${cid}` }},
```
flea89, line 14: Can we be more explicit here and say it's missing instead?

packages/api/src/pins.js
```diff
@@ -0,0 +1,86 @@
...
+      { status: 400 }
+    )
+  }
+
+  try {
+    const normalizedCid = normalizeCid(cid)
+  } catch (err) {
+    return new JSONResponse(err)
```
flea89, line 22: IINW the error format here is slightly different than the one in the spec

```{ error: { reason: string, [detail]: string }}```

It would be great to be consistent with the errors we send back!
We could event create a type to make ourselves accountable to it ?

Also I suggest we keep the reason to be either a textual description associated with the numeric status code or we come up with a set of predefined values to indicate the error in the context of the api.
In both cases we should add more context of what actually happened in the detail.

File: packages/api/src/pins.js [hunk 3]
```diff
@@ -0,0 +1,86 @@
...
+    return new JSONResponse(
+      { error: { reason: 'INVALID_PIN_DATA', details: 'Invalid origins' } },
+      { status: 400 }
+    )
+  }
+
+  // Validate meta
+  if (meta) {
+    const res = new JSONResponse(
+      { error: { reason: 'INVALID_PIN_DATA', details: 'Invalid metadata' } },
```
flea89, line 44: Could we provide a bit more detail to the user about what is wrong with the metadata based on the scenario we're in?

packages/api/test/pin.spec.js
```diff
@@ -0,0 +1,211 @@
...
+
+function getTestJWT (sub = 'test', name = 'test') {
+  return JWT.sign({ sub, iss: JWT_ISSUER, iat: 1633957389872, name }, SALT)
+}
+
+describe.only('POST /pins', () => {
+  it('should receive pin data containing cid', async () => {
+    const name = 'car'
+    // Create token
+    const token = await getTestJWT()
```
flea89, line 16: Could this be moved to a `before` block and saved in the a variable in describe scope instead to avoid repetition?

packages/api/test/pin.spec.js
```diff
@@ -0,0 +1,211 @@
...
+
+    const res = await fetch(new URL('pins', endpoint).toString(), {
+      method: 'POST',
+      headers: {
+        Authorization: `Bearer ${token}`,
+        'Content-Type': 'application/json',
+        'X-Name': name
```
flea89, line 46: I think we do not need X-name header throughout, do we?

packages/api/test/pin.spec.js
```diff
@@ -0,0 +1,211 @@
...
+    })
+
+    assert(res, 'Server responded')
+    assert(res.status, '400')
+    let data = await res.json()
+    let error = data.error
+    assert(error.reason, 'INVALID_PIN_DATA')
+    assert(error.details, 'Invalid meta')
+
+    res = await fetch(new URL('pins', endpoint).toString(), {
```
flea89, line 189: Maybe be this could be split up in a different test to keep them a bit more manageable?

flea89 (commented):
Great to see this taking shape!
Left a few comments, happy to discuss further

packages/api/src/utils/json-response.js
```diff
@@ -1,7 +1,13 @@
...
+
+// TODO: move this to pin.js
+/**
+ * @typedef {{ error: { reason: string, details?: string } }} PinDataError
+ */
+
 export class JSONResponse extends Response {
   /**
-   * @param {BodyInit} body
+   * @param {BodyInit | PinDataError} body
```
flea89, line 10: JSONResponse is used across all the endpoints. While having more strict parameter's types has its own advantages I wonder if keeping it looser just for now (with something like `{Object.<string, number>}`) would be a bit easier for now given the types are currently broken all over the place?

Don't have a strong opinion, just thinking aloud really.

packages/api/test/pin.spec.js
```diff
@@ -0,0 +1,257 @@
...
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify({})
+    })
+
+    assert(res, 'Server responded')
+    assert(res.status, `${ERROR_CODE}`)
+    const data = await res.json()
+    const error = data.error
+    assert(error.reason, INVALID_CID)
```
flea89, line 52: Shouldn't this be `REQUIRED_CID`?

`assert()` is an alias for `assert.ok` which checks if the first parameter is truthy. 
Can please review/update the assertions across this file?

flea89 (commented):
Thanks for actioning the comments! I left a few more.

As you suggested, we should probably look into refactoring validation but that can definitively be done in a follow-up.

packages/api/test/pin.spec.js
alexandrastoica: Good catch... 😭

packages/api/src/utils/json-response.js
alexandrastoica: Do you mean adding `{Object.<string, number>}` instead of `PinDataError`? I'm ok with that, however it does give type error in `pins.js`

flea89: sorry I meant 
```suggestion
   * @param {Object.<string, object>} body
```

flea89: I mean {Object.<string, any>} however, I now realized that might not be the best as well. 
This might need more discussion. 

Shall we roll it back for now and tackle it in a separate issue?

PR #650, opened 2021-11-17
Chore/npm scripts

flea89:
Goal of MR is to simplify onboarding for new members:

- Make it easier for new members to setup the cluster
- Npm scripts to start/stop db (and persist or not data)
- Npm script to run api tests in debug mode


flea89:
@vasco-santos put this together, it's not super polished, but let me know if that's something you might find useful.

vasco-santos:
Thanks, overall I think this is helpful! I would prefer to not have the docker folder in the root though. Any other options you can think of? Usually, we would have it inside the package, but I understand the need to use it in multiple contexts

flea89:
I see your point, not sure about the right place though 🤔 Maybe give the folder a different name would help?

The whole point of having this is to make it super easy to run the cluster locally. 
If it ends up in a different repo it's not a lot better than now, where you need to follow the links from the readme, isn't it?

The other advantage of having it in the repo is to be able for instance to use it for running the cluster for tests.

packages/api/package.json
```diff
@@ -15,6 +15,7 @@
     "build": "WEBPACK_CLI_FORCE_LOAD_ESM_CONFIG=true webpack",
     "test": "npm-run-all -p -r mock:cluster mock:pgrest mock:backup test:e2e -s test:size",
     "test:e2e": "playwright-test \"test/**/*.spec.js\" --sw src/index.js -b webkit",
+    "test:e2e:debug": "playwright-test \"test/**/*.spec.js\" --debug --sw src/index.js -b webkit",
```
alanshaw, line 18:
```suggestion
    "test:e2e:debug": "npm run test:e2e -- --debug",
```

alanshaw (commented):
Do we need to include `docker/cluster/ipfs-cluster-ctl_v0.14.1_darwin-amd64.tar.gz`?

packages/api/package.json
flea89: Ack!

On a related note, is there any specific reason we have webkit, instead of chromium here?
I find the dev experience in chromium a bit nicer, is it ok to change it?

flea89:
> Do we need to include `docker/cluster/ipfs-cluster-ctl_v0.14.1_darwin-amd64.tar.gz`?
Nope, I don't think we do, I'm getting rid of it.


> @flea89 @alanshaw should we have a tools package similarly to what we have in nft.storage where cluster specific things would live?
@vasco-santos  I like the tools folder idea, let me move things around

packages/api/package.json
vasco-santos: we can't 😢  there is no support for chromium with the multipart upload [context](https://github.com/web3-storage/web3.storage/pull/110#issue-948841130)

flea89: Thanks for the context! I thought you had a good reason for that :)

flea89:
@vasco-santos I've followed up on the comments in the review
- created a tool package
- Updated the `test:e2e:debug`
- removed the `docker/cluster/ipfs-cluster-ctl_v0.14.1_darwin-amd64.tar.gz`

plus some overall improv:
- I've created a cli  and got rid of run.sh
- I've exposed the cli as bin so that other packages can use it directly
- added some scripts to run the cluster from the api package
- change the volume setup to use named volumes and fix an issue where volumes wouldn't be removed by running `docker compose ... down -v ..`. It should be working now 🤞

packages/tools/scripts/cli.js
```diff
@@ -0,0 +1,76 @@
...
+ * @param {boolean} [opts.stop]
+ * @param {boolean} [opts.clean]
+ */
+ export async function clusterCmd ({ project, start, stop, clean }) {
+    const composePath = path.join(__dirname, '../docker/cluster/docker-compose.yml')
+
+    if (!project) {
+        throw new Error('A project must be provided as parameter')
+    }
+
```
vasco-santos, line 38: It would be great do add a safeguard to throw if port is already being used. See https://github.com/web3-storage/web3.storage/blob/main/packages/db/scripts/cmds/db.js#L91

vasco-santos (changes requested):
Other than below looks good to me.

Can you also rebase to solve the conflicts?

packages/tools/scripts/cli.js
flea89: Added a check 👍

flea89:
@vasco-santos Updated the PR with the change you requested and rebased it 🎉

vasco-santos (approved):
LGTM

PR #651, opened 2021-11-18
Feat/pin request table

flea89:


flea89:
If implementations is good, we should consider to: 
- rename tables
- tackle todos
- consider if we need pin requests raws for uploads as well.

PR #677, opened 2021-11-23
fix: pin normalize should filter out unexpected pin status

vasco-santos:
Also added pin status with error to the fixture to guarantee it does not appear on the response anyway and avoid new regressions


alanshaw (approved):
LGTM

I just realised we also need deal statuses filtered: https://github.com/web3-storage/web3.storage/blob/cd04716844a646d986f53aeae7c73840a008aad3/packages/api/src/status.js#L32

PR #679, opened 2021-11-23
fix: script startup timeout

alanshaw:
Import just the bits we need because this s3 client library is a monster and imports about 200 modules. Fixes the following error (for now):

```
Script startup timed out.
 [API code: 10021]
```

refs #79


vasco-santos (approved):
LGTM ❤️

PR #682, opened 2021-11-23
fix: deal normalize should filter out unexpected deal status

vasco-santos:
Per https://github.com/web3-storage/web3.storage/pull/677#pullrequestreview-813653126 filters out deals with other status.

Also normalizes activation and termination per:
- https://github.com/web3-storage/web3.storage/blob/cd04716844a646d986f53aeae7c73840a008aad3/packages/api/src/status.js#L40


alanshaw (approved):
LGTM, I think there's a mock that needs to be updated though...

vasco-santos:
yeah, did not realize that. ready

PR #702, opened 2021-11-25
Add endpoint and logic to our api service

flea89:


packages/api/src/pins.js
```diff
@@ -84,24 +143,43 @@ export async function pinPost (request, env, ctx) {
  * @param {import('./index').Ctx} ctx
  */
 export async function pinGet (request, env, ctx) {
-  const requestId = request.params.requestId
-
-  if (!requestId) {
+  let requestId
+  try {
+    requestId = parseInt(request.params.requestId, 10)
```
alexandrastoica, line 148: From the spec it looks like this can be a string (i.e. "requestid": "UniqueIdOfPinRequest"), are we sure we can always coerce this into an int?

flea89: Good catch 👀 , however we're using the PinRequest primary key as a unique id which, and that is set to be a `bigserial`. So in our system, as it stands right now, the id has to be a coercible int.
Am I missing something?

alexandrastoica: 100% agreed on the db side, but if you parseInt a string, it'll return `NaN` and so I guess it'll just fail in the next try/catch? Is that what we want?

flea89: I was indeed missing something! It'd be better to have it just fail here.
Let me update this

alexandrastoica:
LGTM

packages/api/src/pins.js
flea89: done ✅

packages/api/test/pin.spec.js
```diff
@@ -4,256 +4,409 @@ import { endpoint } from './scripts/constants.js'
...
-    const res = await fetch(new URL('pins', endpoint).toString(), {
-      method: 'POST',
-      headers: {
-        Authorization: `Bearer ${token}`,
-        'Content-Type': 'application/json'
-      },
-      body: JSON.stringify({
-        cid: 'bafybeibqmrg5e5bwhx2ny4kfcjx2mm3ohh2cd4i54wlygquwx7zbgwqs4e',
-        meta: 1
+    it('requires a string as requestId', async () => {
```
alexandrastoica, line 262: nit: I think we can improve this to test whether requestId is a string that can be coerced into number

PR #705, opened 2021-11-26
feat: pinning api allowlist

GaryHomewood:
WIP addition of column to user table to implement an allowlist.


GaryHomewood:
Please note:
New column is currently added directly to the user table script, instead of via a migration script.

packages/api/src/pins.js
```diff
@@ -350,6 +350,13 @@ export async function pinDelete (request, env, ctx) {
     )
   }
 
+  if (!(/^\d+$/.test(request.params.requestId))) {
```
alexandrastoica, line 353: Can we remove this from this PR? It'll probably create conflicts with the DELETE one

packages/db/db-client-types.ts
```diff
@@ -145,7 +148,7 @@ export type ContentItem = {
   created?: definitions['upload']['inserted_at']
   pins: Array<{
     status: definitions['pin']['status']
-    updated: definitions['pin']['updated_at']
+    // updated: definitions['pin']['updated_at']
```
flea89, line 151: @GaryHomewood quick question.
Would you mind expanding around what's the rationale here?
If it's something that is not needed shall we go and remove it altogether?

packages/db/db-client-types.ts
```diff
@@ -155,7 +158,7 @@ export type ContentItemOutput = {
   cid: definitions['content']['cid']
   dagSize?: definitions['content']['dag_size']
   pins: Array<PinItemOutput>,
-  deals: Array<Deal>
+  deals?: Array<Deal>
```
flea89, line 161: Can you please add some rationale behind this change? Just struggling to follow the whole chain, it'd really help if you could add some context.

packages/db/db-client-types.ts
```diff
@@ -192,15 +195,15 @@ export type UploadItem = {
...
+  _id?: string
   type: definitions['upload']['type']
   name?: definitions['upload']['name']
   created: definitions['upload']['inserted_at']
   updated: definitions['upload']['updated_at']
   cid: definitions['content']['cid']
   dagSize?: definitions['content']['dag_size']
   pins: Array<PinItemOutput>,
-  deals: Array<Deal>
+  deals?: Array<Deal>
```
flea89, line 206: Same as above 🙏

packages/db/index.js
```diff
@@ -635,10 +637,15 @@ export class DBClient {
...
+        pinningEnabled:pinning_enabled,
+        keys:auth_key_user_id_fkey(
+          _id:id::text,
+          name,
+          secret
+        )
       `)
       .match({
-        issuer
+        issuer: issuer
```
flea89, line 648: This shouldn't be needed isn't it? 
If only the value is provided it is implicitly assigned to a property with the same key name.

flea89 (changes requested):
Great work @GaryHomewood !
I've left some small comments here and there

packages/api/src/pins.js
GaryHomewood: Ok have removed.

packages/api/package.json
```diff
@@ -14,7 +14,7 @@
     "publish": "wrangler publish --env $(whoami)",
     "build": "WEBPACK_CLI_FORCE_LOAD_ESM_CONFIG=true webpack",
     "test": "npm-run-all -p -r mock:cluster mock:pgrest mock:backup test:e2e -s test:size",
-    "test:e2e": "playwright-test \"test/**/*.spec.js\" --sw src/index.js -b webkit",
```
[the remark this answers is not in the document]
GaryHomewood: Ok have removed

packages/db/db-client-types.ts
GaryHomewood: Ok just removed from the `ContentInput` type since it shouldn't need to be specified on create.

packages/db/db-client-types.ts
GaryHomewood: Ok reinstated, since it is a required attribute for  `getStatus`

packages/db/db-client-types.ts
GaryHomewood: Ok reinstated as above, deals required by `listUploads` and `getUpload`

packages/db/index.js
GaryHomewood: No it isn't needed. Have reverted.

flea89:
Hey @vasco-santos in case you missed the request, would you mind having a look this PR 😇 ? Thanks!

packages/api/src/auth.js
```diff
@@ -68,11 +69,40 @@ export function withApiOrMagicToken (handler) {
...
+    const apiToken = await tryWeb3ApiToken(token, env)
+    if (apiToken) {
+      if (apiToken.user.pinningEnabled) {
+        request.auth = { authToken: apiToken, user: apiToken.user }
+        env.sentry && env.sentry.setUser(apiToken.user)
+        return handler(request, env, ctx)
+      }
+      throw new PinningNotEnabledError()
+    }
+    throw new UnrecognisedTokenError()
```
alanshaw, line 97: I'd just have this middleware take an authenticated request.

```suggestion
    const { user } = request.auth
    if (!user.pinningEnabled) {
      throw new PinningNotEnabledError()
    }
    return handler(request, env, ctx)
```

Although I think it would be preferable to store this information in a separate table that lists user IDs with authorisations. We're having an admin area built and it would be safer if it didn't ever write to the user table.

So instead of checking for `user.pinningEnabled` you'd pass the user ID to some function that would check the user's ID is present in the pinning service authorisations table.

packages/api/src/index.js
```diff
@@ -26,7 +26,8 @@ router.all('*', envAll)
 const auth = {
   '🤲': handler => withCorsHeaders(handler),
   '🔒': handler => withCorsHeaders(withApiOrMagicToken(handler)),
-  '👮': handler => withCorsHeaders(withMagicToken(handler))
+  '👮': handler => withCorsHeaders(withMagicToken(handler)),
+  '📌': handler => withCorsHeaders(withPinningEnabledUser(handler))
```
alanshaw, line 30: So this would become:

```suggestion
  '📌': handler => auth['🔒'](withPinningEnabledUser(handler))
```

packages/db/db-client-types.ts
```diff
@@ -192,7 +194,7 @@ export type UploadItem = {
 }
 
 export type UploadItemOutput = {
-  _id: string
+  _id?: string
```
alanshaw, line 197: Hmm, why is this necessary?

packages/db/index.js
```diff
@@ -635,7 +637,12 @@ export class DBClient {
       .select(`
         _id:id::text,
         issuer,
-        keys:auth_key_user_id_fkey(_id:id::text, name,secret)
+        pinningEnabled:pinning_enabled,
+        keys:auth_key_user_id_fkey(
+          _id:id::text,
+          name,
+          secret
+        )
```
alanshaw, line 645: We won't need this if we use a separate table.

alanshaw (changes requested):

packages/api/test/mocks/pgrest/get_user.js
```diff
@@ -0,0 +1,39 @@
+
```
vasco-santos, line 1: Please rebase, given I already got rid of pgrest mocks in API

vasco-santos (changes requested):
Thanks @GaryHomewood other than Alan's comments, I think you only need to get this using the main branch test setup without postgrest mocks

packages/api/src/auth.js
GaryHomewood: Have now implemented. Using a separate table `pinning_authorisation` and a DB method to check for the existence of a row for the authenticated user.

packages/api/src/index.js
GaryHomewood: Have implemented (and used the word `authorised` everywhere instead of `enabled`)

packages/db/db-client-types.ts
GaryHomewood: It isn't! Have  corrected.

GaryHomewood:
Corrected to have user id instead of auth token id on the permissions table, as foreign key to the user table.

packages/db/test/user.spec.js
```diff
@@ -66,6 +66,7 @@ describe('user operations', () => {
     const name = 'test-name'
     const email = 'new-test@email.com'
     const publicAddress = `public_address${Math.random()}`
+    const pinningEnabled = true
```
flea89, line 69: This isn't used. 
Can we get rid of it?

packages/api/test/pin.spec.js
```diff
@@ -568,27 +723,30 @@ describe('Pinning APIs endpoints', () => {
...
         createPinWithStatus('UnpinQueued'),
         createPinWithStatus('PinError')
       ]
-
       assert.strictEqual(getEffectivePinStatus(pins), 'failed')
     })
   })
 
-  describe('DELETE /pins/:requestId', () => {
+  describe.skip('DELETE /pins/:requestId', () => {
```
flea89, line 744: This shouldn't be skipped, isn't it?
Also this is somehow duplicated, can you please double check (see https://github.com/web3-storage/web3.storage/pull/705/files#diff-14c9b706d6ea2d7b7b4b2742105bb323beb6b3fdb23b862926505fcf482eb37aR651)

packages/api/test/pin.spec.js
```diff
@@ -541,10 +608,98 @@ describe('Pinning APIs endpoints', () => {
...
+          cid: 'bafybeibqmrg5e5bwhx2ny4kfcjx2mm3ohh2cd4i54wlygquwx7zbgwqs4e'
+        })
+      })
+
+      assert(!res.ok)
+      assert.strictEqual(res.status, (new PinningNotAuthorisedError()).status)
+    })
+  })
+
+  describe('POST /pins/:requestId', () => {
```
flea89, line 630: I think this suite is somehow duplicated can you please double-check? (see [line](https://github.com/web3-storage/web3.storage/pull/705/files#diff-14c9b706d6ea2d7b7b4b2742105bb323beb6b3fdb23b862926505fcf482eb37aR835) )

flea89 (changes requested):

packages/db/test/user.spec.js
GaryHomewood: Yes, removed.

packages/api/test/pin.spec.js
GaryHomewood: Have reinstated this and removed duplicates.

packages/api/test/pin.spec.js
```diff
@@ -541,10 +608,98 @@ describe('Pinning APIs endpoints', () => {
...
+      })
+      assert(res, 'Server responded')
+      assert(!res.ok, 'Server returns an error')
+      const data = await res.json()
+      const error = data.error
+      assert.strictEqual(error.reason, ERROR_STATUS)
+      assert.strictEqual(error.details, INVALID_REQUEST_ID)
+    })
+
+    it.skip('requires a requestId', async () => {
```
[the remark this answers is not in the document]
GaryHomewood: Has been reinstated.

packages/api/test/pin.spec.js
GaryHomewood: Yep, I have refactored and removed duplicates.

vasco-santos (changes requested):
Overall looking good! Left several suggestions to keep consistency around the codebase.

Please also rebase the PR to fix the conflicts

packages/api/src/auth.js
```diff
@@ -1,6 +1,7 @@
 import * as JWT from './utils/jwt.js'
 import {
   UserNotFoundError,
+  PinningNotAuthorisedError,
```
alanshaw, line 4: Perhaps more commonly?:

```suggestion
  PinningUnauthorizedError,
```

packages/db/postgres/tables.sql
```diff
@@ -241,3 +241,12 @@ CREATE TABLE IF NOT EXISTS name
     inserted_at TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL,
     updated_at  TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL
 );
+
+CREATE TABLE IF NOT EXISTS pinning_authorisation
```
alanshaw, line 245: For NFT.Storage the plan is to have user "tags" and have the `getUser` call join on this table so a separate request does not have to be made. It also allows multiple features/permissions for a given user.

For example:

| User ID | Tags            |
|---------|-----------------|
| 138     | psa             |
| 567     | psa,storage:2TB |
| 875     | storage:10TB    |

This is fine for now though and can be easily refactored and migrated.

alanshaw (changes requested):
Please can we replace "authorised" with "authori**z**ed" for consistency with other instances in the codebase.

packages/db/postgres/tables.sql
alanshaw: @jsdevel can you verify what I've said above is accurate?

packages/api/src/auth.js
GaryHomewood: Ok cool, spelling changed throughout.

packages/api/src/errors.js
```diff
@@ -20,6 +20,15 @@ export class UserNotFoundError extends HTTPError {
...
 UserNotFoundError.CODE = 'ERROR_USER_NOT_FOUND'
 
+export class PinningUnauthorizedError extends HTTPError {
+  constructor (msg = 'Pinning not authorized for this user') {
+    super(msg, 403)
+    this.name = 'PinningUnauthorizedError'
+    this.code = PinningUnauthorizedError.CODE
+  }
+}
+PinningUnauthorizedError.CODE = 'ERROR_PINNING_NOT_AUTHORIZED'
```
alanshaw, line 30:
```suggestion
PinningUnauthorizedError.CODE = 'ERROR_PINNING_UNAUTHORIZED'
```

packages/api/test/pin.spec.js
```diff
@@ -279,9 +289,27 @@ describe('Pinning APIs endpoints', () => {
...
+      const notAuthorizedToken = await getTestJWT('test-upload', 'test-upload')
+      const res = await fetch(new URL('pins', endpoint).toString(), {
+        method: 'GET',
+        headers: {
+          Authorization: `Bearer ${notAuthorizedToken}`,
+          'Content-Type': 'application/json'
+        }
+      })
+      assert(!res.ok)
+      assert.strictEqual(res.status, (new PinningUnauthorizedError()).status)
```
alanshaw, line 303: Suggestion: parse the response JSON and check the `code` is `PinningUnauthorizedError.CODE`.

...and repeat for all tests checking this :)

packages/db/postgres/tables.sql
```diff
@@ -247,3 +246,12 @@ CREATE TABLE IF NOT EXISTS name
...
     updated_at  TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL
 );
+
+CREATE TABLE IF NOT EXISTS pinning_authorization
+(
+  id              BIGSERIAL PRIMARY KEY,
+  -- Points to user allowed to pin content.
+  user_id     BIGINT                                                       NOT NULL REFERENCES public.user (id),
+  inserted_at     TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL,
+  deleted_at      TIMESTAMP WITH TIME ZONE
```
alanshaw, line 256: Some weird spacing here!

```suggestion
  id          BIGSERIAL PRIMARY KEY,
  -- Points to user allowed to pin content.
  user_id     BIGINT NOT NULL REFERENCES public.user (id),
  inserted_at TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL,
  deleted_at  TIMESTAMP WITH TIME ZONE
```

alanshaw (changes requested):
Just that error code that really needs fixing.

packages/api/src/errors.js
GaryHomewood: Corrected.

packages/api/test/pin.spec.js
GaryHomewood: Yep that's cleaner. Changed here and elsewhere.

packages/db/postgres/tables.sql
GaryHomewood: Fixed.

flea89 (approved):

vasco-santos (approved):

PR #707, opened 2021-11-29
fix: Issue 581 Webpack 4 build error

### LINKED ISSUE #581: Angular 12 compatibilty
Hello, as I don't see nobody reporting this and any thing related to angular 2+ I dont know if its only my problem or there is really a bug, but this error appear when I try to run in angular 12, and looks like some kind of babel bug compilation with async functions so I don't know if it can be compatible with angular 12

node 16.1.0
npm 8.1.1

image

dashcraft:
Fixes #581 
This pr reworks the problematic error message to not use the coalesce operator to resolve webpack4 issues.
In the future might look at the current rollup config and transpile to es5 or another valid syntax and export for older build setups. This however is a quick and easy solution for this specific build issue.

I verified this by creating a webpack 4 config and found the issue and resolved locally. This isn't the most foolproof way, but whenever I fixed it this way, create-react-app using webpack 4 successfully compiled, and angular 13 does not have this issue, it used webpack 5 now.

### DIFF
--- a/packages/client/src/lib.js
+++ b/packages/client/src/lib.js
@@ -266,7 +266,7 @@ class Web3Storage {
       if (!res.ok) {
         /* c8 ignore next 2 */
         const errorMessage = await res.json()
-        throw new Error(`${res.status} ${res.statusText} ${'- ' + errorMessage?.message || ''}`)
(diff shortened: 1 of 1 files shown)


alanshaw (approved):

vasco-santos (approved):
thanks

PR #716, opened 2021-11-29
Issue 703 - Navigation Links the Nextjs way

### LINKED ISSUE #703: Links in main navigation are causing full page reload
Should be being handled by the router.

dashcraft:
Fixes #703  navigation links, there where some other consistency updates:
- Navigation Links are handled by nextjs router/link component
- Styles for navbar made consistent during loading so it's not flashing white
- Added the maintenance banner to loading so the navigation bar doesn't move when loading
- Updated page background color to be consistent with navbar color during loading screen

### DIFF
--- a/packages/website/components/navbar.js
+++ b/packages/website/components/navbar.js
@@ -43,6 +43,7 @@ export default function Navbar({ bgColor = '', isLoggedIn, isLoadingUser }) {
     [
       {
         link: 'https://docs.web3.storage/',
+        slug: 'docs',
         name: 'Docs',
         spacing: 'p-3 md:px-6'
       },
--- a/packages/website/components/navbar.js
+++ b/packages/website/components/navbar.js
@@ -1,5 +1,5 @@
 import { useCallback, useMemo, useRef, useState } from 'react'
-import Router from 'next/router'
+import { useRouter } from 'next/router'
 import Link from 'next/link'
 import { getMagic } from '../lib/magic.js'
 import countly from '../lib/countly'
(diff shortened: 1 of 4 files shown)


the-simian (changes requested):
I checked this out - make sure that if you're on a particular route, the nav item for that route is inert. (eg. if you're on home, clicking home should do nothing)

packages/website/components/navbar.js
```diff
@@ -126,10 +126,12 @@ export default function Navbar({ bgColor = '', isLoggedIn, isLoadingUser }) {
           </a>
         </div>
         <div className={clsx("flex items-center", isSmallVariant ? 'justify-end' : '')} style={{ minHeight: 52 }}>
-          { !isSmallVariant && ITEMS.map(item => (
-            <a href={item.link} key={item.name} onClick={onLinkClick} className={clsx('text-sm text-w3storage-purple font-bold no-underline hover:underline align-middle', item.spacing)}>
-              { item.name }
-            </a>
+          {!isSmallVariant && ITEMS.map(item => (
+            <Link href={item.link} key={item.name} >
+              <a onClick={onLinkClick} className={clsx('text-sm text-w3storage-purple font-bold no-underline hover:underline align-middle', item.spacing)}>
```
redaphid, line 131: Shouldn't the `onClick` be on the `<Link>`  element?

redaphid: nvm

redaphid (approved):
Looks good to me!

the-simian (approved):
Navigation is much tighter now, much improved!

PR #719, opened 2021-11-30
fix: encode created date in Link header

alanshaw:
If this value is not encoded, the "+" is interpreted as a space, and it produces an invalid date string, causing the service to respond with "invalid before date".

```js
new URL('https://api.web3.storage/user/uploads?size=25&before=2021-11-25T22:57:31.887+00:00').searchParams.get('before')
"2021-11-25T22:57:31.887 00:00"

new Date('2021-11-25T22:57:31.887 00:00')
Invalid Date
```

The issue is not seen in tests because we happen to not use `+00:00` (we use `Z` instead) but this is a valid date string currently returned by our service:

Screenshot 2021-11-30 at 13 04 22

### DIFF
--- a/packages/api/src/user.js
+++ b/packages/api/src/user.js
@@ -193,7 +193,7 @@ export async function userUploadsGet (request, env) {
 
   const oldest = uploads[uploads.length - 1]
   const headers = uploads.length === size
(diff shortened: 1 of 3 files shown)


vasco-santos (approved):
Good catch

PR #725, opened 2021-12-01
fix: postgres set max parallel workers per gather to 4

vasco-santos:
We changed the max parallel workers per gather to 4 in production when we moved to Postgres. Let's keep track of it in the infra setup scripts

### DIFF
--- a/packages/db/postgres/config.sql
+++ b/packages/db/postgres/config.sql
@@ -1,0 +1,1 @@
+SET max_parallel_workers_per_gather TO 4;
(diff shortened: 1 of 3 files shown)


alanshaw (approved):
LGTM, if you have the time to send the same PR to NFT.Storage that would be amazing 🙏.

vasco-santos:
> LGTM, if you have the time to send the same PR to NFT.Storage that would be amazing 🙏.

I will!

PR #731, opened 2021-12-01
fix: mispelled available

dashcraft:
fix: mispelled available


JeffLowe (approved):
Confirmed respelling.

PR #735, opened 2021-12-02
fix: remove dagcargo materialized views

### LINKED ISSUE #590: [Testing] DB FDW with local DAG Cargo instance
We need to use DAG Cargo for DB testing, so we should just have a Docker setup for it instead of relying on a deployed one.

vasco-santos:
Removes materialized views in favour of using cargo schema

Also removes the need of using dag cargo credentials for DB testing.

Closes #590


packages/db/postgres/fdw.sql
```diff
@@ -20,3 +20,11 @@ USER MAPPING FOR current_user
       user :'DAG_CARGO_USER', 
       password :'DAG_CARGO_PASSWORD'
     );
+
+CREATE
+USER MAPPING FOR web3_user
```
alanshaw, line 25:
```suggestion
USER MAPPING FOR web3_storage
```

alanshaw (changes requested):
Need the grants also.

packages/db/scripts/cmds/db-sql.js
```diff
@@ -15,9 +15,8 @@ const { Client } = pg
  * @param {Object} [opts]
  * @param {boolean} [opts.reset]
  * @param {boolean} [opts.cargo]
- * @param {boolean} [opts.testing]
  */
-export async function dbSqlCmd ({ reset, cargo, testing } = {}) {
```
alanshaw, line 20: We should keep the ability to setup non-testing version for running locally.

alanshaw (changes requested):

packages/db/scripts/cmds/db-sql.js
vasco-santos: so, use the flag testing instead of cargo.

packages/db/scripts/cmds/db-sql.js
```diff
@@ -64,29 +45,59 @@ export async function dbSqlCmd ({ reset, cargo, testing } = {}) {
...
+    // Replace secrets in the FDW sql file
+    fdwSql = fdwSql.replace(":'DAG_CARGO_HOST'", `'${process.env.DAG_CARGO_HOST}'`)
+    fdwSql = fdwSql.replace(":'DAG_CARGO_DATABASE'", `'${process.env.DAG_CARGO_DATABASE}'`)
+    fdwSql = fdwSql.replaceAll(":'DAG_CARGO_USER'", `'${process.env.DAG_CARGO_USER}'`)
+    fdwSql = fdwSql.replaceAll(":'DAG_CARGO_PASSWORD'", `'${process.env.DAG_CARGO_PASSWORD}'`)
 
+    // Create web3_storage role
+    await client.query(`
+CREATE ROLE web3_storage LOGIN;
+    `)
```
alanshaw, line 60: Use the `WEB3_STORAGE_USER` env var or default to `current_user`. 

```suggestion
    fdwSql = fdwSql.replaceAll(":'DAG_CARGO_PASSWORD'", `'${process.env.DAG_CARGO_PASSWORD}'`)
    fdwSql = fdwSql.replaceAll(":WEB3_STORAGE_USER", `${process.env.WEB3_STORAGE_USER || 'CURRENT_USER'}`)
```

alanshaw (changes requested):

packages/db/scripts/cmds/db-sql.js
```diff
@@ -64,29 +45,56 @@ export async function dbSqlCmd ({ reset, cargo, testing } = {}) {
...
+  // if testing or cargo fdw flag not set, you just get the schema, no fdw connection to dagcargo
+  if (testing || !cargo) {
+    await client.query(cargoSchema)
+  } else {
+    // Replace secrets in the FDW sql file
+    fdwSql = fdwSql.replace(":'DAG_CARGO_HOST'", `'${process.env.DAG_CARGO_HOST}'`)
+    fdwSql = fdwSql.replace(":'DAG_CARGO_DATABASE'", `'${process.env.DAG_CARGO_DATABASE}'`)
+    fdwSql = fdwSql.replaceAll(":'DAG_CARGO_USER'", `'${process.env.DAG_CARGO_USER}'`)
+    fdwSql = fdwSql.replaceAll(":'DAG_CARGO_PASSWORD'", `'${process.env.DAG_CARGO_PASSWORD}'`)
+    fdwSql = fdwSql.replaceAll(":'WEB3_STORAGE_USER'", `${process.env.WEB3_STORAGE_USER || 'CURRENT_USER'}`)
```
alanshaw, line 57:
```suggestion
    fdwSql = fdwSql.replaceAll(":WEB3_STORAGE_USER", `${process.env.WEB3_STORAGE_USER || 'CURRENT_USER'}`)
```

I don't know exactly how this works with normal SQL substitution, but I didn't include the quotes in the SQL because it shouldn't be a string...

alanshaw (approved):
LGTM, just one minor fix needed.

packages/db/scripts/cmds/db-sql.js
vasco-santos: well, the others lines also use `'` and seem to be working. I added this because otherwise we had lint problems.

vasco-santos: so I guess we should change 

```
USER MAPPING FOR :WEB3_STORAGE_USER
```

to be 

```
USER MAPPING FOR ':WEB3_STORAGE_USER'
```

packages/db/scripts/cmds/db-sql.js
```diff
@@ -64,29 +45,56 @@ export async function dbSqlCmd ({ reset, cargo, testing } = {}) {
...
-FROM cargo.aggregate_entries 
-WHERE cid_v1 in ('bafybeiaj5yqocsg5cxsuhtvclnh4ulmrgsmnfbhbrfxrc3u2kkh35mts4e');
-`
-      )
-    }
-
+  // if testing or cargo fdw flag not set, you just get the schema, no fdw connection to dagcargo
+  if (testing || !cargo) {
+    await client.query(cargoSchema)
+  } else {
```
alanshaw, line 51: On second thoughts I think it makes more sense to have cargo stuff only when you pass the cargo flag:

```suggestion
  // if testing or cargo fdw flag not set, you just get the schema, no fdw connection to dagcargo
  if (testing && cargo) {
    await client.query(cargoSchema)
  } else if (!testing && cargo) {
```

packages/db/scripts/cmds/db-sql.js
alanshaw: The vars are like this because of psql variable interpolation:
https://www.postgresql.org/docs/9.2/app-psql.html#APP-PSQL-INTERPOLATION

In the SQL we have `:WEB3_STORAGE_USER` and it'll be replaced with `CURRENT_USER` (_not_ a string) (or whatever the env var is set to).

For the other variables, e.g. `:'DAG_CARGO_USER'` it is quoted because we want the substitution to be a string like: `'cargo'`.

I've changed the suggestion so no linting error.

alanshaw (changes requested):

alanshaw (approved):

packages/db/scripts/cmds/db-sql.js
```diff
@@ -64,29 +45,56 @@ export async function dbSqlCmd ({ reset, cargo, testing } = {}) {
...
+  aggregate_cid TEXT NOT NULL,
+  cid_v1 TEXT NOT NULL,
+  datamodel_selector TEXT NOT NULL
+);
+
+CREATE TABLE IF NOT EXISTS cargo.aggregates (
+  aggregate_cid TEXT NOT NULL UNIQUE,
+  piece_cid TEXT UNIQUE NOT NULL,
+  sha256hex TEXT NOT NULL,
+  payload_size BIGINT NOT NULL,
```
alanshaw, line 79:
```suggestion
  export_size BIGINT NOT NULL,
```

https://filecoinproject.slack.com/archives/C02BZPRS9HP/p1638980517480000

alanshaw (changes requested):

packages/db/scripts/cmds/db-sql.js
```diff
@@ -64,29 +45,56 @@ export async function dbSqlCmd ({ reset, cargo, testing } = {}) {
...
     await client.query(fdwSql)
     await client.query(cargoSql)
   }
 
   await client.query(functionsSql)
   await client.end()
 }
+
+const cargoSchema = `
+CREATE SCHEMA IF NOT EXISTS cargo;
```
alanshaw, line 67: Note: I put this in a file in https://github.com/nftstorage/nft.storage/pull/922

alanshaw (approved):

PR #736, opened 2021-12-02
feat/pinning apis delete and replace

alexandrastoica:


packages/db/index.js
```diff
@@ -880,6 +882,34 @@ export class DBClient {
...
+      .match({ auth_key_id: authKey, id: requestId })
+      .filter('deleted_at', 'is', null)
+      .single()
+
+    if (error) {
+      throw new DBError(error)
+    }
+
+    return {
+      _id: data.id
```
alexandrastoica, line 909: @flea89 As per the pinning spec we don't need to return anything, but I guess it's good practice to return id on deletion. What do you think?

PR #751, opened 2021-12-06
fix! : client should not throw on 404

### LINKED ISSUE #687: DB client throws on 404, change to return undefined
Our calling code expects it to return `undefined` for missing row not throw. We give cryptic database error messages back to the user, not a nice `UserNotFoundError` (for example):

143015360-bb3a2187-5512-4b6a-86a1-19dea7d13365

See conevrsation here: https://github.com/web3-storage/web3.storage/pull/648#discussion_r753114449

vasco-santos:
Closes #687

### DIFF
--- a/packages/db/index.js
+++ b/packages/db/index.js
@@ -70,10 +70,10 @@ export class DBClient {
    * Get user by its issuer.
    *
    * @param {string} issuer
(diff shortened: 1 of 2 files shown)


alanshaw (approved):
❤️

PR #757, opened 2021-12-06
chore: created a formatter abstraction for filesize to standardize ie…

dashcraft:
chore: created a formatter abstraction for filesize to standardize iec binary units for all file size display props

I wrapped existing filesize package into a small util formatted for website.

Fixes #757


dashcraft:
image

redaphid (approved):
Nothing too surprising here :)

Issue #717, opened 2021-11-30
API package testing - Spin out docker containers for db and IPFS cluster


flea89:

vasco-santos:
DB part is done in https://github.com/web3-storage/web3.storage/pull/765

Will do the cluster once we get https://github.com/web3-storage/web3.storage/pull/650 merged in

vasco-santos:
So, I spent some time trying to get the cluster to work, but I am getting CORS headers problem.

I think we need to setup in go-ipfs https://discuss.ipfs.io/t/how-to-enable-cors-in-go-ipfs-dockerfile/8543 , which is not possible via ENV vars. We need to setup a custom docker image with a configuration file to make it possible.

PR #765, opened 2021-12-09
fix: api using db instead of mocks for api tests

vasco-santos:
Partially resolves https://github.com/web3-storage/web3.storage/issues/717 by using Docker to run a DB instead of mocks for api tests

Also removes remaining unused mocks from graphql API used when mocking fauna.


packages/api/test/user.spec.js
```diff
@@ -26,7 +27,7 @@ describe('GET /user/account', () => {
     })
     assert(res.ok)
     const data = await res.json()
-    assert(data.usedStorage)
+    assert.strictEqual(data.usedStorage, 0)
```
vasco-santos, line 30: we previously returned a mock with a value that was not 0

packages/api/test/user.spec.js
```diff
@@ -113,7 +114,7 @@ describe('POST /user/tokens', () => {
 describe('DELETE /user/tokens/:id', () => {
   it('error if not authenticated with magic.link', async () => {
     const token = await getTestJWT()
-    const res = await fetch(new URL('user/tokens/xyz', endpoint), {
+    const res = await fetch(new URL('user/tokens/2', endpoint), {
```
vasco-santos, line 117: older value failed because param was not a bigint

packages/api/test/user.spec.js
```diff
@@ -304,7 +215,7 @@ describe('DELETE /user/uploads/:cid', () => {
 
   it('removes an upload', async () => {
     const token = 'test-magic'
-    const res = await fetch(new URL('user/uploads/bafybeibq5kfbnbvjgjg6bop4anhhaqopkc7t6mp2v3er3fkcv6ezhgvavg', endpoint), {
+    const res = await fetch(new URL('user/uploads/bafkreiajkbmpugz75eg2tmocmp3e33sg5kuyq2amzngslahgn6ltmqxxfa', endpoint), {
```
vasco-santos, line 218: Needs to be inserted before

alanshaw (approved):
🚀

PR #778, opened 2021-12-12
fix: remove resize observer, use css for responsive menu

### LINKED ISSUE #771: In older mobile devices the useResizeObserver approach for mobile detection breaks the layout
In older mobile devices (eg. Iphone 6) the useResizeObserver approach for mobile detection breaks the layout

The way the navbar.js handles mobile detection can cause issues in older and older devices. https://caniuse.com/?search=ResizeObserver, shows that this will likely never work in any IE browsers and will probably fail in older versions of safari and all opera.

We should prefer a CSS based solution that is more stable

This issue was discivered in nft.storage https://github.com/nftstorage/nft.storage/issues/915 and identified as existing in web3 https://github.com/web3-storage/web3.storage/blob/main/packages/website/components/navbar.js#L30

The solution implemented in nft should likely be replicated here.

dashcraft:
Fixes #771 

Changes js responsive oberver to be CSS based, tested for accessibility using axe devtools, introduced no new a11y issues using axe for cursory review.

No new global styles introduced, used tailwind to mimick behavior.

Tested on:
- [x] Safari 
- [x] Chrome 
- [x] Firefox


dashcraft:
Mobile safari:

image

dashcraft:
I'm going to resize the button before asking for review.

packages/website/components/navbar.js
```diff
@@ -1,17 +1,16 @@
...
+import countly from "../lib/countly";
+import { useQueryClient } from "react-query";
+import Button from "./button.js";
+import clsx from "clsx";
+import Hamburger from "../icons/hamburger";
 
-import Logo from '../icons/w3storage-logo'
-import Cross from '../icons/cross'
-import Loading from './loading.js'
+import Logo from "../icons/w3storage-logo";
```
Codigo-Fuentes, line 11: looks like you have prettier changing single quotes to double quotes and adding semicolons. Let's not do that in this pr if possible 🙏

Codigo-Fuentes (changes requested):

packages/website/components/button.js
```diff
@@ -82,14 +81,15 @@ export default function Button({
         'items-center',
         'justify-center',
         'px-4',
+        'md:min-h-3.25',
```
Codigo-Fuentes, line 84: it was not a result of your changes but i noticed the login/out button while loading on mobile doesn't have a min-height / width and looks weird; maybe you could fix while you're at it? Recommendation: min-w: 2.5rem, min-h: 2rem;

IMG_5476

packages/website/components/navbar.js
dashcraft: This issue should be resolved now!

packages/website/components/button.js
dashcraft: Great catch! Resolved!

redaphid (approved):
:+1:

Issue #793, opened 2021-12-15
Create pin request - Get dag size


flea89:

alanshaw:
This will have to come from dagcargo after we receive the data.

PR #810, opened 2021-12-16
Refactor Pin APIs and PR feedback

flea89:
In #726 we've received really good feedback and, to avoid pushing the updates directly to that feature branch, we've been working on this [PR](https://github.com/web3-storage/web3.storage/pull/810).

The scope of this PR is mostly:
- Refactor pin_request to create content and related pins on PinRequest creation.
- Tackle all the related feedback and quick wins from #726. (we've updated the comments there to point this out)

Please be aware that not every single point of feedback has been tackled just yet. 

This PR is already too big as it is and, for that reason, we decided to break up the remaining work into smaller tickets to have smaller and more manageable units of work.

 You can see them [here](https://app.zenhub.com/workspaces/web3storage-product-6106d25b58fccc00105aedcb/issues/web3-storage/web3.storage/289) 
We also made sure, where possible, to tie the TODOs in the code with the related tickets.


packages/api/src/utils/pin.js
```diff
@@ -97,9 +109,9 @@ export function toPins (peerMap) {
  * @param {number} checkInterval
  * @return {Promise.<import('@web3-storage/db/db-client-types').PinUpsertInput[]>}
  */
-export async function waitToGetOkPins (cid, cluster, waitTime = MAX_PIN_STATUS_CHECK_TIME, checkInterval = PIN_STATUS_CHECK_INTERVAL) {
+export async function waitOkPins (cid, cluster, waitTime = MAX_PIN_STATUS_CHECK_TIME, checkInterval = PIN_STATUS_CHECK_INTERVAL) {
```
GaryHomewood, line 112: Like the default values!

Maybe this could do with some documentation and comments? And some method renaming for clarity?
Would Promise.all() be any use here?

packages/api/src/pins.js
```diff
@@ -38,7 +38,7 @@ import { waitToGetOkPins } from './utils/pin.js'
  * @param {import('../../db/db-client-types.js').PinItemOutput[]} pins
  * @return {apiPinStatus} status
  */
-export const getPinningAPIStatus = (pins) => {
+export const getEffectivePinStatus = (pins) => {
```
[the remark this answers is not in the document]
flea89: Alan suggested the [change](https://github.com/web3-storage/web3.storage/pull/726#discussion_r764772636), I'd follow his lead here

packages/db/index.js
```diff
@@ -20,17 +20,17 @@ const uploadQuery = `
         content(cid, dagSize:dag_size, pins:pin(status, updated:updated_at, location:pin_location(_id:id, peerId:peer_id, peerName:peer_name, region)))
       `
 
-const PAPinRequestTableName = 'pa_pin_request'
+const psaPinRequestTableName = 'psa_pin_request'
```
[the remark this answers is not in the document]
flea89: `psa stands` for `pinning service api`.

It is the prefix used by nft and pinata. We agreed with Alan to use it as a prefix for pinning service api namespacing across the app.

packages/api/src/utils/pin.js
```diff
@@ -55,25 +55,37 @@ export function toPinStatusEnum (trackerStatus) {
...
- * TODO: refactor car upload to use this instead.
+ * Function that returns list of pins for given CID.
  *
  * @param {string} cid cid to be looked for
  * @param {import('@nftstorage/ipfs-cluster').Cluster} cluster
  * @param {import('@nftstorage/ipfs-cluster').StatusResponse['peerMap']} [peerMap] Optional list of peers, if not provided the fuctions queries the cluster.
  * @return {Promise.<import('@web3-storage/db/db-client-types').PinUpsertInput[]>}
  */
-export async function getOKpins (cid, cluster, peerMap) {
+export async function getPins (cid, cluster, peerMap) {
```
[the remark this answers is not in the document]
flea89: I see your point, however, we're are using consistent naming used by `@nftstorage/ipfs-cluster`.
I'd rather be consistent for clarity and consistency, what do you think?

packages/api/src/utils/pin.js
flea89: Added some documentation, hope it helps. 
Not sure what you have in mind with `Promise.all()`, please feel free to suggest if you have anything specific in mind. Same for a better name for the function.

packages/api/src/utils/pin.js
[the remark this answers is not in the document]
GaryHomewood: Ok cool. Consistency is good.

packages/api/src/utils/pin.js
GaryHomewood: Method comment definitely helps now.
Was just wondering if `Promise.all()` might help or simplify, but probably not in this case since we're just retrying until a time.

packages/db/db-client-types.ts
```diff
@@ -248,33 +248,37 @@ export type ListUploadsOptions = {
...
+export type PsaPinRequestUpsertInput = {
   id?: string,
-  name?: definitions['pa_pin_request']['name'],
+  name?: definitions['psa_pin_request']['name'],
   authKey: string,
-  requestedCid: definitions['pa_pin_request']['requested_cid'],
-  cid: definitions['pa_pin_request']['requested_cid'],
+  sourceCid: definitions['psa_pin_request']['source_cid'],
+  contentCid: definitions['upload']['content_cid'],
+  dagSize?: definitions['content']['dag_size'],
```
vasco-santos, line 257: Will we be able to get this at the PinRequest creation timing? From the API Spec it seems like we won't? https://ipfs.github.io/pinning-services-api-spec/#tag/pins/paths/~1pins/post

packages/api/src/utils/pin.js
```diff
@@ -108,3 +125,22 @@ export async function waitToGetOkPins (cid, cluster, waitTime = MAX_PIN_STATUS_C
...
+ * @param {import('@nftstorage/ipfs-cluster').Cluster} cluster
+ * @param {import('@web3-storage/db').DBClient} db
+ * @param {number} waitTime
+ * @param {number} checkInterval
+ * @return {Promise.<import('@web3-storage/db/db-client-types').PinUpsertInput[]>}
+ */
+export async function waitAndUpdateOkPins (cid, cluster, db, waitTime = MAX_PIN_STATUS_CHECK_TIME, checkInterval = PIN_STATUS_CHECK_INTERVAL) {
+  const okPins = await waitOkPins(cid, cluster, waitTime, checkInterval)
+  for (const pin of okPins) {
+    await db.upsertPin(cid, pin)
```
vasco-santos, line 143: I know this was already like this before. But now that I see it, can we create an issue to have `upsertPins` receiving an array instead? What do you think? We would be able to upsert all of them in one request instead of multiple

packages/db/db-client-types.ts
flea89: I agree, I don't think we will but, honestly, we haven't dug into it just yet.
We have an [issue](https://github.com/web3-storage/web3.storage/issues/793) to look into it more closely.
Can I reserve to give you a more accurate answer when we start working on it?

packages/api/test/mocks/pgrest/patch_pa_pin_request.js
```diff
@@ -12,7 +12,7 @@ module.exports = ({ query }) => {
       {
         _id: 1,
         contentCid: null,
-        requestedCid: 'something',
+        sourceCid: 'something',
```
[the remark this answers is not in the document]
flea89: Good catch!

vasco-santos (approved):
LGTM

packages/api/src/pins.js
```diff
@@ -56,11 +56,8 @@ export const getPinningAPIStatus = (pins) => {
     return 'queued'
   }
 
-  if (pinStatuses.length === 0) {
-    return 'queued'
-    // TODO after some time if there are no pins we should give up and return a failed
-    // status instead
-  }
+  // TODO after some time if there are no pins we should give up and return a failed
+  // status instead
```
alanshaw, line 60: Seems that is what we do now, straight away. Maybe this can be removed?

packages/api/src/pins.js
```diff
@@ -106,7 +103,7 @@ export async function pinPost (request, env, ctx) {
 
   // Validate cid
   try {
-    normalizeCid(cid)
+    pinData.normalizedCid = normalizeCid(cid)
```
alanshaw, line 106: Personally I'd not augment this user provided data with anything but instead pull out the bits you want and combine them any derived/additional data in a different object.

packages/api/src/pins.js
```diff
@@ -165,61 +162,46 @@ export async function pinPost (request, env, ctx) {
...
-  }
+  const pinName = pinData.name || undefined // deal with empty strings
 
-  if (origins) {
-    pinOptions.origins = origins
-  }
+  await env.cluster.pin(cid, {
+    name: pinName,
+    origins,
+    metadata: meta
```
alanshaw, line 172: We can't pass user provided metadata to cluster. Cluster is not multi-tenant so two users pinning the same CID will ovewrite the meta.

packages/api/src/pins.js
```diff
@@ -165,61 +162,46 @@ export async function pinPost (request, env, ctx) {
...
+      waitAndUpdateOkPins.bind(
+        null,
+        normalizeCid,
+        env.cluster,
+        env.db)
+    )
   }
 
-  // TODO: Backups. At the moment backups are related to uploads so
+  // TODO(https://github.com/web3-storage/web3.storage/issues/794)
```
alanshaw, line 202: I think we should remove this todo. This is not where the code for backups of pin requests will live.

packages/db/db-client-types.ts
alanshaw: No we won't have this and can't get it from a worker. It will have to come from dagcargo after Cluster receives the data.

packages/db/index.js
```diff
@@ -763,56 +763,53 @@ export class DBClient {
...
+    /** @type {{ data: string, error: PostgrestError }} */
+    const { data: pinRequestId, error } = await this._client.rpc('create_psa_pin_request', {
+      data: {
+        auth_key_id: pinRequestData.authKey,
+        content_cid: pinRequestData.contentCid,
+        source_cid: pinRequestData.sourceCid,
+        name: pinRequestData.name,
+        dag_size: pinRequestData.dagSize,
+        inserted_at: pinRequestData.created || now,
+        updated_at: pinRequestData.updated || now,
```
alanshaw, line 781: Do we need to generate and send these two fields? The DB has defaults.

packages/api/src/pins.js
flea89: I guess this relates to the fact that it'd be nice to have some kind of background task to check after a given defined threshold (an hour, a day, a week I don't know) to tell the cluster to stop trying to pin the cid (I know you have priority queues now but I still feel it should be good to programmatically stop the cluster trying).
At the same time, we should update the DB and therefore the request status accordingly.

I don't think this is a super high priority but still worth tracking?
If you agree I can update the comment to be more verbose and tie it with a github issue.

packages/api/src/pins.js
flea89: Really good catch, I'll update.

packages/api/src/pins.js
flea89: Ack 👍

packages/db/index.js
flea89: Yeah, I agree, I was "annoyed" by this as well.

IIRC we decided to keep it consistent with the `createUpload` behavior.
We're calling a remote procedure here and we were not quite sure what the behavior would be with defaults (when it comes to PostgreSQL) so we assumed there was some issue with that given your `createUpload` implementation.

packages/api/src/pins.js
flea89: You caught me, I was being lazy 😛. On it.

packages/api/src/pins.js
flea89: ping 😇

PR #822, opened 2021-12-23
fix scroll issue on desktop safari

jchris:
Currently on Safari desktop, the hero image scrolls instead of letting the page scroll. This fixes that. I took a look at desktop Chrome and it didn't break but I'm not sure yet how the automated tests work for multi-platform regressions.

Screen Shot 2021-12-23 at 9 37 36 AM
.


dashcraft (approved):
Looks good to me!

PR #823, opened 2021-12-24
feat: add indexes to backup and set timeouts

vasco-santos:


alanshaw:
I *think* the backup job does not go through heroku so 30s limit does not apply...

Thank you for the indexes though I will use.

vasco-santos:
> I think the backup job does not go through heroku so 30s limit does not apply...

Yes, the timeouts are for metrics. They were also timing out at the time, but since we removed the Pin dag size, they look good. So I will remove the timeouts as we also want to move the metrics to be computed recurrently with a cron job (like nft)

packages/db/postgres/functions.sql
```diff
@@ -222,6 +222,8 @@ CREATE OR REPLACE FUNCTION pin_from_status_total(query_status TEXT) RETURNS TEXT
 AS
 $$
 BEGIN
+  -- Set timeout as imposed by heroku
+  SET LOCAL statement_timeout = '30s';
```
vasco-santos, line 226:
```suggestion
```

packages/db/postgres/functions.sql
```diff
@@ -235,6 +237,8 @@ CREATE OR REPLACE FUNCTION content_dag_size_total() RETURNS TEXT
...
```
vasco-santos, line 241:
```suggestion
```

packages/db/postgres/functions.sql
```diff
@@ -247,6 +251,8 @@ CREATE OR REPLACE FUNCTION pin_dag_size_total() RETURNS TEXT
...
```
vasco-santos, line 255:
```suggestion
```

alanshaw (approved):

alanshaw:
I'm going to switch https://github.com/nftstorage/backup/pull/1 to use `updated_at` which means we don't need to add this additional index.

PR #848, opened 2022-01-07
feat: filter pins by status

GaryHomewood:


LeslieOA (approved):

packages/db/postgres/docker/docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '3.6'
 
 services:
   rest:
-    image: postgrest/postgrest:v8.0.0
```
flea89, line 5: Is this required for the filtering to work?

If that's the case we need to make sure the version we're using on Heroku matches this one.

flea89 (dismissed):
Great to see the progress on this one.
I've added a few comments here and there, let me know what you think.

packages/db/postgres/docker/docker-compose.yml
GaryHomewood: It is, and has already been bumped on main.

packages/db/index.js
```diff
@@ -835,40 +863,58 @@ export class DBClient {
...
-      query = query.like('name', `${opts.name}`)
-    }
-
-    if (opts.name && match === 'iexact') {
-      query = query.ilike('name', `${opts.name}`)
-    }
+    if (!Object.keys(opts).length) {
+      query = query.eq('content.pins.status', PIN_STATUS[PIN_STATUS.Pinned])
+    } else {
+      if (opts.status) {
```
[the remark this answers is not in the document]
GaryHomewood: Agree, the mapping of filter status to DB status should be the responsibility of the API

packages/db/db-client-types.ts
```diff
@@ -294,7 +294,7 @@ export type ListPsaPinRequestOptions = {
   /**
    * status  to match
    */
-  status?: Array<definitions['pin']['status']>
+  statuses?: Array<definitions['pin']['status']>
```
GaryHomewood, line 297: This makes sense, given that the mapping of filter status to DB status is now in the API package. Although the type is now a mix of API filter attributes and DB attributes.

packages/db/index.js
```diff
@@ -836,40 +862,45 @@ export class DBClient {
...
-
-    if (opts.name && match === 'exact') {
-      query = query.like('name', `${opts.name}`)
-    }
-
-    if (opts.name && match === 'iexact') {
-      query = query.ilike('name', `${opts.name}`)
-    }
+    if (!Object.keys(opts).length) {
+      query = query.eq('content.pins.status', 'Pinned')
```
GaryHomewood, line 870: Should `Pinned` be some kind of enum?

flea89: That's a really good point, I thought about it as well.
But given we already have statuses enums (that don't quite do for our case though) I'm afraid adding another one would be a bit confusing. 
So If it's ok with you I'd leave it out this specific ticket.

packages/db/db-client-types.ts
flea89: I don't feel this represents a problem, the other options look reasonable in line.

vasco-santos (approved):
Good work!
lgtm

alanshaw (approved):
LGTM

PR #855, opened 2022-01-11
fix: update deps in client api and w3

vasco-santos:
Was going to update `files-from-path`, but noticed lot's of other deps were not updated for a long time.


packages/api/test/car.spec.js
```diff
@@ -19,7 +19,7 @@ describe('POST /car', () => {
     const { root, car: carBody } = await createCar('hello world!')
 
     // expected CID for the above data
-    const expectedCid = 'bafybeiczsscdsbs7ffqz55asqdf3smv6klcw3gofszvwlyarci47bgf354'
+    const expectedCid = 'bafkreidvbhs33ighmljlvr7zbv2ywwzcmp5adtf4kqvlly67cy56bdtmve'
```
vasco-santos, line 22: this was getting the empty directory CID because default wrapWithDirectory true and no path. Changed it in the createCar function

packages/api/test/mocks/cluster/post_add.js
```diff
@@ -1,35 +1,26 @@
-const { importer } = require('ipfs-unixfs-importer')
-
-const block = {
-  get: async cid => { throw new Error(`unexpected block API get for ${cid}`) },
-  put: async () => { throw new Error('unexpected block API put') }
-}
+const { CarReader } = require('@ipld/car')
```
vasco-santos, line 1: Made it `post_add$format=car.js ` given we always send CARs now

alanshaw (approved):

PR #863, opened 2022-01-13
Get request id by user token

alexandrastoica:
This PR prevents accessing pin requests that don't belong to the user token


packages/api/src/pins.js
```diff
@@ -455,7 +456,7 @@ export async function pinDelete (request, env, ctx) {
 async function replacePin (newPinData, requestId, authToken, env, ctx) {
   let existingPinRequest
   try {
-    existingPinRequest = await env.db.getPsaPinRequest(requestId)
```
flea89, line 458: Not really something you changed here but I noticed it's slightly confusing authtoken is sometimes of type `string` and sometimes of type `AuthToken`. 
It would be nice to be consistent throughout the ticket.
Happy for it to be a separate ticket

flea89 (approved):
LGTM!

packages/api/src/pins.js
alexandrastoica: Yeah I noticed that too. We sometimes pass the token and sometimes the ID, but call it the same. I'll see if I can address that here or open a new ticket.

PR #882, opened 2022-01-19
perf: update pin composite index and adds inserted at index to content

vasco-santos:
This PR updates pin composite index for performance reasons on dag cargo queries, as well as adds an index for `inserted_at` which is used as part of the update dag size job in https://github.com/web3-storage/web3.storage/blob/main/packages/cron/src/jobs/dagcargo.js#L3-L31

### DIFF
--- a/packages/db/postgres/tables.sql
+++ b/packages/db/postgres/tables.sql
@@ -132,8 +132,7 @@ CREATE INDEX IF NOT EXISTS pin_content_cid_idx ON pin (content_cid);
 CREATE INDEX IF NOT EXISTS pin_location_id_idx ON pin (pin_location_id);
 CREATE INDEX IF NOT EXISTS pin_updated_at_idx ON pin (updated_at);
 CREATE INDEX IF NOT EXISTS pin_status_idx ON pin (status);
-CREATE INDEX IF NOT EXISTS pin_composite_pinned_at_idx ON pin (content_cid, updated_at) WHERE status = 'Pinned';
-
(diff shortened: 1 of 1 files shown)


packages/db/postgres/tables.sql
```diff
@@ -160,7 +159,7 @@ CREATE TABLE IF NOT EXISTS upload
 
 CREATE INDEX IF NOT EXISTS upload_auth_key_id_idx ON upload (auth_key_id);
 CREATE INDEX IF NOT EXISTS upload_content_cid_idx ON upload (content_cid);
-CREATE INDEX IF NOT EXISTS upload_updated_at_idx ON upload (updated_at);
+CREATE INDEX IF NOT EXISTS upload_composite_updated_at_and_content_cid_idx ON upload (updated_at, content_cid);
```
ribasushi, line 162: This doesn't help, kill it with 🔥

packages/db/postgres/tables.sql
```diff
@@ -132,8 +132,7 @@ CREATE INDEX IF NOT EXISTS pin_content_cid_idx ON pin (content_cid);
 CREATE INDEX IF NOT EXISTS pin_location_id_idx ON pin (pin_location_id);
```
ribasushi, line 132: Don't forget to kill the (contained in `UNIQUE` )
```
CREATE INDEX IF NOT EXISTS pin_content_cid_idx ON pin (content_cid);
```

ribasushi (approved):

packages/db/postgres/tables.sql
vasco-santos: yeah, just commited that change

PR #884, opened 2022-01-20
`pin_request` ID to `uuid`

LeslieOA:
- feat: set search_path to public for uuid function, use uuid as pk on pin_request


packages/db/postgres/tables.sql
```diff
@@ -176,10 +176,16 @@ CREATE TABLE IF NOT EXISTS backup
...
+-- Setting search_path to public scope for uuid function(s)
+SET search_path TO public;
+DROP extension IF EXISTS "uuid-ossp";
+CREATE extension "uuid-ossp" SCHEMA public;
+
 -- Tracks requests to replicate content to more nodes.
 CREATE TABLE IF NOT EXISTS pin_request
 (
-  id              BIGSERIAL PRIMARY KEY,
+  id              UUID DEFAULT public.uuid_generate_v4() PRIMARY KEY,
```
flea89, line 188: Nice!
I'm wondering though if we should have a separate field altogether for this to avoid exposing the PK?

packages/db/test/pinning.spec.js
```diff
@@ -390,24 +398,33 @@ describe('Pin Request', () => {
...
       assert.ok(deletedPinRequest)
       assert.equal(deletedPinRequest._id, pinRequest._id)
     })
 
     it('does not select pin request after deletion', async () => {
+<<<<<<< HEAD
+      assert.rejects(client.getPsaPinRequest(aPinRequestOutput._id))
+=======
       assert.rejects(client.getPsaPinRequest(authKey, parseInt(aPinRequestOutput._id, 10)))
+>>>>>>> feat/pinning-apis
```
GaryHomewood, line 423: Some merge conflicts here. Looks like it should be authKey plus id not parsed to int?

packages/db/test/pinning.spec.js
```diff
@@ -403,11 +403,11 @@ describe('Pin Request', () => {
     })
 
     it('does not select pin request after deletion', async () => {
-      assert.rejects(client.getPsaPinRequest(authKey, parseInt(aPinRequestOutput._id, 10)))
```
alexandrastoica, line 406: I think we need this `parseInt`? I remember adding it because of type errors.

flea89: This is now a string and not a bigint serial. Hence this has changed, I've added other comments to make sure we update this across the codebase.

flea89 (commented):
Great to see this taking shape!
I've added some comments here and there in the PR.
On top of that, if you look at [packages/api/pin.js](https://github.com/web3-storage/web3.storage/blob/e7cc298d239df1087643f209d44d04226b67cf97/packages/api/src/pins.js#L220) before this we were assuming the `pinRequest id` to be a number while it is not a string. We need to update that to reflect the changes.

packages/db/test/pinning.spec.js
alexandrastoica: yeah that occurred to me after, ignore me :D

alanshaw (approved):

PR #885, opened 2022-01-20
fix: db client do not throw error when no upload found

vasco-santos:
When the delete route is called with an upload that does not exist (it can already be deleted), the DB client should not throw and be the API who creates the error. DB should return null as it couldn't delete any upload and return an ID

### DIFF
--- a/packages/db/index.js
+++ b/packages/db/index.js
@@ -278,7 +278,7 @@ export class DBClient {
       .single()
 
     if (error) {
-      throw new DBError(error)
+      return undefined
     }
 
     return {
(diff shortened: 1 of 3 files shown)


packages/db/index.js
```diff
@@ -278,7 +278,7 @@ export class DBClient {
       .single()
 
     if (error) {
-      throw new DBError(error)
+      return undefined
```
hugomrdias, line 281: i think here it should catch the 206 status code or the `!data` not all others errors

hugomrdias (changes requested):

hugomrdias (approved):

PR #890, opened 2022-01-22
fix: next and ts linting cleanup

orvn:
_(This PR relates to the new site in development in `feat/new-brand-frontend`)_

Adding `next` and `next/core-web-vitals` to the `.eslintrc.js` should identify all the issues that will make the site compatible with `next build` and `next export`. 

Some of those issues are resolved in this PR, as are a lot of the errors that were being thrown from `tsc`. 

`eslint` however (`eslint '{pages,components,hooks,content,store}/**/*.{js,ts,tsx}'` still throws a number of errors that I wanted you to take a look at (3 errors right now that cause linting to exit). These are the same errors that appear when running `npx next build`, so this is probably what's preventing the static site output as well.

### DIFF
--- a/packages/website/assets/illustrations/spring.js
+++ b/packages/website/assets/illustrations/spring.js
@@ -1,3 +1,4 @@
+// @ts-ignore
 import spring from '../images/spring.png'
 /**
  * @param {any} props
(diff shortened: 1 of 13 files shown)


orvn (commented):
All issues that exited to stderr resolved for the time being. Another PR may be opened in the future to resolve minor warnings.

PR #891, opened 2022-01-22
fix: Adding static props for exporting index file

MichaelPhan:
### Changes
- Adding static props to all marketing pages


orvn (approved):
Build successful and index generates in the build preview 🎉

PR #896, opened 2022-01-25
docs: add note about reserved Remote upload type

alanshaw:
In NFT.Storage pin requests are added to the upload table with type "Remote". In dagcargo, PSA pin requests from Web3.Storage are combined with uploads (as they are in NFT.Storage) and are artificially given a type "Remote". Hence specifically adding a "Remote" type here would be problematic.


vasco-santos (approved):

ribasushi (approved):
😻

PR #897, opened 2022-01-25
feat: add origins to psa pin request

alexandrastoica:


packages/db/postgres/pg-rest-api-types.d.ts
```diff
@@ -1825,7 +2000,8 @@ export interface parameters {
   "rowFilter.psa_pin_request.content_cid": string;
   "rowFilter.psa_pin_request.source_cid": string;
   "rowFilter.psa_pin_request.name": string;
-  "rowFilter.psa_pin_request.meta": Record<string, string>;
+  "rowFilter.psa_pin_request.origins": string;
+  "rowFilter.psa_pin_request.meta": string;
```
alexandrastoica, line 2004: Looks like it assigns 'string' as type for origins and meta. They are obviously not, but since this is generated I wonder how we can fix it (without manually changing it)?

alanshaw: Is this a regression in the tool? Previously `meta`  was `Record<string, string>`, which is also a `jsonb` field.

packages/db/test/pinning.spec.js
```diff
@@ -114,9 +116,10 @@ describe('Pin Request', () => {
       assert.strictEqual(savedPinRequest._id, aPinRequestOutput._id)
     })
 
-    it('returns the right object', async () => {
+    it.only('returns the right object', async () => {
```
alanshaw:
```suggestion
    it('returns the right object', async () => {
```

I would also welcome a better test name here...

packages/db/postgres/pg-rest-api-types.d.ts
alexandrastoica: I edited that manually (and happy to do it here) but running `node scripts/cli.js pg-rest-api-types` turns them into a string and since it's automatically generated it might be overridden next time

packages/db/test/pinning.spec.js
alexandrastoica: Thanks just fixed that, will rename

packages/db/postgres/pg-rest-api-types.d.ts
alanshaw: We have the same problem in NFT.Storage https://github.com/nftstorage/nft.storage/blob/ba2f166b1bbfd28a4b7272a65fe8f90b541e5f26/packages/api/src/utils/db-types.d.ts#L902

alexandrastoica: Should I override it in the meantime?

```
"rowFilter.psa_pin_request.origins": string[];
"rowFilter.psa_pin_request.meta": Record<string, string>;
```

alanshaw: I'd leave it as it is generated unless it's failing some CI...?

alanshaw (approved):

PR #906, opened 2022-01-26
Fix PSA pin statuses when unpinned

flea89:
Ticket

### DIFF
--- a/packages/api/src/pins.js
+++ b/packages/api/src/pins.js
@@ -53,7 +53,7 @@ export const getEffectivePinStatus = (pins) => {
 
   // The cluster returns either Unpinned as "initial status",
   // Remote when it has tracked the cid or PinQueued when  it has queued it for pinning.
-  // All this statuses are translate to "queued" in respect to PSA statuses.
+  // All this statuses are translated to "queued" in respect to PSA statuses.
   if (pinStatuses.includes('PinQueued') ||
       pinStatuses.includes('Remote') ||
       pinStatuses.includes('Unpinned')) {
--- a/packages/api/src/pins.js
+++ b/packages/api/src/pins.js
@@ -51,8 +51,12 @@ export const getEffectivePinStatus = (pins) => {
     return 'pinning'
   }
 
+  // The cluster returns either Unpinned as "initial status",
+  // Remote when it has tracked the cid or PinQueued when  it has queued it for pinning.
+  // All this statuses are translate to "queued" in respect to PSA statuses.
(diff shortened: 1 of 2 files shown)


packages/api/test/pin.spec.js
```diff
@@ -667,6 +667,16 @@ describe('Pinning APIs endpoints', () => {
       assert.strictEqual(getEffectivePinStatus(pins), 'queued')
     })
 
+    it('should return "queued" at least 1 pin has Unpinned status', () => {
```
alexandrastoica, line 670: Did you mean to say "IF at least one pin...?"

flea89: good catch!

packages/api/src/pins.js
```diff
@@ -51,8 +51,12 @@ export const getEffectivePinStatus = (pins) => {
     return 'pinning'
   }
 
+  // The cluster returns either Unpinned as "initial status",
+  // Remote when it has tracked the cid or PinQueued when  it has queued it for pinning.
```
vasco-santos, line 55:
```suggestion
  // The cluster returns as an initial status either "Unpinned",
  // "Remote" when it has tracked the cid or "PinQueued" when  it has queued it for pinning.
```

vasco-santos: nit: I think it reads more clear this way

packages/api/src/pins.js
```diff
@@ -51,8 +51,12 @@ export const getEffectivePinStatus = (pins) => {
...
+  // All this statuses are translate to "queued" in respect to PSA statuses.
```
vasco-santos, line 56:
```suggestion
  // All this statuses are translated to "queued" in respect to PSA statuses.
```

packages/api/src/pins.js
alanshaw: Remote means another node in the cluster is pinning it.

alanshaw: I think anything after `Queued` - at that point it has been allocated and all the nodes know where it should be pinned.

packages/api/src/pins.js
alanshaw, line 56:
```suggestion
  // All these statuses are translated to "queued" in respect to PSA statuses.
```

alanshaw (approved):

packages/api/src/pins.js
flea89: I've updated the whole block, hope it is clearer now.

PR #907, opened 2022-01-26
chore: improve release ci

hugomrdias:
This PR scopes release please to each package workflow.

- this way it depends on tests and checks actually passing
- enables multiple release PRs to be merged at the same time
- it only creates release PRs for the packages that we want
- github releases titles are also better 😅 no more `api api-v1.0.0`


vasco-santos (approved):
LGTM 👌🏼

olizilla:
@hugomrdias can you add a description here about how this improves things plz. Not doubting that it does, just I can't guess from the PR why this helps.

hugomrdias:
> @hugomrdias can you add a description here about how this improves things plz. Not doubting that it does, just I can't guess from the PR why this helps.

done in the OP

PR #914, opened 2022-01-26
fix: pin ipfs and ipfs cluster image versions

flea89:


vasco-santos (commented):
Why fixing the versions?

flea89:
Hey Vasco, tests are currently failing on main and that's quite disruptive. As you suggested that's likely because of cluster image caching.
So I was trying to check if by "forcing" the version we'd solve the problem.

On top of that, I think that pinning the version gives us more control over what's happening and we can better align prod environment and local. 
For instance, that would limit the amount of missed bugs that are due to a misalignment of versions between env.

That's the reasoning around the change, I'd love to hear your point of view @vasco-santos ?

vasco-santos:
> On top of that, I think that pinning the version gives us more control over what's happening and we can better align prod environment and local.

We currently don't operate the Cluster, so it is not clear what version it will be running. So, using always `latest` is likely more safe because we will use it in the dev/ci environments before our cluster is updated. In other words, potential regressions will likely be caught by us before production env changes.

> So I was trying to check if by "forcing" the version we'd solve the problem.

Did it solve? If yes, I think we can move with this as a temporary fix to make the cached versions disappear.

But long term I don't think this is a good solution, given that what will happen is that versions won't be updated over time and we will miss out on updates in the cluster. For having manual updates it would create a maintenance burden on us, of needing to track Cluster releases and manually PR every time.

flea89:
Fair enough, I was missing some context here.  Thanks!
I still feel this will bite us back at some point, but I see your points!

> Did it solve?
It did, not sure if it worth merging this in though. I'd expect this to stop being a problem by itself 🤞  
We can keep monitoring the errors and decide accordingly

Issue #918, opened 2022-01-27
Update CRON job to also check pin status for downgraded V0 CID


mbommerez:
There currently is a bug on the CRON job that will check Pin Status for pins that are in status Unpinned. For the moment this job only checks normalised CIDS, instead of (also) checking the source CIDs for their pin status. The outcome is that the pin status is not correctly updated and stays stuck on Unpinned, if the file in question has been uploaded with a CID that is not the default normalised CID format (v1 b32 encoded).

This bug has not been observed in web3.storage because the client libraries create the v1 b32 encoded form by default

The quickest solution consists of the CRON job to also downgrade the CID to v0 and check against that with all cluster peers, in case a user pins a CID that is of a v0 format, rather than the normalised one.

Issue #900, opened 2022-01-25
Shared NFT.Storage + Web3.Storage library


dchoi27:
Start with PSA API implementation validation, create broader test library, then can implement shared features

dchoi27:
We are achieving the bulk of this with Uploads v2, closing this one

PR #919, opened 2022-01-27
fix: check if unpinned then downgrade to v0

LeslieOA:
Temporary workaround for #918:

> There currently is a bug on the CRON job that will check Pin Status for pins that are in status Unpinned. For the moment this job only checks normalised CIDS, instead of (also) checking the source CIDs for their pin status. The outcome is that the pin status is not correctly updated and stays stuck on Unpinned, if the file in question has been uploaded with a CID that is not the default normalised CID format (v1 b32 encoded).


packages/api/src/utils/cid.js
```diff
@@ -14,3 +14,11 @@ export function normalizeCid (cid) {
     throw new InvalidCidError(cid)
   }
 }
+
+export function denormalizeCid (cid) {
+  try {
+    return CID.parse(cid).toV0.toString()
```
alanshaw, line 20:
```suggestion
    return CID.parse(cid).toV0().toString()
```

packages/cron/src/lib/errors.js
```diff
@@ -0,0 +1,79 @@
...
+   * @param {string} cid
+   */
+  constructor (cid) {
+    super(`Invalid CID: ${cid}`)
+    this.name = 'InvalidCid'
+    this.status = 400
+    this.code = InvalidCidError.CODE
+  }
+}
+InvalidCidError.CODE = 'ERROR_INVALID_CID'
```
alanshaw, line 79: How come ALL the errors been repeated here?

alanshaw (changes requested):

packages/cron/src/lib/errors.js
LeslieOA: Only repeating `InvalidCidError` now. 
Would be good to have [shared utils](https://github.com/web3-storage/web3.storage/pull/919#discussion_r793786870) in the future (maybe even across Web3 and NFT).

alanshaw: It is planned https://github.com/web3-storage/web3.storage/issues/900 😉

packages/cron/src/jobs/pins.js
```diff
@@ -63,9 +64,22 @@ export async function updatePinStatuses ({ cluster, db }) {
         return null // not tracked by our cluster
       }
 
-      const status = toPinStatusEnum(peerMap[pin.location.peerId].status)
+      let status = toPinStatusEnum(peerMap[pin.location.peerId].status)
 
       if (status !== 'Pinned' && status !== 'Remote') {
+        const cidV0 = denormalizeCid(pin.contentCid)
```
alanshaw, line 70: Not all CIDs can be downgraded. You need to try/catch here (or do the check).

See https://github.com/multiformats/cid#cidv0

As we've discussed, the likelyhood is high that this is a v0 CID but we can't guarantee. We should not halt processing in the exceptional case.

packages/cron/src/jobs/pins.js
```diff
@@ -63,9 +64,22 @@ export async function updatePinStatuses ({ cluster, db }) {
         return null // not tracked by our cluster
       }
 
-      const status = toPinStatusEnum(peerMap[pin.location.peerId].status)
+      let status = toPinStatusEnum(peerMap[pin.location.peerId].status)
 
       if (status !== 'Pinned' && status !== 'Remote') {
```
alanshaw, line 69: If the status is `Unpinned` that's when we need to try the other version.

THEN we need this check (`if (status !== 'Pinned' && status !== 'Remote') {`) as it was before.

packages/cron/src/lib/cid.js
```diff
@@ -0,0 +1,24 @@
...
+export function normalizeCid (cid) {
+  try {
+    const c = CID.parse(cid)
+    return c.toV1().toString()
+  } catch (err) {
+    throw new InvalidCidError(cid)
+  }
+}
+
+export function denormalizeCid (cid) {
```
alanshaw, line 18: It's denormalized in the sense that it's an encoding that's not the normalized encoding...but I think "downgrade" is a better description of what's happening here.

```suggestion
export function downgradeCid (cid) {
```

packages/cron/src/lib/cid.js
```diff
@@ -0,0 +1,24 @@
...
+  try {
+    return CID.parse(cid).toV0().toString()
+  } catch (err) {
+    throw new InvalidCidError(cid)
+  }
```
alanshaw, line 23: It's not an invalid CID if it can't be downgraded. We should let this throw if the CID cannot be downgraded to v0.

```suggestion
  try {
    cid = CID.parse(cid)
  } catch (err) {
    throw new InvalidCidError(cid)
  }
  return cid.toV0().toString()
```

packages/api/src/utils/cid.js
```diff
@@ -14,3 +14,11 @@ export function normalizeCid (cid) {
...
   }
 }
+
+export function denormalizeCid (cid) {
+  try {
+    return CID.parse(cid).toV0().toString()
+  } catch (err) {
+    throw new InvalidCidError(cid)
+  }
+}
```
alanshaw, line 24: This is unused right? Lets remove.

alanshaw (changes requested):

packages/cron/src/jobs/pins.js
```diff
@@ -52,21 +53,29 @@ export async function updatePinStatuses ({ cluster, db }) {
...
-        reSyncPins.push(pin)
+        const cid = status === 'Unpinned' ? downgradeCid(pin.contentCid) : pin.contentCid
+        try {
+          peerMap = await getPinStatus(cid)
+        } catch (err) {
+          reSyncPins.push(pin)
+          return null
+        }
```
alanshaw, line 78: We end up unnecessarily calling `getPinStatus` twice in the case where status is not unpinned, pinned or remote. We also don't push onto `reSyncPins` unless there is an error calling `getPinStatus`.

We need to do this check after the first call to `getPinStatus` _if_ the status is `Unpinned` and then continue as before.

i.e.

```js
      try {
        peerMap = await getPinStatus(pin.contentCid)
      } catch (err) {
        reSyncPins.push(pin)
        return null // Cluster could not find the content, please check later
      }

      if (!peerMap[pin.location.peerId]) {
        return null // not tracked by our cluster
      }

      let status = toPinStatusEnum(peerMap[pin.location.peerId].status)

      // If "Unpinned", try a v0 CID
      if (status === 'Unpinned') {
        let cidv0
        try {
          cidv0 = downgradeCid(pin.contentCid)
        } catch (err) {
          // Cluster says this is unpinned, but it is not a v0 CID.
          log(`⚠️ Unable to downgrade CID: ${pin.contentCid}`)
          reSyncPins.push(pin)
          return null
        }

        try {
          peerMap = await getPinStatus(cidv0)
        } catch (err) {
          reSyncPins.push(pin)
          return null
        }

        if (!peerMap[pin.location.peerId]) {
          return null // not tracked by our cluster
        }

        status = toPinStatusEnum(peerMap[pin.location.peerId].status)
      }

      if (status !== 'Pinned' && status !== 'Remote') {
        // as before...
```

alanshaw (changes requested):

alanshaw (approved):

PR #927, opened 2022-01-28
feat: pins list filter by meta

GaryHomewood:


packages/api/src/pins.js
```diff
@@ -355,6 +356,16 @@ function parseSearchParams (params) {
     opts.limit = limit
   }
 
+  if (meta) {
+    if (!JSON.parse(meta) || Array.isArray(meta) || Object.entries(meta).some(([, v]) => typeof v !== 'string')) {
```
vasco-santos, line 360: Might be worth adding a comment that spec only accepts JSON object param

packages/db/index.js
```diff
@@ -935,7 +936,9 @@ export class DBClient {
       query = query.gte('inserted_at', opts.after)
     }
 
-    // TODO(https://github.com/web3-storage/web3.storage/issues/798): filter by meta is missing
+    if (opts.meta) {
+      query = query.eq('meta', opts.meta)
```
vasco-santos, line 940: I thinks this will not be enough per https://ipfs.github.io/pinning-services-api-spec/#section/Custom-metadata/Pin-metadata

- lookup returns pins that have meta with all key-value pairs matching the passed values
- pin metadata may have more keys, but only ones passed in the query are used for filtering

vasco-santos (changes requested):

packages/api/src/pins.js
GaryHomewood: Yep comment added.

packages/db/index.js
GaryHomewood: Now corrected, filters on all the key/values passed.

alanshaw (dismissed):

packages/api/src/pins.js
```diff
@@ -355,6 +356,17 @@ function parseSearchParams (params) {
     opts.limit = limit
   }
 
+  if (meta) {
+    // Must be a string representation of a JSON object.
+    if (!JSON.parse(meta) || Array.isArray(meta) || Object.entries(meta).some(([, v]) => typeof v !== 'string')) {
```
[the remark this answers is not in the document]
GaryHomewood: Good point, have improved parsing and validation, also now pass object to db list method.

packages/api/src/utils/psa.js
```diff
@@ -204,8 +205,29 @@ export function validateSearchParams (queryString) {
...
+    try {
+      metaJson = JSON.parse(meta)
+    } catch (e) {
+      return {
+        error: new PSAErrorInvalidData(INVALID_META),
+        data: undefined
+      }
+    }
+
+    if (Object.entries(metaJson).some(([, v]) => typeof v !== 'string')) {
```
alexandrastoica, line 220: Do we need this validation? Based [on nft's code](https://github.com/nftstorage/nft.storage/blob/main/packages/api/src/routes/pins-list.js#L69) it doesn't seem like meta values need to be strings (I know it used to be like that originally, but I don't think we've ever confirmed it)

If the values/keys do need to be strings, I think it can be achieved by adding something like this to the validator:
```
    meta: {
      type: 'object',
      keys: { type: 'string' },
      values: { type: 'string' }
    },
```
But I'd test this to make sure ^

alexandrastoica (changes requested):

packages/api/src/utils/psa.js
GaryHomewood: Good suggestion, have added to the validator to enforce string type values. 

`JSON.parse()` will validate for string value keys.

packages/api/test/pin.spec.js
```diff
@@ -140,6 +140,54 @@ describe('Pinning APIs endpoints', () => {
...
+            Authorization: `Bearer ${token}`,
+            'Content-Type': 'application/json'
+          }
+        })
+
+      assert(res, 'Server responded')
+      assert.strictEqual(res.status, ERROR_CODE)
+      const error = await res.json()
+      assert.strictEqual(error.reason, PSAErrorInvalidData.CODE)
+      assert.strictEqual(error.details, 'Meta should be an object with string values')
```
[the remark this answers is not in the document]
alexandrastoica: Cool, nice to see the validator works so well

alexandrastoica (approved):
LGTM

vasco-santos (approved):
LGTM

PR #928, opened 2022-01-28
fix: update package.json main entry

flea89:
After merging [PR](https://github.com/web3-storage/web3.storage/pull/902/files) trying to run the local API I got 

```
> npm start
....
Uncaught SyntaxError: Unexpected token ':'
  at line 1
 [API code: 10021]
 ```
 
The problem got resolved by updating the main api entry to `index.mjs` (as pointed out to the [doc])(https://developers.cloudflare.com/workers/cli-wrangler/configuration#modules)

Interestingly after checkouting the main branch now it works without this, but I guess that's because of caching

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "type": "module",
   "license": "(Apache-2.0 OR MIT)",
-  "main": "dist/worker.js",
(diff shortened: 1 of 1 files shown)


alanshaw (approved):
❤️ thank you and sorry!

PR #929, opened 2022-01-31
Fix/pin list count

flea89:
In this PR:
- Limit the list pin query (by using a `LIMIT`ing query and not by doing programmatically)
- ~~Make sure the count is returning the right view count expected by the APIs (The assumption here is: when a user provides `before`/`after` the total count should still be the total for the provided filters without the `after` and `before`) Note that is this is the wrong assumption and it should indeed return the count of the items filtered by the provided dates the code can be simplified and we can avoid running 2 queries.~~
- Made sure the right count is returned by the APIs, matching [this](https://github.com/ipfs-shipyard/js-mock-ipfs-pinning-service/blob/main/service/pins.js#L106) and pinata's implementations. 
- Fixed an issue where the default status value is provided only if no options are provided. IMO a default pinned value should be passed also when either/or `limit`, `after` and `before` are provided by the user. Thoughts?
- cleaned up some tests

### DIFF
--- a/packages/db/index.js
+++ b/packages/db/index.js
@@ -881,7 +881,7 @@ export class DBClient {
    * Get a filtered list of pin requests for a user
    *
    * @param {string} authKey
-   * @param {import('./db-client-types').ListPsaPinRequestOptions} opts
+   * @param {import('./db-client-types').ListPsaPinRequestOptions} [opts]
    * @return {Promise<import('./db-client-types').ListPsaPinRequestResults> }> }
    */
   async listPsaPinRequests (authKey, opts = {}) {
--- a/packages/db/index.js
+++ b/packages/db/index.js
@@ -888,90 +888,60 @@ export class DBClient {
     const match = opts?.match || 'exact'
     const limit = opts?.limit || 10
 
-    /**
-     * PSA Pinning APIs spec does not have concept of page as an offset,
-     * but rather it relies on `before` and `after` options to offset the results "pages".
-     * Unfortunately this means that if we add our before/after filters to the query and then we count,
-     * instead of getting the total size of the table or view we get the size of a "page".
-     * To workaround this issue for now we run 2 queries, one for selecting where we pass the dates and
-     * another one for counting where we exclude them.
-     *
-     * @param {object} opt
-     * @param {boolean} [opt.count] Whether or not to return the count
-     * @returns
-     */
-    const createQueryNoDates = ({ count = false } = {}) => {
-      let query = this._client
-        .from(psaPinRequestTableName)
-        .select(listPinsQuery, count
-          ? {
-              count: 'exact'
-            }
-          : {})
-        .eq('auth_key_id', authKey)
-        .is('deleted_at', null)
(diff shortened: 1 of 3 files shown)


packages/db/index.js
```diff
@@ -890,63 +890,61 @@ export class DBClient {
 
     let query = this._client
       .from(psaPinRequestTableName)
-      .select(listPinsQuery)
+      .select(listPinsQuery, {
+        count: 'exact'
+      })
       .eq('auth_key_id', authKey)
       .is('deleted_at', null)
+      .limit(limit)
```
vasco-santos, line 898: Don't know if this is against the spec in any way, but if not we should consider adding a max to the limit (in the API). We do that for the [uploads as well[(https://github.com/web3-storage/web3.storage/blob/main/packages/api/src/user.js#L170), to avoid long queries that might overwhelm the system

packages/db/index.js
```diff
@@ -890,63 +890,61 @@ export class DBClient {
```
vasco-santos, line 890: can we document `opts` param above as optional? 
```js
@param {import('./db-client-types').ListPsaPinRequestOptions} [opts]
```

packages/db/test/pinning.spec.js
```diff
@@ -364,43 +315,64 @@ describe('Pin Request', () => {
...
     })
 
     it('filters items by provided status', async () => {
-      const { results: pins } = await client.listPsaPinRequests(authKeyPinList, {
+      const { results: pins, count } = await client.listPsaPinRequests(authKeyPinList, {
         statuses: ['Pinning', 'PinError']
       })
 
       assert.strictEqual(pins.length, 2)
+      assert.strictEqual(count, 2)
```
vasco-santos, line 352: nit: it would be nice to not hardcode the number and count the actual value from `pinRequestsInputs`, to avoid future test failures with more fixtures added

vasco-santos (changes requested):
Overall looking good, few comments below

packages/db/index.js
flea89: This is a really valid point and we're already tackling it as part of this [PR](https://github.com/web3-storage/web3.storage/pull/937/files#diff-9a666433f02a3fe427683b8f92fddccd7b1c9a9618d4b8b36b5f779b38559eaaR18) [here](https://github.com/web3-storage/web3.storage/pull/937/files#diff-9a666433f02a3fe427683b8f92fddccd7b1c9a9618d4b8b36b5f779b38559eaaR18)

Can we disregard it in the context of this one?

packages/db/test/pinning.spec.js
flea89: I see your point, at the same time this feels like the usual balance of maintainable tests vs striving to not write tests that are too clever. I feel in this case it is better to hardcode it.

packages/db/index.js
vasco-santos: Ok, let's track it in #937

vasco-santos (approved):
LGTM

Issue #659, opened 2021-11-19
Simple Mutability with IPNS records


alanshaw:
Add a simple mutability primitive to Web3.Storage that does not tie users into Web3.Storage or centralized technologies. Build a HTTP API for Web3.Storage that allows developers to create a update mutable references using IPNS records.

More details: https://www.notion.so/protocollabs/Mutable-References-0b4df016bf9a4d538d2b7c28c8974ead

- [ ] MVP - HTTP endpoints (create + update) and client support https://github.com/web3-storage/web3.storage/pull/648
- [ ] Websockets support on server and client (listen for changes to a given key) https://github.com/web3-storage/web3.storage/pull/653
- [ ] Bundle size optimizations (pull in just the dependencies we need)
- [ ] Update [docs.web3.storage](http://docs.web3.storage) with client library additions and new tutorial(s)
- [ ] IPNS republishing (publish updated records to the DHT/pubsub so available for resolution via gateway)
- [ ] Spec site (something like [https://ipfs.github.io/pinning-services-api-spec/](https://ipfs.github.io/pinning-services-api-spec/))

Optional

- HTTP endpoints for creating keys, updating a record by sending value + private key (easy usage from the command line with `curl`)

resolves #511

jchris:
Maybe w3name records should not expire? That seems like a potential rug pull when they expire in a year.

PR #932, opened 2022-01-31
feat: websockets support for w3name

alanshaw:
This PR adds support for websockets to w3name. Establish a websocket connection to `wss://api.web3.storage/name/:key/watch` and receive JSON encoded messages like: `{ key: string, value: string, record: string }` when updates to the key are published.

* [x] Refactor to ESM to allow usage of "Durable Objects" https://github.com/web3-storage/web3.storage/pull/902
* [x] Configuration to enable `NameRoom` as a durable object
* [ ] Client API (will do in separate PR)

supersedes https://github.com/web3-storage/web3.storage/pull/653

Related to https://github.com/web3-storage/web3.storage/issues/659

### DIFF
--- a/packages/api/src/env.js
+++ b/packages/api/src/env.js
@@ -36,6 +36,29 @@ import pkg from '../package.json'
  * @property {S3Client} [s3Client]
  * @property {string} [s3BucketName]
  * @property {string} [s3BucketRegion]
+ * // Durable Objects
+ * @property {DurableObjectNamespace} NAME_ROOM
+ *
+ * From: https://github.com/cloudflare/workers-types
+ *
+ * @typedef {{
+ *  toString(): string
+ *  equals(other: DurableObjectId): boolean
+ *  readonly name?: string
+ * }} DurableObjectId
+ *
+ * @typedef {{
+ *   newUniqueId(options?: { jurisdiction?: string }): DurableObjectId
+ *   idFromName(name: string): DurableObjectId
+ *   idFromString(id: string): DurableObjectId
+ *   get(id: DurableObjectId): DurableObjectStub
+ * }} DurableObjectNamespace
+ *
+ * @typedef {{
+ *   readonly id: DurableObjectId
+ *   readonly name?: string
+ *   fetch(requestOrUrl: Request | string, requestInit?: RequestInit | Request): Promise<Response>
+ * }} DurableObjectStub
  */
 
 /**
--- a/packages/api/src/name.js
+++ b/packages/api/src/name.js
@@ -157,8 +157,7 @@ export class NameRoom {
   static async join (req, ns, key) {
     const roomId = ns.idFromName(key)
     const room = ns.get(roomId)
-    const url = new URL(req.url)
-    url.pathname = '/websocket'
+    const url = new URL('/websocket', req.url)
     return room.fetch(url, req)
   }
 
@@ -171,8 +170,7 @@ export class NameRoom {
   static async broadcast (req, ns, key, data) {
     const roomId = ns.idFromName(key)
     const room = ns.get(roomId)
-    const url = new URL(req.url)
-    url.pathname = '/broadcast'
+    const url = new URL('/broadcast', req.url)
     return room.fetch(url, { method: 'POST', body: JSON.stringify(data) })
   }
 }
--- a/packages/api/src/name.js
+++ b/packages/api/src/name.js
@@ -69,10 +69,11 @@ export async function namePost (request, env, ctx) {
   )
 
   ctx.waitUntil((async () => {
+  // await ((async () => {
     const record = await env.db.resolveNameRecord(key)
     if (!record) return // shouldn't happen
     const { value } = ipns.unmarshal(uint8ArrayFromString(record, 'base64pad'))
-    const data = { value: uint8ArrayToString(value), record }
+    const data = { key, value: uint8ArrayToString(value), record }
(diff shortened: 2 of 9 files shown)


packages/api/src/name.js
```diff
@@ -65,5 +68,111 @@ export async function namePost (request, env) {
...
+  /**
+   * @param {Request} req
+   * @param {import('./env').DurableObjectNamespace} ns
+   * @param {string} key IPNS Key ID
+   */
+  static async join (req, ns, key) {
+    const roomId = ns.idFromName(key)
+    const room = ns.get(roomId)
+    const url = new URL(req.url)
+    url.pathname = '/websocket'
```
vasco-santos, line 161:
```suggestion
    const url = new URL('/websocket', req.url)
```

packages/api/src/name.js
```diff
@@ -65,5 +68,111 @@ export async function namePost (request, env) {
...
+   * @param {Request} req
+   * @param {import('./env').DurableObjectNamespace} ns
+   * @param {string} key IPNS Key ID
+   * @param {{ value: string, record: string }} data
+   */
+  static async broadcast (req, ns, key, data) {
+    const roomId = ns.idFromName(key)
+    const room = ns.get(roomId)
+    const url = new URL(req.url)
+    url.pathname = '/broadcast'
```
vasco-santos, line 175:
```suggestion
    const url = new URL('/broadcast', req.url)
```

packages/api/src/name.js
```diff
@@ -65,5 +68,111 @@ export async function namePost (request, env) {
...
+}
+
+/**
+ * A Cloudflare Durable Object. Each instance is a group of users interested in
+ * a particular IPNS key ID. The static API is used by workers to communicate
+ * with it.
+ */
+export class NameRoom {
+  constructor () {
+    /** @type {Array<{ webSocket: WebSocket }>} */
```
vasco-santos, line 108: Any reason to be an array of Objects instead of an array of Websocket instances?

packages/api/test/name.spec.js
```diff
@@ -38,3 +42,128 @@ describe('POST /name/:key', () => {
...
+    // so we should only receive 2 messages.
+    const expectedMsgCount = 2
+    /** @type {import('websocket').Message[]} */
+    const msgs = []
+    const deferred = defer()
+    conn.on('message', msg => {
+      msgs.push(msg)
+      if (msgs.length >= expectedMsgCount) {
+        deferred.resolve()
+      }
```
vasco-santos, line 87: should we throw if `msgs.length > expectedMsgCount` and change this condition to `===` instead?

packages/api/test/name.spec.js
```diff
@@ -38,3 +42,128 @@ describe('POST /name/:key', () => {
...
+    const name0Value0 = '/ipfs/bafybeiauyddeo2axgargy56kwxirquxaxso3nobtjtjvoqu552oqciudrm'
+    const name0Value1 = '/ipfs/bafkreiem4twkqzsq2aj4shbycd4yvoj2cx72vezicletlhi7dijjciqpui'
+    const name1Value0 = '/ipfs/bafkreid7fbwjx4swwewit5txzttoja4t4xnkj3rx3q7dlbj76gvixuq35y'
+    const name1Value1 = '/ipfs/bafybeihiyjghsq7gob7vj3vurqf5i4eth3h57ixpaajdxtvi7p4snhag2a'
+    const name0Record0 = await createNameRecord(name0.privateKey, name0Value0)
+    const name0Record1 = await updateNameRecord(name0.privateKey, name0Record0, name0Value1)
+    const name1Record0 = await createNameRecord(name1.privateKey, name1Value0)
+    const name1Record1 = await updateNameRecord(name1.privateKey, name1Record0, name1Value1)
+
+    const publishRecord = async (key, record) => {
```
vasco-santos, line 122: Let's have this function only once and not part of a test case?

vasco-santos (approved):
LGTM! Great work @alanshaw  🚀 ❤️ 

Left few suggestions for you to consider

packages/api/src/name.js
alanshaw: Not really, I guess I was thinking we might want to attach some other session information in the future...

packages/api/test/name.spec.js
alanshaw: There's an assertion below that checks `msgs.length` === `expectedMsgCount`. If we throw here we'll get an uncaught exception...

vasco-santos (approved):

packages/api/src/env.js
```diff
@@ -36,6 +36,29 @@ import pkg from '../package.json'
...
+ *   idFromName(name: string): DurableObjectId
+ *   idFromString(id: string): DurableObjectId
+ *   get(id: DurableObjectId): DurableObjectStub
+ * }} DurableObjectNamespace
+ *
+ * @typedef {{
+ *   readonly id: DurableObjectId
+ *   readonly name?: string
+ *   fetch(requestOrUrl: Request | string, requestInit?: RequestInit | Request): Promise<Response>
+ * }} DurableObjectStub
```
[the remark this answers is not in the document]
hugomrdias: these arent available in the global scope with worker-types in the tsconfig ?

hugomrdias (approved):
LGTM just on comment inline

packages/api/src/env.js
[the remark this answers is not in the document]
alanshaw: Not in this project...

PR #933, opened 2022-01-31
Fix/psa unpinned status

flea89:
Fix a bug where initial requests to pin a cid return failed.

Related [PR](https://github.com/web3-storage/web3.storage/pull/906)

### DIFF
--- a/packages/db/utils.js
+++ b/packages/db/utils.js
@@ -12,7 +12,7 @@ export function normalizeUpload (upload) {
     ...nUpload,
     ...upload.content,
     pins: normalizePins(upload.content.pins, {
-      filterOkStatuses: true
+      isOkStatuses: true
     })
   }
 }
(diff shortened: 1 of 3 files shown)


packages/db/utils.js
```diff
@@ -48,10 +50,18 @@ export function normalizeContent (content) {
...
  *
  * @param {Array<import('./db-client-types').PinItem>} pins
+ * @param {object} opt
+ * @param {boolean} [opt.filterOkStatuses]
  * @return {Array<import('./db-client-types').PinItemOutput>}
  */
-export function normalizePins (pins) {
-  return pins.filter(pin => PIN_STATUS.has(pin.status))
+export function normalizePins (pins, {
+  filterOkStatuses = false
```
vasco-santos, line 58: nit: Not strong opinion, but I would rename to `isOkStatuses` instead

vasco-santos (approved):
LGTM, minor details in comments

PR #934, opened 2022-01-31
Fix: docker cluster container name conflict

flea89:
This PR fixes an a docker conflict issue that prevents the cluster to start due to a conflict in containers names.

### DIFF
--- a/packages/tools/docker/cluster/docker-compose.yml
+++ b/packages/tools/docker/cluster/docker-compose.yml
@@ -36,7 +36,6 @@ services:
   ##################################################################################
 
     ipfs0:
-      container_name: ipfs0
       image: ipfs/go-ipfs:v0.10.0 # update this when go-ipfs M1 macs https://github.com/ipfs/go-ipfs/issues/8645
       volumes:
         - ipfs0:/data/ipfs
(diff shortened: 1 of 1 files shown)


vasco-santos (commented):
What is the motivation for getting this conflict? 

As far as I know, we only get this conflict if the clean command did not run. We run it in the tests https://github.com/web3-storage/web3.storage/blob/main/packages/api/test/hooks.js#L63

I would much rather try to understand why this is problematic than leaking our local environments with unused docker containers. 

If you are spinning it up manually, perhaps better to add a script in tools package json to stop and clean

flea89:
I actually changed this [before esm refactor ticket](https://github.com/web3-storage/web3.storage/commit/c1f7a1bdd77d47d7f1a25642b2e6bb371aa36a79) was merged in. And I suspect before that point we were not properly cleaning the containers?

I did not experience the issue lately so I don't think this is a problem anymore. 
I had a look at the [hooks](https://github.com/web3-storage/web3.storage/blob/main/packages/api/test/hooks.js#L63)  as well and it all makes sense.

Basically what you say makes sense and I'm happy to close the PR.

PR #935, opened 2022-02-01
fix: include meta for replace pin

GaryHomewood:


packages/api/src/pins.js
```diff
@@ -144,7 +144,8 @@ async function createPin (normalizedCid, pinData, authTokenId, env, ctx) {
 
   await env.cluster.pin(cid, {
     name: pinName,
-    origins
+    origins,
+    metadata: pinMeta
```
alanshaw, line 148: We can't send metadata to cluster it is not multi-tenant - what happens if someone else pins the same CID?

alanshaw (changes requested):

packages/api/src/pins.js
GaryHomewood: It looks like its passed in NFT `pins-add`. Is that different?

alanshaw: Nope, it should not be sent there either. Can you send a PR to fix? 🙏

GaryHomewood: Ok cool. Removed it here, created an issue on NFT

[https://github.com/nftstorage/nft.storage/issues/1207](https://github.com/nftstorage/nft.storage/issues/1207)

alanshaw (approved):

PR #937, opened 2022-02-01
feat/psa error refactor

alexandrastoica:
- Adds PSA error classes to `errors.js`
- Uses schema validators to parse payload for listing and adding pins
- Fixed a bug where we were using the wrong cid when replacing a pin


packages/api/test/pin.spec.js
```diff
@@ -550,7 +557,7 @@ describe('Pinning APIs endpoints', () => {
     })
 
     it('returns the pin request', async () => {
-      const sourceCid = 'QmVGv1UK8EvhD9KMHdKBB1LWadiYai6sY5GHL6h7MHRWzf'
+      const sourceCid = 'bafybeidhbtemubjbsbuhyai5oaebqf2fdrvhnshbkncyqpnoy2bl2mpt4q'
```
alexandrastoica, line 560: @LeslieOA I had to change this for my test to pass. Not sure if it's coming from the cid normalisation work you did? cc: @flea89

packages/api/src/errors.js
```diff
@@ -77,3 +77,63 @@ export class InvalidCidError extends Error {
...
+  /**
+   * @param {string} details
+   */
+  constructor (details = 'Requested data was not found.') {
+    super(PSAErrorDBNotFound.REASON)
+    this.details = details
+    this.status = 404
+  }
+}
+PSAErrorDBNotFound.REASON = 'PSA_DB_NOT_FOUND'
```
GaryHomewood, line 127: Maybe this could be named differently. I read it as meaning "database not found". Is it `PSAErrorResourceNotFound` ?

alexandrastoica: Thanks! I'm not good with naming things, wasn't happy with this one either 😅

packages/api/src/utils/psa.js
```diff
@@ -56,19 +60,185 @@ export const psaStatusesToDBStatuses = (statuses) => {
...
+  if (!cid) {
+    return {
+      error: new PSAErrorRequiredData(REQUIRED_CID),
+      data: undefined
+    }
+  }
+
+  // Validate CID.
+  try {
+    opts.cid = normalizeCid(cid)
```
flea89, line 136: Similarly to my other comment for the listing, sourceCid should not be overriden by the normalised one given sourceCid is used as a source of truth in cluster/db operations.

When we are creating a pin, we should pass both the source cid and the normalised one. If you have a look to createPin function we pass both the normalised pin as first parameter and the pinObject containing the sourceCid.

packages/api/src/errors.js
```diff
@@ -77,3 +77,63 @@ export class InvalidCidError extends Error {
   }
 }
 InvalidCidError.CODE = 'ERROR_INVALID_CID'
+
+export class PinningServiceApiError {
```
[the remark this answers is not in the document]
alexandrastoica: If I do that, in `error-handler.js`, `err` will be an instance of both `PinningServiceApiError` and `HTTPError` so the conditionals won't work basically.

packages/api/src/error-handler.js
```diff
@@ -1,19 +1,32 @@
...
- * @param {Error & {status?: number;code?: string;}} err
+ * @param {Error & {status?: number;code?: string;reason?: string; details?: string}} err
  * @param {import('./env').Env} env
  */
 export function errorHandler (err, { sentry }) {
   console.error(err.stack)
 
+  let status = err.status || 500
+
+  if (err instanceof PinningServiceApiError) {
```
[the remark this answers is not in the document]
alexandrastoica: sorry, I'm not sure what you mean by this?

packages/api/test/pin.spec.js
LeslieOA: @alexandrastoica Thanks for checking! It's from [Fix/psa unpinned status #933](https://github.com/web3-storage/web3.storage/pull/933) methinks. CC @flea89

packages/api/src/utils/psa.js
```diff
@@ -56,19 +60,185 @@ export const psaStatusesToDBStatuses = (statuses) => {
...
+      try {
+        normalizedCids.push(normalizeCid(cid))
+      } catch (e) {
+        return {
+          error: new PSAErrorInvalidData(INVALID_CID),
+          data: undefined
+        }
+      }
+    }
+    opts.cid = normalizedCids
```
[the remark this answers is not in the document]
alexandrastoica: I just translated this from what [we had](https://github.com/web3-storage/web3.storage/pull/937/files#diff-9a666433f02a3fe427683b8f92fddccd7b1c9a9618d4b8b36b5f779b38559eaaL276), but I'll verify this and remove it. Thanks!

GaryHomewood (approved):
The validator is cool. Nice work.

packages/api/src/utils/psa.js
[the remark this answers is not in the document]
alexandrastoica: Ok so I checked this and you're right, [we filter by `source_cid`](https://github.com/web3-storage/web3.storage/blob/main/packages/db/index.js#L910), but that was there before so I'd say it was a bug?

If I don't normalise the CID for filtering, [there's a test that fails](https://github.com/web3-storage/web3.storage/blob/main/packages/api/test/pin.spec.js#L130), so basically it doesn't check whether the CID(s) passed is/are valid. How do we check source CIDs if they are valid? or should we simply query the DB and send back 404?

Let me know what you think. Thanks!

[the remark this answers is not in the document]
alexandrastoica: Synced offline, my bad, read a line wrong :( Thanks!

packages/api/src/pins.js
```diff
@@ -427,34 +243,27 @@ async function replacePin (newPinData, requestId, authTokenId, env, ctx) {
...
   let pinStatus
   try {
     pinStatus = await createPin(existingPinRequest.contentCid, newPinData, authTokenId, env, ctx)
   } catch (e) {
-    return new JSONResponse(
-      { error: { reason: `DB Error: ${e}` } },
-      { status: 501 }
-    )
+    console.error(e)
+    throw new PSAErrorDB()
```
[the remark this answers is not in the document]
alexandrastoica: Added a better error, thanks!

flea89 (approved):

packages/api/src/errors.js
```diff
@@ -77,3 +77,65 @@ export class InvalidCidError extends Error {
   }
 }
 InvalidCidError.CODE = 'ERROR_INVALID_CID'
+
+export class PinningServiceApiError extends Error {
+  /**
+   *
+   * @param {string} reason
```
vasco-santos, line 84: Can we follow Error namings here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error#parameters and use message instead of reason?

packages/api/src/errors.js
```diff
@@ -77,3 +77,65 @@ export class InvalidCidError extends Error {
...
+   * @param {string} [details]
+   * @param {number} [status]
+   */
+  constructor (reason = 'PSA_ERROR', details, status = 400) {
```
vasco-santos, line 88: status is the second argument in HTTPError, can we have it be in the same position here?

packages/api/test/pin.spec.js
```diff
@@ -144,12 +139,52 @@ describe('Pinning APIs endpoints', () => {
           }
         })
 
-      assert.strictEqual(res.status, ERROR_CODE)
+      assert.strictEqual(res.status, 400)
```
vasco-santos, line 142: Can we still use a constant for the error code?

packages/api/src/utils/psa.js
```diff
@@ -56,19 +61,191 @@ export const psaStatusesToDBStatuses = (statuses) => {
...
+export const REQUIRED_CID = 'CID is required.'
+export const REQUIRED_REQUEST_ID = 'Request id is required.'
+
+export const DEFAULT_PIN_LISTING_LIMIT = 10
+export const MAX_PIN_LISTING_LIMIT = 1000
+
+// Validation schemas
+const listPinsValidator = new Validator({
+  type: 'object',
+  required: [],
```
vasco-santos, line 78: is status required? I was looking into nft.storage schema and it is required there. In the spec https://ipfs.github.io/pinning-services-api-spec/#tag/pins/paths/~1pins/get it is mentioned as `Array of strings (Status) non-empty`

vasco-santos (changes requested):
Looking good overall, left some comments that should be addressed in review

packages/api/src/utils/psa.js
vasco-santos: @lidel Can you please let us know your thoughts here?

packages/api/src/errors.js
```diff
@@ -77,3 +77,65 @@ export class InvalidCidError extends Error {
...
+PSAErrorResourceNotFound.REASON = 'PSA_RESOURCE_NOT_FOUND'
+
+export class PSAErrorDB extends PinningServiceApiError {
+  /**
+   * @param {string} details
+   */
+  constructor (details = 'DB transaction failed.') {
+    super(PSAErrorDB.REASON)
+    this.details = details
+    this.status = 501
```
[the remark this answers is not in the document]
alexandrastoica: Good point, I'll go with `500` (`Internal Server Error`)

packages/api/src/utils/psa.js
lidel: iirc an explicit `status` is required to remove any ambiguity – the field is marked as `non-empty` in the spec for that reason.
if it is missing, you should return HTPP 400 BAD_REQUEST with `error.description` informing user that explicit `status` has to be provided.


The idea here is that the decision around implicit defaults should happen on a client, not the server.
For example, `ipfs pin remote ls --service=foo` without `--status`  defaults to `pinned` and sends `pins?status=pinned` on the wire.

packages/api/src/error-handler.js
```diff
@@ -3,22 +3,32 @@ import { JSONResponse } from './utils/json-response.js'
...
 export function errorHandler (err, { sentry }) {
   console.error(err.stack)
 
+  let status = err.status || 500
+
+  if (sentry && status >= 500) {
+    sentry.captureException(err)
+  }
+
+  if (err.IS_PSA_ERROR) {
```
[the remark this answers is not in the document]
alexandrastoica: I created [a follow up](https://github.com/web3-storage/web3.storage/issues/976) for this and noted it in the code.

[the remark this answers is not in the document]
vasco-santos: yeah, the error already existed, so let's follow up on that issue thanks

alexandrastoica:
@vasco-santos I addressed all comments, feel free to take another look when you get a chance. Thanks!

packages/api/src/utils/psa.js
```diff
@@ -56,19 +61,192 @@ export const psaStatusesToDBStatuses = (statuses) => {
...
+  } else {
+    opts.limit = DEFAULT_PIN_LISTING_LIMIT
+  }
+
+  if (name) opts.name = name
+  if (meta) opts.meta = meta
+  if (match) opts.match = match
+  if (before) opts.before = before
+  if (after) opts.after = after
+  if (status) opts.status = status.split(',')
```
vasco-santos, line 212:
```suggestion
  opts.status = status.split(',')
```

vasco-santos: it will always exist at this stage or failed, right?

packages/api/src/utils/psa.js
```diff
@@ -56,19 +61,192 @@ export const psaStatusesToDBStatuses = (statuses) => {
...
+
+  const result = listPinsValidator.validate(opts)
+
+  let data
+  let error
+
+  if (result.valid) {
+    // Map statuses for DB compatibility.
+    if (status) opts.statuses = psaStatusesToDBStatuses(opts.status)
```
vasco-santos, line 221:
```suggestion
    opts.statuses = psaStatusesToDBStatuses(opts.status)
```

vasco-santos (approved):
LGTM
nit comment inline

packages/api/src/utils/psa.js
alexandrastoica: At this stage it may or may not exist. Removing the conditional will throw `Cannot read properties of undefined (reading 'split')`

packages/api/src/utils/psa.js
alexandrastoica: Here we can remove the `if` indeed

vasco-santos:
Thanks @alexandrastoica 
You just need to rebase this PR to fix the conflict and you can merge this

PR #938, opened 2022-02-01
fix: use upsertpins

GaryHomewood:
We were looping a list of pins and using `upsertPin` to change the status to 'Pinned'. Instead we should use `upsertPins` to batch update all of them in one request.

### DIFF
--- a/packages/api/src/utils/pin.js
+++ b/packages/api/src/utils/pin.js
@@ -135,18 +135,18 @@ export async function waitOkPins (cid, cluster, waitTime = MAX_PIN_STATUS_CHECK_
  * @param {import('@web3-storage/db').DBClient} db
  * @param {number} waitTime
  * @param {number} checkInterval
- * @return {Promise.<import('@web3-storage/db/db-client-types').PinsUpsertInput[]>}
+ * @return {Promise.<import('@web3-storage/db/db-client-types').PinUpsertInput[]>}
  */
 export async function waitAndUpdateOkPins (cid, cluster, db, waitTime = MAX_PIN_STATUS_CHECK_TIME, checkInterval = PIN_STATUS_CHECK_INTERVAL) {
   const okPins = await waitOkPins(cid, cluster, waitTime, checkInterval)
   const pins = okPins.map((pin) => {
     return {
-      _id: pin._id,
+      id: pin._id,
       status: pin.status,
       cid,
       locationId: pin.location._id
     }
   })
   await db.upsertPins(pins)
-  return pins
+  return okPins
 }
--- a/packages/api/src/utils/pin.js
+++ b/packages/api/src/utils/pin.js
@@ -135,10 +135,19 @@ export async function waitOkPins (cid, cluster, waitTime = MAX_PIN_STATUS_CHECK_
  * @param {import('@web3-storage/db').DBClient} db
  * @param {number} waitTime
  * @param {number} checkInterval
- * @return {Promise.<import('@web3-storage/db/db-client-types').PinUpsertInput[]>}
+ * @return {Promise.<import('@web3-storage/db/db-client-types').PinsUpsertInput[]>}
  */
 export async function waitAndUpdateOkPins (cid, cluster, db, waitTime = MAX_PIN_STATUS_CHECK_TIME, checkInterval = PIN_STATUS_CHECK_INTERVAL) {
   const okPins = await waitOkPins(cid, cluster, waitTime, checkInterval)
-  await db.upsertPins(okPins)
-  return okPins
+  const pins = okPins.map((pin) => {
+    return {
+      id: pin.id,
+      status: pin.status,
+      cid,
+      locationId: pin.location.id,
+      updated: new Date().toISOString()
+    }
+  })
+  await db.upsertPins(pins)
+  return pins
 }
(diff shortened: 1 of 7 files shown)


flea89 (approved):
LGTM!

packages/api/src/utils/pin.js
```diff
@@ -139,8 +139,6 @@ export async function waitOkPins (cid, cluster, waitTime = MAX_PIN_STATUS_CHECK_
  */
 export async function waitAndUpdateOkPins (cid, cluster, db, waitTime = MAX_PIN_STATUS_CHECK_TIME, checkInterval = PIN_STATUS_CHECK_INTERVAL) {
   const okPins = await waitOkPins(cid, cluster, waitTime, checkInterval)
-  for (const pin of okPins) {
-    await db.upsertPin(cid, pin)
-  }
+  await db.upsertPins(okPins)
```
flea89, line 142: I think there's something going on with passing the wrong objects here.

If you compare [upserPin](https://github.com/web3-storage/web3.storage/blob/1373dbbffa804b04f2f23d80d142fc960d28af6c/packages/db/index.js#L419) and [upsertPins](https://github.com/web3-storage/web3.storage/blob/main/packages/db/index.js#L447) they are quite different.

[upserPin](https://github.com/web3-storage/web3.storage/blob/1373dbbffa804b04f2f23d80d142fc960d28af6c/packages/db/index.js#L419) call a remote procedure and takes care of creating/updating the location in the DB while [upsertPins](https://github.com/web3-storage/web3.storage/blob/main/packages/db/index.js#L447) is a straight upsert in the `pin` table.

The problem here is also that the parameter type annotation is wrong [here](https://github.com/web3-storage/web3.storage/blob/1373dbbffa804b04f2f23d80d142fc960d28af6c/packages/db/index.js#L445), the function does not expect an `PinUpsertInput[]`, especially if you look at the item shape of that array, it should have a `pin_location_id: string` instead of a `location: Location`. 

Please have a look [here](https://github.com/web3-storage/web3.storage/blob/1373dbbffa804b04f2f23d80d142fc960d28af6c/packages/cron/src/jobs/pins.js#L101) to have an example of what is actually expected. 

We need to make sure we update the function type annotation and we pass the right objects around.

flea89 (changes requested):

packages/api/src/utils/pin.js
GaryHomewood: Created a new type `PinsUpsertInput` and used this instead.

packages/db/index.js
```diff
@@ -438,7 +438,7 @@ export class DBClient {
   /**
    * Upsert given pin status.
    *
-   * @param {Array<import('./db-client-types').PinUpsertInput>} pins
+   * @param {Array<import('./db-client-types').PinsUpsertInput>} pins
    */
   async upsertPins (pins) {
     const { error } = await this._client
```
flea89, line 444: We need to make sure we're mapping the objet to the db columns names here if we stick with using the client.

flea89 (changes requested):

packages/db/index.js
```diff
@@ -438,7 +438,7 @@ export class DBClient {
   /**
    * Upsert given pin status.
```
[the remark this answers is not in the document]
GaryHomewood: Note added to make it explicit it's currenlt only used to batch update status.

packages/db/index.js
GaryHomewood: Object to column mapping added.

flea89 (approved):
LGTM,
thanks!

packages/api/src/utils/pin.js
```diff
@@ -135,12 +135,18 @@ export async function waitOkPins (cid, cluster, waitTime = MAX_PIN_STATUS_CHECK_
...
+  const pins = okPins.map((pin) => {
+    return {
+      _id: pin._id,
+      status: pin.status,
+      cid,
+      locationId: pin.location._id
+    }
+  })
+  await db.upsertPins(pins)
+  return pins
```
vasco-santos, line 151: Any reason for changing the return value?

vasco-santos (commented):
I know the motivation to use upsertPins when I read the code. But it is important to give the context to reviewers and community on the PR description. Can you please add it?

packages/cron/src/jobs/pins.js
```diff
@@ -99,11 +99,10 @@ export async function updatePinStatuses ({ cluster, db }) {
       log(`📌 ${pin.contentCid}@${pin.location.peerId}: ${pin.status} => ${status}`)
 
       return {
-        id: pin._id,
+        _id: pin._id,
```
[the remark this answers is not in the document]
GaryHomewood: @vasco-santos  I had it as `_id` to indicate that it is a `string` not a number, and to keep it consistent with the other DB client types we have for pins. Should we just have `id` throughout, for all the types? Should that be a separate ticket?

packages/api/src/utils/pin.js
GaryHomewood: Not specifically, so I have reverted it.

packages/cron/src/jobs/pins.js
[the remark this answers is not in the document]
GaryHomewood: Corrected to use `id` not `_id`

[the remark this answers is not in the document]
vasco-santos: @GaryHomewood let's discuss this in another ticket. The thing is, we don't want to break API contract and we were using `_id` before due to Fauna. I think for minimizing these issues, we should use `id` everywhere in the DB package, and only transform in the API before sending back to users. What do you think? Let's sync in an issue to open convo for everyone

vasco-santos (approved):
Awesome LGTM!

PR #944, opened 2022-02-02
fix: highlight selection tweaks

svvimming:
Issues flagged in [this task](https://app.asana.com/0/0/1201717596388768/f) with text selection have been fixed. A new mixin called `gradient_Text_Selection` has been added and must be included on a parent of any gradient text element in order for the correct highlighting to appear when selecting it.


orvn (approved):
All areas tested successfully

PR #945, opened 2022-02-02
feat: marketing site nav account dropdown

svvimming:
Account dropdown added to the nav on the marketing site. Also a small issue was fixed with the accordion expand all option displaying in the nav account dropdown on mobile.


orvn (approved):
The design (specifically the reduced opacity background on the subnav) may change in the future, but for now this is sufficient for both light and dark headers/hero areas

marketing-site-subnav-contrast

Issue #563, opened 2021-10-25
Add admin website for web3.storage


atopal:
## User Story
As an admin user, I want an admin area on the website, so I can in the future manage users, turn and off features and administer the site. 

## Scope:  
* Unified admin for both web3 and nft.storage.  (But initial implementation needed only for web3) 

## Rough Approach:
* spin up next.js site, 
* confirm UI then connect to cloudflare (hugo to sync with Aaron to provide access)
* house codebase in the nft.storage github organization (per Mikel on 11.19)

AC:
- [ ] admins can see admin area
- [ ] users who have access to cloulflare worker we create for this can access the admin website (POSSIBLY using the cloudflare teams feature)

Linked to #536
Linked to nft.storage issue https://github.com/nftstorage/nft.storage/issues/606
Impacted by https://github.com/nftstorage/nft.storage/issues/826

JeffLowe:
FYI, check in with Alan mid next week on progress to determine whether we need to split task up more (leading into next week's cycle goal)

JeffLowe:
@alanshaw Are there any preferences in auth to use for the admin?  Apparently we still need to pick an auth solution even if that's a One Time Pin to integrate with Cloudflare for Teams: https://developers.cloudflare.com/cloudflare-one/identity

JeffLowe:
Per convo with Mikel today, we'll house this codebase in nft org.  Thus to keep tasking DRY, eliminating the tasks here in web3 and leaving only the tasks in nft org.  First release will still provide ban key capabilities for web3.

PR #947, opened 2022-02-03
Feature #563: Schema updates for admin site.

trigramdev9:
Porting over from NFT to support api token blocking in the admin site: https://github.com/nftstorage/nft.storage/pull/1203

### DIFF
--- a/packages/db/postgres/tables.sql
+++ b/packages/db/postgres/tables.sql
@@ -1,3 +1,13 @@
+-- Auth key blocked status type is the type of blocking that has occurred on the api
+-- key.  These are primarily used by the admin app.
+CREATE TYPE auth_key_blocked_status_type AS ENUM
+(
+  -- The api key is blocked.
+  'Blocked',
+  -- The api key is unblocked.
+  'Unblocked'
+);
+
 -- A user of web3.storage.
 CREATE TABLE IF NOT EXISTS public.user
 (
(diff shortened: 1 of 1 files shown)


packages/db/postgres/tables.sql
```diff
@@ -260,3 +280,19 @@ CREATE TABLE IF NOT EXISTS pinning_authorization
   inserted_at     TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL,
   deleted_at      TIMESTAMP WITH TIME ZONE
 )
```
vasco-santos, line 282:
```suggestion
);
```

vasco-santos (changes requested):
Thanks for the PR @trigramdev9 

It was missing a `;` in the previous SQL statement that is breaking CI

packages/db/postgres/tables.sql
```diff
@@ -260,3 +280,19 @@ CREATE TABLE IF NOT EXISTS pinning_authorization
...
+
+CREATE VIEW admin_search as
+select
+  u.id::text as user_id,
+  u.github_id as username,
```
vasco-santos, line 287:
```suggestion
  u.github as username,
```

Not sure if this is what you want, just to make sure. user.github is a numeric github ID, not the handle. We don't track the handle in web3.storage

packages/db/postgres/tables.sql
trigramdev9: woops.  my bad!

packages/db/postgres/tables.sql
trigramdev9: removed this field.

vasco-santos (approved):
LGTM

PR #955, opened 2022-02-04
Adding user_tag table to facilitate user restricted application features.

trigramdev9:
* pinning allowlist and user account storage limits can be managed by this single table without the need for one-off tables e.g. pinning_authorization.
* this allows the admin.storage site to easily manage tags.

nft pr [here](https://github.com/nftstorage/nft.storage/pull/1243)

### DIFF
--- a/packages/db/postgres/tables.sql
+++ b/packages/db/postgres/tables.sql
@@ -8,6 +8,14 @@ CREATE TYPE auth_key_blocked_status_type AS ENUM
   'Unblocked'
 );
 
+-- User tags are associated to a user for the purpose of granting/restricting them
+-- in the application.
+CREATE TYPE user_tag_type AS ENUM
+(
+  'PINNING',
+  'STORAGE_LIMIT'
+);
+
 -- A user of web3.storage.
 CREATE TABLE IF NOT EXISTS public.user
 (
@@ -23,6 +31,17 @@ CREATE TABLE IF NOT EXISTS public.user
   public_address  TEXT                                                          NOT NULL UNIQUE,
   inserted_at     TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL,
   updated_at      TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL
+);
+
+CREATE TABLE IF NOT EXISTS public.user_tag
+(
+  id              BIGSERIAL PRIMARY KEY,
+  user_id         BIGINT                                                        NOT NULL REFERENCES public.user (id),
+  tag             user_tag_type                                                 NOT NULL,
+  -- tag_value is useful for certain tags like STORAGE_LIMIT e.g. tag="STORAGE_LIMIT", tag_value="1TB"
+  tag_value       TEXT                                                                  ,
+  inserted_at TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now())     NOT NULL,
+  deleted_at  TIMESTAMP WITH TIME ZONE
 );
 
 CREATE INDEX IF NOT EXISTS user_updated_at_idx ON public.user (updated_at);


LeslieOA:
@trigramdev9 This makes a lot of sense and makes extending feature access trivial in the future.
I guess "tag" is usefully more generic than `auth` or `permissions` in this case.
Would we need a massive enum of all potential values for `tag_value` also, or would that be excessive?

trigramdev9:
> Would we need a massive enum of all potential options for tag_value also, or would that be excessive?

That's a great question.  I asked dchoi on slack if there is a finite set of values (i.e. for storage limits) and he said no, so I'm somewhat assuming it may be more flexible to have `tag_value` remain a `text` field.

dashcraft (approved):
Identical, LGTM!

packages/db/postgres/tables.sql
```diff
@@ -25,6 +33,17 @@ CREATE TABLE IF NOT EXISTS public.user
...
 
+CREATE TABLE IF NOT EXISTS public.user_tag
+(
+  id              BIGSERIAL PRIMARY KEY,
+  user_id         BIGINT                                                        NOT NULL REFERENCES public.user (id),
+  tag             user_tag_type                                                 NOT NULL,
+  -- tag_value is useful for certain tags like STORAGE_LIMIT e.g. tag="STORAGE_LIMIT", tag_value="1TB"
+  tag_value       TEXT                                                                  ,
+  inserted_at TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now())     NOT NULL,
+  deleted_at  TIMESTAMP WITH TIME ZONE
```
vasco-santos, line 44: What happens if tag deleted and then created for same user? In other words, should we have `updated_at`?

packages/db/postgres/tables.sql
```diff
@@ -25,6 +33,17 @@ CREATE TABLE IF NOT EXISTS public.user
   updated_at      TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL
 );
 
+CREATE TABLE IF NOT EXISTS public.user_tag
+(
+  id              BIGSERIAL PRIMARY KEY,
+  user_id         BIGINT                                                        NOT NULL REFERENCES public.user (id),
+  tag             user_tag_type                                                 NOT NULL,
+  -- tag_value is useful for certain tags like STORAGE_LIMIT e.g. tag="STORAGE_LIMIT", tag_value="1TB"
```
vasco-santos, line 41: I am a bit concerned here. Why a text with "1TB" string?

Two things here:
- text for being a generic value is ok, but I would prefer to see a type enum with that type the value is and a text value that you can parse as stated by type
- We stored user used storage as bytes number in the DB. We should follow same pattern here and use bytes number value.

With above, I would expect the admin dashboard to be able to receive and parse tag values such as 1TB, but convert it to bytes instead. Everything would be handled in bytes within API logic instead of conversations for every single validation of used storage

packages/db/postgres/tables.sql
flea89: I wonder if it's better to keep it this way to allow to have a "history".
So we would have a new row for every state change, what do you think?

vasco-santos: From https://www.notion.so/pl-strflt/Enforcing-user-storage-capacity-limits-a29aa59f28484842bea2721f5982234a#c7c52385776f49148c28d4fbdede6f1b we were already keeping a history table, that is basically interacted by admin dashboard. It also has "reason" for blocking, or unblocking. With that in mind, I think we should just have the latest in this table, which is the table that controls the product flows authorization. Any thing that I am missing?

vasco-santos: I think that is fine, have no strong preference.

But, thinking better about `updated_at` individually, we should still have it as `tag_value` can be updated. Or were you considering deleting and adding a new one?

trigramdev9: @vasco-santos i believe admin would delete then add a new row to preserve a history.  updated_at muddies history a bit.

packages/db/postgres/tables.sql
```diff
@@ -25,6 +33,17 @@ CREATE TABLE IF NOT EXISTS public.user
...
+  tag_value       TEXT                                                                  ,
```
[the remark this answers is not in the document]
trigramdev9: yea.  are you thinking null values would consume too much space in the db?  the alternative is a user_tag value table we could join on.

packages/db/postgres/tables.sql
trigramdev9: Given that `user_tag` is meant to be a very general table, it follows that potential values for tags may range from all sorts of different data types.  `PINNING` doesn't need a value, `STORAGE_LIMIT` uses bytes as you mentioned: but what about future tags?  To make the tag value `bytes` assumes that all future tags will have byte values.  Also thinking about this from an admin perspective, I wonder if a user will want to type `1000000000000` when they really mean `1TB`?

packages/db/postgres/tables.sql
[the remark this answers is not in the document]
adamalton: I think my answer above provides a suggested resolution to this question.

packages/db/postgres/tables.sql
jsdevel: @adamalton i see.  yea, i like your proposal and feel it will fit in and be generic enough for all use cases i can think of moving forward.  can you see any downsides to this approach?  i do wonder if being too general here is a limiting factor, but for now it seems good.  not sure how i missed the proposal from @vasco-santos to have another column to denote the type but that makes perfect sense to me.

packages/db/postgres/tables.sql
trigramdev9: actually on further thought i'm gonna add `updated_at`.

packages/db/postgres/tables.sql
adamalton: > can you see any downsides to this approach?
@jsdevel I think that this approach gives us the best balance that we're going to get between flexible/generic and not too crazy. The only thing that ran through my mind was that casting values from text to their declared type (e.g. `parseInt(text_value)`) feels like it's veering a little bit towards `eval(untrusted_value_from_db)`! But I think it's safely a good distance away from that yet!

packages/db/postgres/tables.sql
adamalton: @trigramdev9 what's your thinking behind adding `updated_at`? I thought that the `inserted_at` and `deleted_at` combo to give us history was quite neat. Interested to know what's swayed you the other way.

packages/db/postgres/tables.sql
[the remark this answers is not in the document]
vasco-santos: not concerned on space, just on consumer expectations to be different for each type of tag. Which would leave us with hardcoded logic for each context. That's why I suggested the type column, which seems we are aligned :)

packages/db/postgres/tables.sql
vasco-santos: if we want to go with delete and create new one, sounds good to not have `updated_ at`

packages/db/postgres/tables.sql
vasco-santos: Thanks @adamalton to put that code snippet together! That was exactly my thoughts

PR #969, opened 2022-02-10
feature/psa-metrics

alexandrastoica:
Adds metrics:
- upload by type (car, blob, multipart, upload)
- pin requests count
- refactors pins statuses listing


LeslieOA (approved):
LGTM. I know at some point we'll have to consider caching of metrics as proposed on NFT.storage

packages/db/postgres/functions.sql
```diff
@@ -325,6 +325,19 @@ BEGIN
 END
 $$;
 
+CREATE OR REPLACE FUNCTION uploads_by_type(query_type TEXT) RETURNS TEXT
```
flea89, line 328: I'm wondering why do we need a procedure here? Shouldn't we just use client count API?
I see that's how it is done for other metrics, but I'm not sure if/why it's needed?

I wonder if was done like this because of the Supabase client issue loading the all table even when just "counting"? 
But that can be workarounded by adding `range(0, 1)`

Probably one for @vasco-santos

vasco-santos: For `sum` we can't do with the client, and that was the main motivation to not use the client. The `getPinStatusMetrics`  using `pin_from_status_total` I don't recall if there was a reason. Either the cast to pin status, or was just added here with the `sums` and we didn't consider using the client. 

We use the client for other metrics counts, so if we can use it here, I think we should

vasco-santos: @alexandrastoica did you find a reason to not have it using supabase client?

packages/api/src/metrics.js
```diff
@@ -54,22 +66,14 @@ export async function metricsGet (request, env, ctx) {
...
     '# TYPE web3storage_pins_total counter',
     `web3storage_pins_total ${pinsTotal}`,
+    ...pinsMetrics.map(
+      ({ status, total }) =>
+        `web3storage_pins_total{status="${status}"} ${total || 0}`
+    ),
 
-    '# HELP web3storage_pins_status_queued_total Total number of pins that are queued.',
-    '# TYPE web3storage_pins_status_queued_total counter',
-    `web3storage_pins_status_queued_total ${pinsQueuedTotal}`,
```
vasco-santos, line 60: I would like to change this. But we need to take into account older values. Should we rename things here to Prometheus best practises and do a sum in Grafana query with the older name?

cc @hugomrdias @alanshaw

packages/db/postgres/functions.sql
alexandrastoica: Will change to client! :)

packages/api/src/metrics.js
alanshaw: yes!

packages/db/metrics.js
```diff
@@ -30,6 +30,30 @@ export async function getUploadMetrics (client) {
   }
 }
 
+const uploadTypeMapping = {
+  uploads_car_total: 'Car',
+  uploads_blob_total: 'Blob',
+  uploads_multipart_total: 'Multipart',
+  uploads_upload_total: 'Upload'
```
alanshaw, line 37: The DB client shouldn't be concerned with the keys we're using in the metrics. An actual upload "type" should be passed to `getUploadTypeMetrics`.

alanshaw (changes requested):

vasco-santos (approved):
LGTM 👌🏼 

@alexandrastoica please make sure that when this PR is merged to leave a comment in the Release PR that when the PR is releases we need to update Grafana Dashboard linking to this PR  (a Release PR for the API will be created or updated)

Thanks :)

PR #973, opened 2022-02-11
chore: use miniflare for local dev

olizilla:
- add a .env.local file to configure miniflare for local dev.
- update README with a much simpler "Getting Started" path.

Having just waded through the old getting started process again, this is *so much better* and I love it.

License: (Apache-2.0 AND MIT)

### DIFF
--- a/packages/api/README.md
+++ b/packages/api/README.md
@@ -4,7 +4,33 @@ The HTTP interface implemented as a Cloudflare Worker
 
 ## Getting started
 
-One time set up of your cloudflare worker subdomain for dev:
+We use miniflare to run the api locally, and docker to run ipfs-cluster and postgres with postREST.
+
+```sh
+# Copy <.env.local.tpl> to `.env.local` and fill out the missing worker vars.
+cp .env.local.tpl .env.local
+
+# Install the deps
+npm install
+```
+
+With docker running locally, start all the things:
+
+```
+npm start
+```
+
+🎉 miniflare is running in watch mode; you can save changes to the api code and the worker will update.
+
+Kill the process to stop miniflare, then run `npm run stop` to shutdown the cluster and postgres
+
+```sh
+npm run stop
+```
+
+## Setting up a cloudflare worker
+
+One time set up of your cloudflare worker subdomain. You only need to do this if you want to test in a real cloudflare worker.
 
 - `npm install` - Install the project dependencies
 - Sign up to Cloudflare and log in with your default browser.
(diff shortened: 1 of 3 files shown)


olizilla:
it blend!

```sh
❯ npm start                                                                                                                                                               16:34:42

> @web3-storage/api@5.1.3 start
> npm-run-all -p cluster:start pg:start -s miniflare


> @web3-storage/api@5.1.3 pg:start
> cd ../db && npm start


> @web3-storage/api@5.1.3 cluster:start
> web3-storage-tools cluster --start


> @web3-storage/db@4.0.4 start
> node scripts/cli.js db --start --project web3-storage


> @web3-storage/api@5.1.3 miniflare
> miniflare --env .env.local --watch


> @web3-storage/api@5.1.3 build
> node scripts/cli.js build

[mf:inf] Worker reloaded! (2.10MiB)
[mf:wrn] Worker's uncompressed size exceeds the 1MiB limit! Note that your worker will be compressed during upload so you may still be able to deploy it.
[mf:inf] Listening on :8787
[mf:inf] - http://127.0.0.1:8787
[mf:inf] - http://192.168.1.31:8787
running without backups wired up
POST /car 200 OK (187.29ms)
```

packages/api/README.md
```diff
@@ -4,7 +4,33 @@ The HTTP interface implemented as a Cloudflare Worker
...
+
+🎉 miniflare is running in watch mode; you can save changes to the api code and the worker will update.
+
+Kill the process to stop miniflare, then run `npm run stop` to shutdown the cluster and postgres
+
+```sh
+npm run stop
+```
+
+## Setting up a cloudflare worker
```
flea89, line 31: I'm wondering if this section is still relevant?  Do we want to keep it here in case a developer needs/wants to test the code against a real worker?
If the answer is yes I think we should update this because it might be a bit confusing to a developer without the right context. Also, some of the things in there are not accurate anymore. ie We'd need to update [this](https://github.com/web3-storage/web3.storage/pull/973/files#diff-f3a9540595a860f97e334d7faaa974596dd800c0bb472c65427fddbde50d51b4L77) to launch wrangler dev (not in local mode) instead of miniflare.

If not we could probably get rid of most of this and delete related npm scripts (localtunnel ones)

flea89 (commented):
Thanks @olizilla, we've longed for this for some time!

I'll actually test this next week, but I left a question in the meantime.

hugomrdias (approved):
LGTM

packages/api/README.md
flea89: Ok sure.
I'll try to find some spare time to open a PR with some updates to the Readme as follow-up!

vasco-santos (approved):
🚀

PR #975, opened 2022-02-11
feat Button hover effects

svvimming:
Hover effects have been added for two buttons across the entire site (Marketing + App). Dark button types now have a blurred gradient outline on hover and animated gradient background on active. Pink-blue gradient buttons now have their gradients animated on hover.

### DIFF
--- a/packages/website/components/button/button.scss
+++ b/packages/website/components/button/button.scss
@@ -39,8 +39,8 @@
     }
   }
 
-  &.dark {
-    background-color: $ebony;
+  &.dark,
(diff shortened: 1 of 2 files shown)


orvn (approved):

orvn (changes requested):
Add gradient glow on hover to light buttons as well.

orvn (approved):
Great! There's a particular button type I missed (`.outline-dark`), but will update requirements for that and it can be a separate PR.

PR #977, opened 2022-02-14
fix: invalid CAR errors

vasco-santos:
This PR fixes invalid CAR errors being 500 Internal server error, which will cause alerts to be triggered to sentry for user input.


yusefnapora (approved):
lgtm 👍

PR #979, opened 2022-02-15
Addressing PR feedback for user_tag table to allow tag value type (see #955)

trigramdev9:
see #955
Cousin pr in nft.storage https://github.com/nftstorage/nft.storage/pull/1378

### DIFF
--- a/packages/db/postgres/tables.sql
+++ b/packages/db/postgres/tables.sql
@@ -12,8 +12,17 @@ CREATE TYPE auth_key_blocked_status_type AS ENUM
 -- in the application.
 CREATE TYPE user_tag_type AS ENUM
 (
-  'PINNING',
-  'STORAGE_LIMIT'
+  'PSA_ENABLED',
+  'STORAGE_LIMIT_BYTES'
+);
+
+CREATE TYPE user_tag_value_type AS ENUM
+(
+  'bigint',
+  'boolean',
+  'integer',
+  'real',
+  'text'
 );
 
 -- A user of web3.storage.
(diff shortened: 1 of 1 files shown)


packages/db/postgres/tables.sql
```diff
@@ -38,9 +47,10 @@ CREATE TABLE IF NOT EXISTS public.user_tag
   id              BIGSERIAL PRIMARY KEY,
   user_id         BIGINT                                                        NOT NULL REFERENCES public.user (id),
   tag             user_tag_type                                                 NOT NULL,
-  -- tag_value is useful for certain tags like STORAGE_LIMIT e.g. tag="STORAGE_LIMIT", tag_value="1TB"
-  tag_value       TEXT                                                                  ,
+  value           TEXT                                                          NOT NULL,
+  value_type      user_tag_value_type                                           NOT NULL,
   inserted_at TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now())     NOT NULL,
+  updated_at  TIMESTAMP WITH TIME ZONE                                                  ,
```
vasco-santos, line 53: Let's talk this sync, but for arguments in the other PR, probably you were right in not having it and relying on delete + create again.

vasco-santos (commented):
Thanks for adding the type. Regarding `updated_ at`, let's sync and decide

LeslieOA (dismissed):
Could we also update `user_tag_type` with the following [as per **@adamalton**'s comment](https://github.com/web3-storage/web3.storage/pull/955#discussion_r806124199) please?

```sql
CREATE TYPE user_tag_type AS ENUM
(
  'PSA_ENABLED',
  'STORAGE_LIMIT_BYTES'
);
```

trigramdev9:
> Could we also update `user_tag_type` with the following [as per **@adamalton**'s comment](https://github.com/web3-storage/web3.storage/pull/955#discussion_r806124199) please?
> 
> ```sql
> CREATE TYPE user_tag_type AS ENUM
> (
>   'PSA_ENABLED',
>   'STORAGE_LIMIT_BYTES'
> );
> ```

I felt it was redundant to include the value type in the name of the type here as we're now specifying what type it is in the columns.

LeslieOA:
@trigramdev9 Good point (regarding bytes). Updated.
`PSA_ENABLED` all good?

vasco-santos (approved):
LGTM

PR #984, opened 2022-02-15
Adding ACCOUNT_ENABLED user_tag_type to support blocking/unblocking a user

trigramdev9:
* Also adding a reason field on the user_tag table.


vasco-santos (changes requested):
Please see https://github.com/nftstorage/nft.storage/pull/1384#pullrequestreview-883317399

vasco-santos (approved):
LGTM

Issue #923, opened 2022-01-28
Unskip CORS test and remove dependency on ipfs.io gateway in tests


alanshaw:
A CORS test that uses the ipfs.io gatewayhas been skipped. This needs to be re-enabled and at the same time we should remove that dependency by using a gateway from an IPFS node we're already running in docker for the tests.

See: https://github.com/web3-storage/web3.storage/pull/902#discussion_r794398150

PR #992, opened 2022-02-16
Move GATEWAY to env and expose ipfs gateway

flea89:
[Issue](https://github.com/web3-storage/web3.storage/issues/923) 
Moving GATEWAY to environment to allow to point to dockerized ipfs gateway in development.

### DIFF
--- a/packages/api/src/car.js
+++ b/packages/api/src/car.js
@@ -9,7 +9,7 @@ import * as cbor from '@ipld/dag-cbor'
 import * as pb from '@ipld/dag-pb'
 import retry from 'p-retry'
 import { InvalidCarError } from './errors.js'
-import { GATEWAY, LOCAL_ADD_THRESHOLD, MAX_BLOCK_SIZE } from './constants.js'
+import { LOCAL_ADD_THRESHOLD, MAX_BLOCK_SIZE } from './constants.js'
 import { JSONResponse } from './utils/json-response.js'
 import { getPins, PIN_OK_STATUS, waitAndUpdateOkPins } from './utils/pin.js'
 import { normalizeCid } from './utils/cid.js'
@@ -64,8 +64,8 @@ export async function carGet (request, env, ctx) {
   } = request
   // gateway does not support `carversion` yet.
   // using it now means we can skip the cache if it is supported in the future
-  const url = new URL(`/api/v0/dag/export?arg=${cid}&carversion=1`, GATEWAY)
-  res = await fetch(url, { method: 'POST' })
+  const url = new URL(`/api/v0/dag/export?arg=${cid}&carversion=1`, env.GATEWAY)
+  res = await fetch(url.toString(), { method: 'POST' })
   if (!res.ok) {
     // bail early. dont cache errors.
     return res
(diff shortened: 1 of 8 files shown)


packages/api/wrangler.toml
```diff
@@ -43,7 +43,7 @@ bindings = [{ name = "NAME_ROOM", class_name = "NameRoom0" }]
 account_id = "fffa4b4363a7e5250af8357087263b3a" # Protocol Labs CF account
 zone_id = "7eee3323c1b35b6650568604c65f441e"    # web3.storage zone
 route = "https://api.web3.storage/*"
-vars = { CLUSTER_API_URL = "https://web3.storage.ipfscluster.io/api/", ENV = "production", PG_REST_URL = "https://web3-storage-pgrest-prod.herokuapp.com" }
+vars = { CLUSTER_API_URL = "https://web3.storage.ipfscluster.io/api/", ENV = "production", PG_REST_URL = "https://web3-storage-pgrest-prod.herokuapp.com", GATEWAY_URL = "https://ipfs.io" }
```
vasco-santos, line 46: Can we add this to staging and other envs?

packages/api/src/car.js
```diff
@@ -64,8 +64,8 @@ export async function carGet (request, env, ctx) {
   } = request
   // gateway does not support `carversion` yet.
   // using it now means we can skip the cache if it is supported in the future
-  const url = new URL(`/api/v0/dag/export?arg=${cid}&carversion=1`, GATEWAY)
-  res = await fetch(url, { method: 'POST' })
+  const url = new URL(`/api/v0/dag/export?arg=${cid}&carversion=1`, env.GATEWAY_URL)
+  res = await fetch(url.toString(), { method: 'POST' })
```
vasco-santos, line 68: Do we need this change?

vasco-santos (commented):
thanks for getting this

packages/api/src/car.js
flea89: VSCode intellisense was screaming at me the first param type was wrong. 
According to `lib.dom.d.ts` used by VSCide that should be of `type RequestInfo = Request | string;`
But looking at the MDN doc the first param can be 

- A string or any other object with a [stringifier](https://developer.mozilla.org/en-US/docs/Glossary/Stringifier) — including a [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) object — that provides the URL of the resource you want to fetch.
- A [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) object.

so the answer is no, it's not.
Happy to revert 👍

packages/api/wrangler.toml
flea89: Good catch!
We definitively need staging to point to [https://ipfs.io](https://ipfs.io)

For other cloudfare dev envs I would suggest we keep the [https://ipfs.io](https://ipfs.io) (rather than the local version) to avoid having to create another tunnel, would you agree?

Honestly, I expect now that we have miniflare, that devs would use that for development rather than this (in most cases) so it shouldn't be a big deal, would you agree?

vasco-santos: Yes, let's point to ipfs.io

packages/api/src/car.js
vasco-santos: ok, let's keep. Just wanted to know why

packages/api/wrangler.toml
flea89: Cool, ✅

vasco-santos (approved):
LGTM! 👌🏼

PR #1000, opened 2022-02-17
feat: #954 add custom max chunk size to client put and putCar options

### LINKED ISSUE #954: Can I change the default chuck size?
Default chunk size is about 10MB, So the progress won't change for a long time,  is there any way I can change the chunk size so I can let the progress change more frequently. thanks

akazwz:
### Changes

-   add custom chunk size to client put and putCar options: fixes #954


packages/client/src/lib.js
```diff
@@ -109,12 +110,12 @@ class Web3Storage {
         })),
         blockstore,
         wrapWithDirectory,
-        maxChunkSize: 1048576,
+        maxChunkSize: maxChunkSize,
```
alanshaw, line 113:
```suggestion
        maxChunkSize,
```

packages/client/src/lib/interface.ts
```diff
@@ -101,6 +101,10 @@ export type PutOptions = {
    * Maximum times to retry a failed upload. Default: 5
    */
   maxRetries?: number
+  /**
+   * Maximum chunk size to upload. Default: 1024 * 1024 * 10 (in bytes)
```
alanshaw, line 105:
```suggestion
   * Maximum chunk size to upload in bytes. Default: 10,485,760
```

packages/client/src/lib/interface.ts
```diff
@@ -138,6 +142,10 @@ export type PutCarOptions = {
    * Maximum times to retry a failed upload. Default: 5
    */
    maxRetries?: number
+  /**
+   * Maximum chunk size to upload. AS the targetSize as passed to carbites TreewalkSplitter. Default: 1024 * 1024 * 10 (in bytes).
```
alanshaw, line 146:
```suggestion
   * Maximum chunk size to upload in bytes. Default: 10,485,760
```

alanshaw (approved):

alanshaw:
@vasco-santos @olizilla are you ok with this addition?

vasco-santos:
> @vasco-santos @olizilla are you ok with this addition?

We should probably link to recommended values, or have a ceiling. Is there any benchmarks with different chunk sizes you are aware of?

packages/client/src/lib.js
```diff
@@ -109,12 +110,12 @@ class Web3Storage {
         })),
         blockstore,
         wrapWithDirectory,
-        maxChunkSize: 1048576,
+        maxChunkSize,
```
alanshaw, line 113: Hold up, this shouldn't have been changed - this is the max block size (I know confusingly named) which we set to 1MB.

```suggestion
        maxChunkSize: 1048576,
```

Although we should really pull this out into a constant and use it here, so we don't get confused again!:

```js
const MAX_BLOCK_SIZE = 1048576
```

alanshaw (changes requested):
@akazwz there's one fix that I've highlighted. The addition needed here is to add some validation on the `maxChunkSize` option to make sure it is less than 100MiB and greater than or equal to 1MB (`MAX_BLOCK_SIZE`).

packages/client/src/lib.js
akazwz: @alanshaw Thanks for the suggestions, I already did some validation on the `maxChunkSize `and added the MAX_BLOCK_SIZE constant, have a look when you have time

packages/client/src/lib.js
```diff
@@ -97,9 +98,13 @@ class Web3Storage {
     onRootCidReady,
     onStoredChunk,
     maxRetries = MAX_PUT_RETRIES,
+    maxChunkSize = MAX_CHUNK_SIZE,
     wrapWithDirectory = true,
     name
   } = {}) {
+    if (maxChunkSize >= 104857600 || maxChunkSize < 1048576) {
```
vasco-santos, line 105: @akazwz now that we have constant, we should use them in this conditional and other below. Also, we should have a constant for the 100MB upper limit

vasco-santos (changes requested):
Thanks for the PR, left a comment about using the constants everywhere instead of still hardcoding the numbers

alanshaw (approved):

vasco-santos (approved):

vasco-santos (changes requested):
@akazwz can you please add a test to get the changes validated and code coverage CI happy?

alanshaw:
Ah we need tests for the coverage to pass!

akazwz:
@alanshaw @vasco-santos Already added tests for custom chunk size, but I am not familiar with the tests, can you please check if it works or give some suggestions.

packages/client/test/put.spec.js
```diff
@@ -169,6 +207,17 @@ describe('putCar', () => {
     }
   })
 
+  it('errors for CAR with wrong max chunk size', async () => {
+    const client = new Web3Storage({ token, endpoint })
+    const carReader = await createCar('hello world')
+    try {
+      await client.putCar(carReader, {maxChunkSize: 10})
```
vasco-santos, line 214:
```suggestion
      await client.putCar(carReader, { maxChunkSize: 10 })
```

vasco-santos (changes requested):
Thanks a lot @akazwz this should be the last change needed

akazwz:
@vasco-santos Already done, thanks for the suggestions

vasco-santos (approved):

PR #1004, opened 2022-02-17
Renaming user blocked and psa user_tags (closes #1002)

### LINKED ISSUE #1002: Change ACCOUNT_ENABLED user tag to ACCOUNT_RESTRICTED
This is a proposal to change the `ACCOUNT_ENABLED` user tag that was added in #984 to be called `ACCOUNT_RESTRICTED` instead (and to invert its behaviour).

This came up when we (Potato) were discussing the implementation of #988, which in brief, will allow the disabling of some (but not all) of a user's account functionality.

The reason for changing `ACCOUNT_ENABLED` to `ACCOUNT_RESTRICTED` is two-fold:

1. The requirements of #988 are to restrict _some_, but not all, of a user's account functionality, so rather than being a binary on/off enabled/disabled thing, it's actually just putting a set of restrictions on the account. So "restricted" seems to make more sense than "enabled/disabled".
2. You would usually expect a user's account to be enabled/unrestricted by default (i.e. if a user creates a new account, you would expect its initial state to be unrestricted, rather than starting as restricted and requiring a flag to derestrict it). So it seems to make more sense that the tag in the `user_tags` table is for the _restriction_ rather than the lack of restriction.
    - Obviously we can choose what the behaviour is when this tag doesn't exist for a user, so it could be done either way round, but it just feels a bit more intuitive to do it this way round.

@vasco-santos @jsdevel @JeffLowe any objections to us making this change?

trigramdev9:
nft.storage PR [here](https://github.com/nftstorage/nft.storage/pull/1397)

### DIFF
--- a/packages/db/postgres/tables.sql
+++ b/packages/db/postgres/tables.sql
@@ -12,18 +12,9 @@ CREATE TYPE auth_key_blocked_status_type AS ENUM
 -- in the application.
 CREATE TYPE user_tag_type AS ENUM
 (
-  'ACCOUNT_ENABLED',
-  'PSA_ENABLED',
-  'STORAGE_LIMIT_BYTES'
-);
-
-CREATE TYPE user_tag_value_type AS ENUM
-(
-  'bigint',
-  'boolean',
-  'integer',
-  'real',
-  'text'
+  'HasAccountRestriction',
+  'HasPsaAccess',
+  'StorageLimitBytes'
 );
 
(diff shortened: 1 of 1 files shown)


alanshaw (approved):
The renaming LGTM...however the other enums in the system do not follow this casing convention. I suggest we change for consistency:

```
HasAccountRestriction
HasPsaAccess
StorageLimitBytes
```

PR #1008, opened 2022-02-18
feat: use user tags to check PSA auth

alexandrastoica:
Removes the old `pinning_authorization` table and uses user tags instead


packages/api/test/fixtures/init-data.sql
```diff
@@ -81,11 +81,10 @@ VALUES (
   1669394359626000000
 );
 
--- user 'test-pinning' is authorized
-INSERT INTO pinning_authorization (user_id)
-VALUES (4);
-INSERT INTO pinning_authorization (user_id)
-VALUES (5);
+-- -- user 'test-pinning' is authorized
```
flea89, line 84: nit: double comment -- --

packages/api/test/fixtures/init-data.sql
```diff
@@ -81,11 +81,10 @@ VALUES (
...
+INSERT INTO public.user_tag (user_id, tag, value, value_type, reason)
+VALUES  (4, 'PSA_ENABLED', true, 'boolean', 'test'),
+        (5, 'PSA_ENABLED', true, 'boolean', 'test');
```
trigramdev9, line 87: please pull the latest.  `value_type` has been removed.

packages/db/postgres/pg-rest-api-types.d.ts
```diff
@@ -1906,6 +2068,24 @@ export interface definitions {
...
+     * Note:
+     * This is a Primary Key.<pk/>
+     */
+    id: number;
+    /**
+     * Note:
+     * This is a Foreign Key to `user.id`.<fk table='user' column='id'/>
+     */
+    user_id: number;
+    tag: "ACCOUNT_ENABLED" | "PSA_ENABLED" | "STORAGE_LIMIT_BYTES";
```
trigramdev9, line 2082: please pull latest.  these have been renamed.

packages/db/index.js
```diff
@@ -138,16 +138,18 @@ export class DBClient {
...
    */
   async isPinningAuthorized (userId) {
-    const { error, count } = await this._client
-      .from('pinning_authorization')
-      .select('id', { count: 'exact' })
+    const { count, error } = await this._client
+      .from('user_tag')
+      .select('value', { count: 'exact' })
       .eq('user_id', userId)
+      .eq('tag', 'PSA_ENABLED')
```
trigramdev9, line 151: please pull latest.  this has been renamed to `HasPsaAccess`

trigramdev9 (changes requested):

alexandrastoica:
SQL for migrating the existing data from `pinning_authorization` to `user_tag`

```
BEGIN TRANSACTION;
INSERT INTO public.user_tag
(
    user_id,
    tag,
    value,
    reason,
	inserted_at
)
SELECT
     user_id,
     'HasPsaAccess' as tag,
     'true' as value,
     'Approved access' as reason,
	 inserted_at
 FROM public.pinning_authorization
 WHERE deleted_at IS NULL;
 
INSERT INTO public.user_tag
(
    user_id,
    tag,
    value,
    reason,
	inserted_at,
	deleted_at
)
SELECT
     user_id,
     'HasPsaAccess' as tag,
     'false' as value,
     'Revoked access' as reason,
	 inserted_at,
	 deleted_at
 FROM public.pinning_authorization
 WHERE deleted_at IS NOT NULL;

DROP TABLE public.pinning_authorization;
COMMIT;
```

alanshaw:
@alexandrastoica could you add the migration SQL to here please? https://github.com/web3-storage/web3.storage/tree/main/packages/db/postgres/migrations

flea89:
@alanshaw thanks for pointing that out! We noticed this morning that folder was created 🎉 

While I understand this is not the full, polished solution for migrations, (since we're still missing the automation side of it) I do believe moving migrations scripts to version control is a great improvement to what we're doing now. 
So big plus and thanks 🙏  ! 

I'm wondering if we should be adding some documentation to make sure developers know what's expected when altering schemas?

Do we have an issue to track the more holistic DB migration epic/approach?

alexandrastoica:
@alanshaw added the migration file. Let me know if this is ok to merge. Thanks!

alanshaw (approved):
LGTM

vasco-santos:
Runing SQL migration for deploy got:

```
web3-storage-prod::HEROKU_POSTGRESQL_AQUA=> INSERT INTO public.user_tag
(
    user_id,
    tag,
    value,
    reason,
    inserted_at
)
SELECT
    user_id,
    'HasPsaAccess' as tag,
    'true' as value,
    'Approved access' as reason,
    inserted_at
 FROM public.pinning_authorization
 WHERE deleted_at IS NULL;
ERROR:  invalid input value for enum user_tag_type: "HasPsaAccess"
LINE 11:     'HasPsaAccess' as tag,
```
Looking into it

flea89:
@vasco-santos the user_tag enum was created as part of a separate PR from trigram folks, it's possible they haven't created a migration to create the ENUM?
```
-- User tags are associated to a user for the purpose of granting/restricting them
-- in the application.
CREATE TYPE user_tag_type AS ENUM
(
  'HasAccountRestriction',
  'HasPsaAccess',
  'StorageLimitBytes'
);
```

trigramdev9:
> @flea89 thanks, you were right! The type was outdated and it was merged without updating the type. We need to improve communication on DB changes

@vasco-santos i'll take ownership for this in this instance.  if you recall the conversations we were having a few weeks ago about naming...  i meant to follow up once we were all in agreement and the conversations had died down to avoid unnecessary DB updates in prod, but i forgot to do so.

Issue #754, opened 2021-12-06
Give user information on whether all CAR chunks got uploaded for larger uploads


dchoi27:
Sister issue to: https://github.com/nftstorage/nft.storage/issues/896

In addition, #689 now shows everything in the uploads table as `Available`. This isn't actually true though - if there was a multipart upload for a file larger than 100MB, but not every chunk got uploaded (e.g. connection got interrupted), then the upload still shows up in the table even if not everything is retrievable in that situation. So we'll want to modify this as well.

dchoi27:
Let's check the error message when not all CAR chunks get uploaded (not 100% sure if it's been implemented in both sides).

Otherwise brainstorming for this is here: https://hackmd.io/E0jxc0pCSxSkYgSUUq9lJA?view

dchoi27:
Just a heads up - there's a website component to this that we'll probably only want to incorporate after website redesign launch. When we're creating tickets for this epic, let's label the front-end one with the topic/frontend label so we're reminded to tackle it after the redesign is launched. cc @orvn 

Also - when the CAR streaming work lands in NFT.Storage and we've ported it to Web3.Storage, figuring out what CARs are incomplete will get a lot easier for uploads moving forward

orvn:
Is the goal here for the web UI to display how many chunks are uploaded and have a status `complete` visually, in the app UI? @dchoi27 

Btw, the `availability` column was added in the new UI, in parallel with https://github.com/web3-storage/web3.storage/pull/729

![availibility-web3s-files-table](https://user-images.githubusercontent.com/3689720/152891242-6b621de5-2703-4f87-a684-e3c6942c4997.png)

dchoi27:
Cool! yeah - potentially more a warning if we don't think we have all the content (so doesn't have to be its own content, can just be something visually), but depends on the solution that potato decides on

mikeal:
Why don’t we verify the upload is complete by asking for the last block in the last CAR file from a gateway after the upload?

flea89:
**Hypothesis**
Read this in the context of the quick fix (to avoid the frustration of users thinking something is available while it is not) rather than the re-architecture of how uploads work.
I agree while that's required it definitively requires some more thoughts and changes to how control flow and data structures in the backend and defo has a dependency on the unix-fs-importer work.

> Why don’t we verify the upload is complete by asking for the last block in the last CAR file from an IPFS Gateway?

[...]

PR #1009, opened 2022-02-18
feat: remove availabilty and change status to be more general.

flea89:
With this PR, I'd like to open up a discussion around how we display statuses and availability to users.
In this proposed solution we would slightly change how we communicate the status back to the user, hopefully, this approach would simplify and be less confusing to users.

The assumption here is that [https://web3.storage/files/](https://web3.storage/files/) should fulfill the following user outcomes when it communicating the status of an upload:

> As a user I want to know what's the status of my upload (rather than knowing what is the status of the cluster) so I reliably know if it was successful or not and my content is retrievable through the gateway, APIs, cli etc

Users don't really care about Web3.storage infra, if it's a IPFS cluster or something different to get hold of their uploads ( in fact we know it's likely it won't be a cluster in the future), so giving that level of detail doesn't add value to the users, in fact it might even confuse them.

At the moment, before changing how uploads work, the idea was to show the `Availability` by relying on status === Pinned (discussed [here](https://github.com/web3-storage/web3.storage/issues/754#issuecomment-1040184161)) but practically that means we're displaying the "same" information twice in the table. I wondering if that would just confuse the user even more.

## Solution
- Remove Availability col
- Update `Pin Status` header to `Status`
- Communicate to user failed uploads could stay in `Queuing` state indeterminately. 
- Update statuses  to be either `Pinned` or `Queuing` (will have a `Failed` in the future as well)
- Update tooltips and wording accordingly

## Todo
- The status is currently updated asynchronously, is that an acceptable compromise for now or do we need to look into ways to provide up to date statuses to users
- As I mentioned above we now just have  `Pinned` and `Queuing`, I'm wondering if we should make them, even more, cluster/ipfs agnostic.

### DIFF
--- a/packages/website/pages/files.js
+++ b/packages/website/pages/files.js
@@ -44,8 +44,7 @@ const QuestionMark = () => (
 )
 
 const TOOLTIPS = {
-  PIN_STATUS: (<span>Reports the status of a file or piece of data stored on Web3.Storage’s IPFS Cluster. Status might not be fully up-to-date. Data is still available even when still in Queuing state.</span>),
-  AVAILABILITY: (<span>Reports the status of a file or piece of data stored on Web3.Storage’s IPFS nodes.</span>),
+  UPLOAD_STATUS: (<span>Reports the status of a file or piece of data stored on Web3.Storage. Status might not be fully up-to-date.</span>),
 
   CID: (<span>
     The <strong>c</strong>ontent <strong>id</strong>entifier for a file or a piece of data.<span> </span>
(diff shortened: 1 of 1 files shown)


dchoi27:
Left a comment on the issue: https://github.com/web3-storage/web3.storage/issues/754#issuecomment-1047277778

Basically - think this PR in full is OK if we think the Status column will have things move from Queued to Pinned faster than things were moved before the Availability column was introduced, but if not, then we just get into the situation we used to have where users only saw Queued all the time and had a ton of questions about it (even though their data was effectively on 1+ nodes and available).

alanshaw (approved):

PR #1010, opened 2022-02-18
feat: Outline button hover effect

svvimming:
Hover/active effects with animated gradient backgrounds for buttons with light/dark outline.

### DIFF
--- a/packages/website/components/button/button.scss
+++ b/packages/website/components/button/button.scss
@@ -244,6 +244,14 @@
         opacity: 1;
       }
     }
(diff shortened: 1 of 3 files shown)


orvn (changes requested):
@svvimming there's an issue with the 3rd outline button on the _Tiers_ page, where the bottom of the button is not clickable/hoverable due to an invisible element. Can you resolve that?

Also, is this styled for both dark and light outline buttons?

orvn (approved):

Issue #1016, opened 2022-02-21
/version endpoint not returning version info


olizilla:
Looks like the git commit info is not getting into the api build env anymore. the /version enpoint is not returning version info

```sh
❯ curl https://api.web3.storage/version -s |  jq
{
  "mode": "rw"
}
```

https://github.com/web3-storage/web3.storage/blob/6147c69c60b284ef9fcde4197cfbdc96b72e37e4/packages/api/src/version.js#L9-L17

Issue #983, opened 2022-02-15
Source maps in Sentry with ESM


vasco-santos:
I believe that with #902 ESM refactor our Sentry source maps were affected and now they have the build stack trace which is not friendly for debugging potential problems.

We need to get this fixed to help diagnosing alerts

olizilla:
The gory details on how we got it working last time https://github.com/web3-storage/web3.storage/pull/505

PR #1024, opened 2022-02-22
fix: version endpoint info

olizilla:
tweak the esbuild define config to set the git info in the api build.

The `define` api only replaces the exact token, where it finds it seperated by whitespace, with the exact value provided. So the key `VERSION` was not matching on `env.VERSION`, so our envAll function now sets these values on the env for us by pulling them from magic globals that esbuild will set.

I removed the `define` config for ENV as i'm pretty sure that is set at runtime by cloudflare/miniflare and we shouldn't be attempting to burn it into the worker bundle, but would like someone to sanity check me there.

```console
❯ curl "https://api.web3.storage/version"
{"mode":"rw"}

❯ curl "http://127.0.0.1:8787/version"
{"version":"web3-api@5.1.4-dev+9068490","commit":"9068490aaa4dbe037e8e6aa9eaad706a46f11aed","branch":"main","mode":"rw"}
```

see: https://esbuild.github.io/api/#define

fixes: #1016 

License: (Apache-2.0 AND MIT)

### DIFF
--- a/packages/api/scripts/cli.js
+++ b/packages/api/scripts/cli.js
@@ -23,6 +23,7 @@ prog
     // custom prefix here not pkg.name.
     // See https://docs.sentry.io/platforms/javascript/guides/cordova/configuration/releases/
     const sentryRelease = `web3-api@${pkg.version}-${opts.env}+${git.short(__dirname)}`
+    console.log(`Building ${sentryRelease}`)
 
     await build({
       entryPoints: [path.join(__dirname, '..', 'src', 'index.js')],
@@ -58,6 +59,7 @@ prog
 
     // Sentry release and sourcemap upload
     if (process.env.SENTRY_UPLOAD === 'true') {
+      console.log(`Uploading to Sentry ${sentryRelease}`)
       const cli = new Sentry(undefined, {
         authToken: process.env.SENTRY_TOKEN,
         org: 'protocol-labs-it',
--- a/packages/api/scripts/cli.js
+++ b/packages/api/scripts/cli.js
@@ -17,7 +17,7 @@ const prog = sade('api')
 prog
   .command('build')
   .describe('Build the worker.')
-  .option('--env', 'Environment', 'dev')
+  .option('--env', 'Environment', process.env.ENV)
   .action(async (opts) => {
     // Release name cannot contain slashes, and are global per org, so we use
     // custom prefix here not pkg.name.
--- a/packages/api/scripts/cli.js
+++ b/packages/api/scripts/cli.js
@@ -49,7 +49,6 @@ prog
         VERSION: JSON.stringify(version),
         COMMITHASH: JSON.stringify(git.long(__dirname)),
         BRANCH: JSON.stringify(git.branch(__dirname)),
-        ENV: opts.env || 'dev',
         global: 'globalThis'
       },
       minify: opts.env !== 'dev',
--- a/packages/api/scripts/cli.js
+++ b/packages/api/scripts/cli.js
@@ -22,7 +22,7 @@ prog
     // Release name cannot contain slashes, and are global per org, so we use
     // custom prefix here not pkg.name.
     // See https://docs.sentry.io/platforms/javascript/guides/cordova/configuration/releases/
-    const version = `web3-api@${pkg.version}-${opts.env}+${git.short(__dirname)}`
+    const sentryRelease = `web3-api@${pkg.version}-${opts.env}+${git.short(__dirname)}`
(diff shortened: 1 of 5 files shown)


packages/api/scripts/cli.js
```diff
@@ -49,7 +49,6 @@ prog
         VERSION: JSON.stringify(version),
         COMMITHASH: JSON.stringify(git.long(__dirname)),
         BRANCH: JSON.stringify(git.branch(__dirname)),
-        ENV: opts.env || 'dev',
```
vasco-santos, line 52: What was the reasoning to remove this? Can't we keep it?
We need this to distinguish the env where the error happens (prod or staging)

olizilla: this is the `define` config for esbuild. It tells it to replace strings in the built bundle... our "env" value should come from the env vars passed to miniflare or cloudflare, rather than being baked into the bundle.

olizilla:
yes, that looks better

```console
❯ curl "http://127.0.0.1:8787/version"
{"version":"5.1.4","commit":"73d2a7759d626fb1127e18f0730b929ca61cc66a","branch":"fix-api-version","mode":"rw"}
```

olizilla:
I'd like to tackle #983 as a seperate issue, as it may take a while to get sentry source maps working again.

alanshaw (approved):

packages/api/scripts/cli.js
vasco-santos: I thought this was needed in the bundle for https://docs.sentry.io/product/sentry-basics/environments/ but yeah, I am not sure on it

vasco-santos: Ok, I see it is likely the end line with `.newDeploy()`

vasco-santos (approved):

packages/api/scripts/cli.js
olizilla: yep, sentry can pick the env out of the version string, or we also pass it to the `newDeploy` call explicitly. Then at runtime, env.ENV should have been set in Cloudflare as a runtime env var, which Sentry sets on it's requests to their api when flagging errors, so reporting is logged under the correct build.

olizilla: hmm... what is not clear yet is how/if we are setting the env.ENV correctly at build time... I don't see any evidence of that in https://github.com/web3-storage/web3.storage/blob/main/.github/workflows/api.yml yet

vasco-santos: via wrangler.toml file?

olizilla: check this

```console
❯ wrangler build --env production
🌀  Running npm run build

> @web3-storage/api@5.1.4 build
> node scripts/cli.js build

{ 'opts.env': 'dev', 'process.env.ENV': undefined }

❯ wrangler build --env oli
🌀  Running npm run build

> @web3-storage/api@5.1.4 build
> node scripts/cli.js build

{ 'opts.env': 'dev', 'process.env.ENV': undefined }
```

olizilla: i think `opts.env` as passed to the build cli is always `dev` and that's why we're seeing all our releases listed as `dev` on sentry

Screenshot 2022-02-23 at 11 43 46

olizilla: the env gets passed in explicitly in nft.storage by calling a different build command per env

https://github.com/nftstorage/nft.storage/blob/c47f94fc267c80279791f82d60a5ee031847b87e/packages/api/wrangler.toml#L66-L68

packages/api/scripts/cli.js
```diff
@@ -17,12 +17,12 @@ const prog = sade('api')
 prog
   .command('build')
   .describe('Build the worker.')
-  .option('--env', 'Environment', 'dev')
+  .option('--env', 'Environment', process.env.ENV)
```
olizilla, line 20: process.env.ENV should be set, and the default here was masking a problem that the value was not being set in dev nor in CI at deploy. I will fix this more gracefully in a subsequent PR

vasco-santos:
> TODO: The repitition of the ENV value is annoying. we also have 3 seperate env contexts to reason about

Yeah, let's figure out better way as follow up. No need here, as this was already the case

olizilla:
It blend

```
curl "https://api-staging.web3.storage/version"
{"version":"5.2.0","commit":"049f9b823a0b6406a6cc8105ca65c459a7646d15","branch":"main","mode":"rw"}
```

PR #1026, opened 2022-02-22
fix: remove availability column

alanshaw:
It has traditionally taken a while to get an updated pin status for an upload. It's due to a number of reasons which we're working to resolve.

_Often times_ when an upload has a pin status of `PinQueued` it IS actually already available on IPFS. _However_, this is not ALWAYS the case. If a large upload comprising of multiple CARs does not complete fully then the status will remain as `PinQueued`/`PinError` and since we haven't received all the data is is NOT available on IPFS (and will never be unless the upload is retried).

IMHO we should NEVER say that something is available when there is a chance it is not. Lets temporarily remove this column until we can communicate availability beetter without misleading users.

### DIFF
--- a/packages/website/pages/files.js
+++ b/packages/website/pages/files.js
@@ -44,8 +44,7 @@ const QuestionMark = () => (
 )
 
 const TOOLTIPS = {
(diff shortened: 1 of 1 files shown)


vasco-santos (approved):

olizilla:
see https://github.com/web3-storage/web3.storage/pull/1009

alanshaw:
Closed in favour of https://github.com/web3-storage/web3.storage/pull/1009

PR #1033, opened 2022-02-23
chore: sentry sourcemaps for worker.mjs

olizilla:
- [x] rename output to be worker.mjs see: https://developers.cloudflare.com/workers/cli-wrangler/configuration#modules
- [x] add request info to sentry log.  see: https://github.com/robertcepa/toucan-js#equivalent-of-above-as-a-module-mjs
- [x] tell sentry to upload bundle with .mjs extention. it only uploads .js by default. https://docs.sentry.io/product/cli/releases/#sentry-cli-sourcemaps
- [x] strip the `sourceMappingURL` comment from the bundle by using `external` as esbuild sourcemap mode. This is one less thing to go wrong / confuse sentry.
- [x] strip `.` from the start of the `filename` on stackframes as reported by the worker environment, so we just get reports from '/worker.mjs` and sentry can find the bundle for it.

Screenshot shows we now get **stacktraces!** _and_ **Request info** like the url and the user agent!!11!

Screenshot 2022-02-25 at 22 01 21

fixes: https://github.com/web3-storage/web3.storage/issues/983

previous: https://github.com/web3-storage/web3.storage/pull/505

License: (Apache-2.0 AND MIT)

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -71,7 +71,7 @@
   },
   "bundlesize": [
     {
-      "path": "./dist/index.mjs",
+      "path": "./dist/worker.mjs",
       "maxSize": "1 MB"
     }
   ]
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "type": "module",
   "license": "(Apache-2.0 OR MIT)",
-  "main": "./dist/index.mjs",
+  "main": "./dist/worker.mjs",
   "scripts": {
     "lt": "npm-run-all -p lt:*",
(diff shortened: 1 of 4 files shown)


olizilla:
AHA. We're not uploading our bundle because it ends with .mjs... the `include` option here only looks for .map and .js files https://github.com/web3-storage/web3.storage/blob/b9bcc9779e78b13c5e0f502636c5e071605191b4/packages/api/scripts/cli.js#L76-L77 

https://docs.sentry.io/platforms/javascript/guides/capacitor/sourcemaps/uploading/cli/

olizilla:
More details here: https://github.com/robertcepa/toucan-js/issues/54#issuecomment-1054169127

vasco-santos (approved):
LGTM

Issue #990, opened 2022-02-16
Storage limit - Framework to get users and record emails sent


LeslieOA:
Overarching user needs:
* Users need to receive warning emails from about nearing (milestones) their storage quota or exceeding it. User gets an email notification from support@web3.storage when they’ve hit the following thresholds: 75%, 80%, 85%, 90%, with instructions on how to request more.
* Admins (support@web3.storage) need to receive an email when a user has gone over their allowed storage quota.

[Notion - Enforcing-user-storage-capacity-limits](https://www.notion.so/pl-strflt/Enforcing-user-storage-capacity-limits-a29aa59f28484842bea2721f5982234a)

We can compare a users default (1TiB) or increased (`user_tag` table, `tag` of `STORAGE_LIMIT_BYTES`)  storage quota to what their current used storage. 

## Tasks

1. DB function to retrieve all users with storage used with a percentage from/to range of their allocated storage quota
User details returned:

- Email
- First and last name
- Storage quota
- Storage used

2. DB function to save details of sent email

- User ID
- Email type
- Email message ID
- Sent at

2. Node function in the `cron` package to call the DB functions to get the users that need emailing and save details of email sent.

**NOTE** 
Sending the email with MailChimp in a follow-up ticket:
[Add MailChimp email service](https://app.zenhub.com/workspaces/web3storage-product-6106d25b58fccc00105aedcb/issues/web3-storage/web3.storage/1153)

---

#### Threshold milestones and email frequency (TBC):

* Weekly email: equal to or greater than 75% of storage quota used (but below 85%)
* Weekly email: equal to or greater than 85% of storage quota used (but below 95%)
* Daily email: equal to or greater than 90% of storage quota used

> Note: This quota includes both uploads via the web UI and pins via the PSA (Pinning Service API). 

A cron job every `{daily}` (at a minimum). 

MailChimp's API is the preferred method of email automation. 

#### Questions

Do we need a `storage_usage_exceeded` table (as a snapshot of users who are over their quota) or will querying on cron and upload/pin suffice?

CC @dchoi27 for email frequency.

dchoi27:
SGTM - I'd say crank it up to daily at 90% (and replace the last 2 bullets)

adamalton:
I'm re-opening this, because #1273 is marked as fixing this, so as that is not yet closed, it makes sense for this to still be open. (Although it's slightly confusing, as both #1156 and #1273 are both marked as fixing this!)

PR #1044, opened 2022-02-25
feat: add pinned storage to user account API

### LINKED ISSUE #1043: Add the user's pinned storage to the API
Amend the DB function `user_used_storage` to return both the uploaded and pinned storage used for a user.

Amend the API endpoint `/user/account` to return both uploaded and pinned storage quotas used.

GaryHomewood:
Resolves #1043 

Amend `/user/account` API endpoint to return both uploaded and pinned storage usage for the user.

DB function amended to include a separate query to get the aggregated size of pinned content. This query is potentially expensive, joining 4 tables. Some benchmarking done with indexes added to the `psa_pin_request` table, but should be further tested against large data volumes before deploy to production.


packages/db/index.d.ts
```diff
@@ -38,7 +38,7 @@ export class DBClient {
   getMetricsValue (key: string): Promise<{ total: number }>
   upsertUser (user: UpsertUserInput): Promise<UpsertUserOutput>
   getUser (issuer: string): Promise<UserOutput>
-  getUsedStorage (userId: number): Promise<number>
+  getUsedStorage (userId: number): Promise<Record<string, string>>
```
flea89, line 41: Given we know the exact keys of the object returned here, I wonder if we should be more explicit with the return type

packages/db/postgres/functions.sql
```diff
@@ -251,17 +251,37 @@ BEGIN
...
+  INTO used_storage.uploaded::TEXT
+  FROM upload u
+  JOIN content c ON c.cid = u.content_cid
+  WHERE u.user_id = query_user_id::BIGINT
+  AND u.deleted_at is null;
+
+  SELECT COALESCE((
+    SELECT SUM(c.dag_size)
+    FROM psa_pin_request pr
+    JOIN pin p ON p.content_cid = pr.content_cid
```
flea89, line 273: nit.
I know it doesn't change, and we could leave this as is... but the order of the joins here is bothering me.
To map the fk relations should be psa_pin_request JOIN content JOIN pin.

flea89 (changes requested):
Nice one!
Left some comments!

packages/db/postgres/migrations/001-user-used-storage.sql
```diff
@@ -0,0 +1,35 @@
+CREATE TYPE used_storage AS (uploaded TEXT, pinned TEXT);
+
+DROP FUNCTION user_used_storage(bigint);
```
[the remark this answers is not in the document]
GaryHomewood: We do for a migration because the return type of the function is being changed.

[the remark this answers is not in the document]
flea89: Thanks Gary for clarifying, it is also explained in the docs [here](https://www.postgresql.org/docs/9.1/sql-createfunction.html#:~:text=CREATE%20FUNCTION%20defines%20a%20new,created%20in%20the%20specified%20schema.)

packages/db/index.js
```diff
@@ -160,20 +160,20 @@ export class DBClient {
...
-   * Get used storage in bytes.
+   * Get used storage in bytes, both uploaded and pinned.
    *
    * @param {number} userId
-   * @returns {Promise<number>}
+   * @returns {Promise<import('./db-client-types').UsedStorage>}
    */
   async getUsedStorage (userId) {
-    /** @type {{ data: string, error: PostgrestError }} */
+    /** @type {{ data: import('./db-client-types').UsedStorage, error: PostgrestError }} */
```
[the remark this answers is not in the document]
GaryHomewood: Have made this function responsible for converting to number so this return type is now correct.

packages/db/index.d.ts
GaryHomewood: Good point. Have made the return type explicit.

packages/db/postgres/functions.sql
GaryHomewood: Changed to map the fk relationships.

packages/db/postgres/functions.sql
```diff
@@ -251,17 +251,37 @@ BEGIN
...
+  SELECT COALESCE(SUM(c.dag_size), 0)
+  INTO used_storage.uploaded::TEXT
+  FROM upload u
+  JOIN content c ON c.cid = u.content_cid
+  WHERE u.user_id = query_user_id::BIGINT
+  AND u.deleted_at is null;
+
+  SELECT COALESCE((
+    SELECT SUM(c.dag_size)
+    FROM psa_pin_request pr
```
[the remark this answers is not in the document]
GaryHomewood: Renamed for clarity.

packages/db/postgres/functions.sql
```diff
@@ -251,17 +251,37 @@ BEGIN
...
+    JOIN pin p ON p.content_cid = pr.content_cid
+    JOIN content c ON c.cid = pr.content_cid
+    JOIN auth_key a ON a.id = pr.auth_key_id
+    WHERE a.user_id = query_user_id::BIGINT
+    AND a.deleted_at is null
```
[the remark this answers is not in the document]
GaryHomewood: Ok have removed for now, for consistency.

packages/db/test/pinning.spec.js
```diff
@@ -153,6 +155,13 @@ describe('Pin Request', () => {
         .map((p) => p.status)
       assert.deepStrictEqual(statuses, [pins[0].status, pins[1].status])
     })
+
+    it('includes pinned size in used storage', async () => {
```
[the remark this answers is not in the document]
GaryHomewood: Have left the test for pinned storage, since it does seem worth checking that its incremented here when we are pinning content.

packages/db/postgres/functions.sql
```diff
@@ -251,17 +251,37 @@ BEGIN
...
-  )::TEXT;
+  SELECT COALESCE(SUM(c.dag_size), 0)
+  INTO used_storage.uploaded::TEXT
+  FROM upload u
+  JOIN content c ON c.cid = u.content_cid
+  WHERE u.user_id = query_user_id::BIGINT
+  AND u.deleted_at is null;
+
+  SELECT COALESCE((
+    SELECT SUM(c.dag_size)
```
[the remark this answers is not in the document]
GaryHomewood: Ok, added extra test in `pinning.spec.js` to ensure null treated as 0 for used storage pinned amount.

packages/db/index.d.ts
flea89: Thanks!

packages/db/postgres/functions.sql
flea89: Thanks and sorry for being annoying!

packages/db/postgres/functions.sql
```diff
@@ -251,17 +251,38 @@ BEGIN
...
+  AND u.deleted_at is null;
+
+  -- NOTE: auth_key.deleted_at is NOT included in the where clause, since the 
+  -- auth_key is only used to authorize a request.
+  SELECT COALESCE((
+    SELECT COALESCE(SUM(c.dag_size), 0)
+    FROM psa_pin_request psa_pr
+    JOIN content c ON c.cid = psa_pr.content_cid
+    JOIN pin p ON p.content_cid = psa_pr.content_cid
+    JOIN auth_key a ON a.id = psa_pr.auth_key_id
```
vasco-santos, line 277: This query is dangerous in terms of execution time needed considering we are using 4 tables.

We should add an index for `psa_pr.content_cid`. Also not entirely sure if we need an individual index for `p.content_cid`. Can you do some benchmarks on these?

vasco-santos (dismissed):
Good call on separating both used storage for uploads and pinning for the consumer to decide what to do with them 👍🏼

packages/db/index.js
```diff
@@ -160,20 +160,23 @@ export class DBClient {
...
+    /** @type {{ data: { uploaded: string, pinned: string }, error: PostgrestError }} */
     const { data, error } = await this._client.rpc('user_used_storage', { query_user_id: userId }).single()
 
     if (error) {
       throw new DBError(error)
     }
 
-    return data || 0 // No uploads for the user
+    return {
+      uploaded: parseTextToNumber(data.uploaded),
```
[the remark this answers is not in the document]
flea89: The `MAX_SAFE_INTEGER` is `9007199254740991`,  if I've done my conversion right that translates to  `7.999999999999999112 Pebibyte`, which for a single user should be safe enough? 
And even if not safe we're raising if that's the case.
I wonder if letting the client deal with it is just moving a problem that we'd have anyway?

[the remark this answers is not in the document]
vasco-santos: > which for a single user should be safe enough?

I think it should be enough yes, even though I feel it would be safer to simply rely on BigInt in client instead.

> I wonder if letting the client deal with it is just moving a problem that we'd have anyway?

The client would be able to use BigInt. Otherwise we would need to serialize from API to client.

What do you think?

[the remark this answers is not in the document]
flea89: I think both solutions have their merits and cons, and I'm not too opinionated to go one route or the other.

Here's main drivers for having the DB package accountable for the conversion (and possible error) has the merits of:
- db client packages don't need to be aware of the casting (and its problems). That means if we end up having to use that number in other contexts is less likely we end up with a "transparent" bugs.
- The same applies to APIs clients.

I'd like to avoid that, those are the kind of bugs that are really tricky to find.

packages/db/postgres/functions.sql
vasco-santos: So, I think we did not get super better results in `psa_pin_request.content_cid` index because we still have a low number of pin requests. 

@alanshaw @hugomrdias what do you think about these indexes?

olizilla:
> should be further tested against large data volumes before deploy to production.

Has this happened? Are you satisfied that this change will performant enough not to be a problem?

olizilla:
Please fix update the PR with conflict fixes

packages/db/index.js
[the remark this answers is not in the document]
olizilla: it's not a big deal for now. If we are lucky enough to get a user with 7 PiB, then we can review this. For now, let's do the most convenient thing for the calling code and give them numbers as numbers. If this turns out to be a problem, then we can send them as strings and let the rendering code parse it with BigInt, which is well supported now.

olizilla:
Also, we can roll it out and monitor it. If it starts to get slow we can refactor to store the usage stats in the db and update them during an upload.

packages/db/postgres/functions.sql
```diff
@@ -251,17 +251,38 @@ BEGIN
 END
 $$;
 
-CREATE OR REPLACE FUNCTION user_used_storage(query_user_id BIGINT) RETURNS TEXT
-  LANGUAGE plpgsql
+CREATE TYPE used_storage AS (uploaded TEXT, pinned TEXT);
+
+CREATE OR REPLACE FUNCTION user_used_storage(query_user_id BIGINT)
```
trigramdev9, line 256: what is the purpose of this table?  if an admin increases the user's storage limit, will entries in this table be out of date or invalid?

trigramdev9 (changes requested):

packages/db/postgres/functions.sql
trigramdev9: @the-simian had another valid use case: if an admin _decreases_ the user's limit.  are these scenarios handled?

GaryHomewood: This is an addition to the existing function `user_used_storage` which gets the current amount of storage used for a specified user, for display on the web3 users `account` page. The extra SQL statement gets the current amount of pinned storage used and returns that as well.

The amount of storage used is independent of the users storage limit. A related ticket #990 builds on this and checks the calculated storage amounts against the user's quota limits, to notify them when they are approaching their limit.

packages/db/postgres/migrations/001-user-used-storage.sql
```diff
@@ -0,0 +1,36 @@
...
+  SELECT COALESCE((
+    SELECT COALESCE(SUM(c.dag_size), 0)
+    FROM psa_pin_request psa_pr
+    JOIN content c ON c.cid = psa_pr.content_cid
+    JOIN pin p ON p.content_cid = psa_pr.content_cid
+    JOIN auth_key a ON a.id = psa_pr.auth_key_id
+    WHERE a.user_id = query_user_id::BIGINT
+    AND psa_pr.deleted_at is null
+    AND p.status = 'Pinned'
+    GROUP BY p.status
```
olizilla, line 30: We need to pick out distinct CIDs here before aggregating the `dag_size`. We only want to sum the `dag_size` for unique root CIDs stored for a given user, rather than every redundant copy of that data (pin) that we store.

In prod, there will generally be 3 or more pin rows for a given `content_cid`; one for each cluster node the data is duplicated to. 

I've see 15(!?) rows on `bafybeidluj5ub7okodgg5v6l4x3nytpivvcouuxgzuioa6vodg3xt2uqle` that has dag_size: `140147` for my user id.
Screenshot 2022-03-24 at 14 21 20

Add the sum on dag_size is coming back as `2102205` rather than `140147`

Screenshot 2022-03-24 at 14 21 49

I need to look in to why we have 15 rows and not ~3 for that CID, but that's a seperate issue.

olizilla (changes requested):

packages/db/postgres/functions.sql
trigramdev9: i feel dumb.  not sure how i missed that this was a function, not a table.  thank you!

packages/db/postgres/migrations/001-user-used-storage.sql
GaryHomewood: Corrected to aggregate the size for distinct CIDs:

Screenshot 2022-03-25 at 17 08 16

packages/db/postgres/functions.sql
olizilla: @vasco-santos I'm happy to get this PR shipped without the indexes and monitor it. We can follow up with a PR to add indexes on all the foreign key source columns that are used in joins.

olizilla (approved):
Nice! Good work on adding the extra test to check that it's summing the dag_size for unique root CIDs.

one nitpick, this changeset includes `packages/website/package-lock.json` which shouldn't be included in this PR. In general, we shouldn't have `package-lock.json` files in the packages/* directories, and they are not used in prod, so it's minor, but please trim your PRs down to just the files related to the issue being fixed.

Otherwise, merge when ready!

PR #1053, opened 2022-03-02
feat: implement account restriction

alexandrastoica:
API implementation for account restriction:
- Adds middleware to check if account is restricted
- Moves and adds user tag based tests to `auth.spec.js`
- If user account has `HasAccountRestricted` user tag, they can no longer (even if they are PSA enabled):
  - POST /pins
  - POST /pins/:reqId
  - POST /car
  - POST /upload
  - POST /name/:key
- Adds some documentation to the readme

**If a user is restricted and tries to upload a file on the website, the upload will fail and a tooltip with the restriction error will show.**


packages/api/test/pin.spec.js
```diff
@@ -676,6 +676,44 @@ describe('Pinning APIs endpoints', () => {
       assert.strictEqual(data.pin.cid, sourceCid)
       assert.notDeepEqual(data.status, 'failed')
     })
+
+    it('should throw if account is PSA enabled, but restricted', async () => {
```
flea89, line 680: nit.

You've rightly added tests in the context of every single endpoint that has to be restricted.

I wonder if having a whole describe block for pinning authorization where we add these tests, would make them a bit more focused and allow you to avoid some code duplication.  

ie. you have the list of restrict-able tests and loop through them.
Not a big deal, just an idea.

packages/api/src/auth.js
```diff
@@ -1,11 +1,12 @@
 import * as JWT from './utils/jwt.js'
```
flea89, line 1: It'd be great to add some documentation, in this folder to add some rationale on how we're handling the different levels of auth, restrictions, psa, banning etc so that we can refer back to it to understand our choices.

flea89 (commented):
Great work @alexandrastoica!
I've left some comments.

packages/api/test/pin.spec.js
alexandrastoica: I actually debated whether we should have an `auth.spec.js` where we test all the auth things. Having all in one file might help with documentation too since all the expected behaviour will be there.What do you think?

packages/api/src/index.js
```diff
@@ -41,16 +42,16 @@ router.get('/status/:cid',          mode['👀'](auth['🤲'](statusGet)))
 router.get('/car/:cid',             mode['👀'](auth['🤲'](carGet)))
 router.head('/car/:cid',            mode['👀'](auth['🤲'](carHead)))
 
-router.post('/car',                 mode['📝'](auth['🔒'](carPost)))
+router.post('/car',                 mode['📝'](auth['🔒'](auth['🚫'](carPost))))
 router.put('/car/:cid',             mode['📝'](auth['🔒'](carPut)))
```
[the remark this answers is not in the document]
alexandrastoica: Sounds fair

packages/api/test/pin.spec.js
flea89: I like it!

packages/db/index.js
```diff
@@ -139,6 +139,28 @@ export class DBClient {
...
+      .select('id', { count: 'exact' })
+      .eq('user_id', userId)
+      .eq('tag', 'HasAccountRestriction')
+      .eq('value', true)
+      .filter('deleted_at', 'is', null)
+
+    if (error) {
+      throw new DBError(error)
+    }
+
```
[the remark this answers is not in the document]
alexandrastoica: Expecting only one entry for this and pinning auth.

packages/api/src/auth.js
alexandrastoica: Done in readme

packages/api/src/index.js
```diff
@@ -41,16 +42,16 @@ router.get('/status/:cid',          mode['👀'](auth['🤲'](statusGet)))
...
-router.post('/pins',                mode['📝'](auth['📌'](pinPost)))
-router.post('/pins/:requestId',     mode['📝'](auth['📌'](pinPost)))
-router.get('/pins/:requestId',      mode['👀'](auth['📌'](pinGet)))
-router.get('/pins',                 mode['👀'](auth['📌'](pinsGet)))
-router.delete('/pins/:requestId',   mode['📝'](auth['📌'](pinDelete)))
+router.post('/pins',                mode['📝'](auth['🔒'](auth['🚫'](auth['📌'](pinPost)))))
+router.post('/pins/:requestId',     mode['📝'](auth['🔒'](auth['🚫'](auth['📌'](pinPost)))))
+router.get('/pins/:requestId',      mode['👀'](auth['🔒'](auth['📌'](pinGet))))
+router.get('/pins',                 mode['👀'](auth['🔒'](auth['📌'](pinsGet))))
+router.delete('/pins/:requestId',   mode['📝'](auth['🔒'](auth['🚫'](auth['📌'](pinDelete)))))
```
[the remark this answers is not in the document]
alexandrastoica: It was added on line 30 but I just removed it to be able to add the account restriction middleware where necessary.

```'📌': handler => auth['🔒'](withPinningAuthorized(handler))```

File: packages/db/index.js (hunk 4 above)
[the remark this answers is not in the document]
Comment (adamalton) [reply packages/db/index.js:160]: I might be misunderstanding this, but why would this query becomes invalid if the `user_tag_is_not_deleted_idx` index is removed? Wouldn't that just make the query slower? It wouldn't make it not work, right?

File: packages/db/index.js [hunk 6]
```diff
@@ -180,7 +180,7 @@ export class DBClient {
       throw new DBError(error)
     }
 
-    return count > 0
+    return count === 1
```
adamalton, line 183: Would it be better to throw an error if there's more than one row returned? Otherwise, currently, if there's more than one row set to `true`, it will return `false`.

I'd actually be tempted to make a generic `getUserTagValue` function which you pass `'HasAccountRestriction'` or `'HasPsaAccess'` (or whatever) into, and it does the query and the count check, and then returns the value.

packages/api/src/index.js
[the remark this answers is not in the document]
adamalton: Ah right, yes, I'd missed that! Thanks 🙂

packages/db/index.js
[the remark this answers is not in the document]
flea89: If that index is removed multiple `deleted` tags would be allowed isn't it? we'd remove the unique constraint. 
In that scenario, just counting wouldn't be enough. Wouldn't it? Because I would count an "out-of-date" tag.

In that scenario the truth would come from the latest tag.
Hence I was suggesting ways to prevent that if that constraint ever gets removed.
Is it making sense?


We should add a ticket for psa access as well!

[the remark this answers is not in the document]
adamalton: Ah, right, yes, got you. I think we could fix this (for all user tags) by implementing my suggestion here: https://github.com/web3-storage/web3.storage/pull/1053#discussion_r825887574

packages/db/index.js
alexandrastoica: Added the generic function and left the two existing ones as wrappers to check the type. Thanks!

adamalton (approved):

packages/db/index.js
```diff
@@ -130,26 +130,51 @@ export class DBClient {
...
 
+  /**
+   * Check that a user account is restricted.
+   *
+   * @param {number} userId
+   * @returns {Promise<boolean>}
+   */
+  async isAccountRestricted (userId) {
+    return await this.getUserTagValue(userId, 'HasAccountRestriction') === 'true'
+  }
```
vasco-santos, line 141:
```suggestion
  isAccountRestricted (userId) {
    return this.getUserTagValue(userId, 'HasAccountRestriction') === 'true'
  }
```

Given we not wrap this with try catch, no need to create extra promise layer with async await. We can just return the promise and ease a bit async callstack

packages/db/index.js
```diff
@@ -130,26 +130,51 @@ export class DBClient {
...
   /**
    * Check that a user is authorized to pin.
    *
    * @param {number} userId
    * @returns {Promise<boolean>}
    */
   async isPinningAuthorized (userId) {
-    const { count, error } = await this._client
+    return await this.getUserTagValue(userId, 'HasPsaAccess') === 'true'
+  }
```
vasco-santos, line 151: same as above, suggestion not possible because of line 149 not changed 😓

packages/api/src/index.js
```diff
@@ -27,7 +27,8 @@ const auth = {
   '🤲': handler => withCorsHeaders(handler),
   '🔒': handler => withCorsHeaders(withApiOrMagicToken(handler)),
   '👮': handler => withCorsHeaders(withMagicToken(handler)),
-  '📌': handler => auth['🔒'](withPinningAuthorized(handler))
+  '📌': handler => withPinningAuthorized(handler),
+  '🚫': handler => withAccountNotRestricted(handler)
```
vasco-santos, line 31: Curious if you considered any way of combining these 2 when needed. It would be great if we had a clean way of not needing two HTTP Req + 2 SQL ops just to authorize something. 

We would need to add extra logic to DB Client, probably options to changed functions to allow combining requests with multiple tags at once

packages/db/index.js
alexandrastoica: I tried this but then it complains

> This condition will always return 'false' since the types 'Promise<string>' and '"true"' have no overlap. Did you forget to use 'await'?

I'm not sure how we can avoid this since we're comparing it to `true` and we need the promise to be resolved before that. Let me know if you had something different in mind though.

For context: the user tag values are always a string and we don't know what type to expect. To avoid repeating logic, I created `getUserTagValue` to then use wrappers such as this to return the right type (in this case boolean, hence `=== 'true'`).

packages/api/src/index.js
alexandrastoica: I had this thought and I think it's a good optimisation to aim for. Do you think this is a blocker for now? I'd rather tackle it in a separate PR/ticket to keep this one from dragging on too much, but let me know if it's better to do it here.

alanshaw (changes requested):
Can we please get all the user tags in the [auth middleware](https://github.com/web3-storage/web3.storage/blob/main/packages/api/src/auth.js) and add them to `request.auth`?

Then the account restriction middleware just checks the tags and any other code that needs to inspect tags has them readily available.

packages/db/index.js
```diff
@@ -177,6 +157,38 @@ export class DBClient {
     return data.length ? data[0].value : undefined
   }
 
+  /**
+   * Returns all the active (non-deleted) user tags for a user id.
+   *
+   * @param {number} userId
+   * @returns {Promise<string[] | undefined>}
```
adamalton, line 164: Should this be `@returns {Promise<object[]>}` or `@returns {Promise<{tag: string, value: string}[]>`?

adamalton (approved):

packages/api/README.md
```diff
@@ -115,6 +115,10 @@ The given API has a set of three different authentication levels:
 
 The 👮 API methods are only allowed with a Magic Token, and consequently only available via https://web3.storage
 
+### Account restriction
+
+If a user's account is restricted, it means that they might have gone over the storage limit assigned to them. This restriction disables several actions such as uploading files, adding and replacing pin requests, or publishing a name record. Note that even if the account has [pinning service API access](https://docs.web3.storage/how-tos/pinning-services-api/#requesting-access), account restriction will disable adding and replacing of pins. It is however still possible to delete pins and create/delete API tokens. For more information, please email <support@web3.storage>.
```
olizilla, line 120: > It is however still possible to delete pins and create/delete API tokens.

I'm new to this piece of work, so probably missing some context, but I'm wondering why we would still allow folks to delete pins and create/delete api tokens when their account is locked?

packages/api/src/index.js
olizilla: Aye, I think just add an issue for that work is all that's needed here.

olizilla (approved):
Nice!

It'd be good to work together in an issue on the the emoji-auth follow up. I'd like to collapse the number of possible states back down by merging "requires user" with "user account is not locked". Also we can improve the readability a little by passing a list of middleware to each router method rather than wrapping, see: https://github.com/kwhitley/itty-router#middleware

packages/api/README.md
mbommerez: @olizilla This is to allow them to delete files to go back below their quota. Also, if their token was compromised and abused, they can create a new token and block that token and clean up their account.

packages/api/src/index.js
alexandrastoica: Tracking [here](https://github.com/web3-storage/web3.storage/issues/1162)

PR #1001, opened 2022-02-17
feat!: new website frontend and app ui

orvn:
Ships new NextJS website including:

- Marketing site
- App UI
- Docs

See commit list for details, since this is an initiative with a large collection of epics

> Note: working on getting the build running correctly with Cloudflare pages

1-home

2-appui

3-faq

PR #1061, opened 2022-03-03
fix(docs): update docusaurus & fix error in sidebar script

yusefnapora:
Hey @orvn, this targets the new frontend PR branch & fixes up the sidebar script to get rid of the build error. It also updates docusaurus to the latest beta. 

The code that actually displays the sidebar and TOC on the redoc page doesn't seem to be working with the redesign, but that doesn't feel like a blocker to launch. We can fix it later (or not, if we end up moving to nextra).


orvn (approved):
Looks good, will see if it builds as a workspace in the PR against `main` in #1001

yusefnapora:
Right on - feel free to merge or cherry pick as you see fit :)

orvn:
Made a few more adjustments and tried some tests before merging so that it builds to the right place in `/feat/new-brand-frontend/`

PR #1064, opened 2022-03-04
fix: docs deployment issues

joannaong:


orvn (approved):
This PR fixes some issues that appeared as a product of a necessary package update to docs. As well as from changes required to support a nested directory as the base path.

Issue #1036, opened 2022-02-24
Validate received CAR Files


vasco-santos:
@ribasushi found some invalid CARs that we are storing in backups.

We should discuss if we should inspect and validate the blocks when they are uploaded and fail if the CAR is invalid.

For one of the cars in bafybeiahynaxxbthonbdxyyasbitb52knmftegymlkc6cxttdfcwovsz7m ( don't know which, test )
```
Error: mismatch in content integrity
name: bafkreie3ufxjrzjvee2gvq4xqdncmfgmnioxew3wmqdxgd532dve3fl2tm
data: bafkreicozng7awrdjyglxmi6wnxbcpi7reqzsybbcmrj2sbc44afrzzr4y
```
for one in bafybeie5npyt4gn3c2yav3ntrojyxeaxf4fwdin3dsbgrqh745g4trjxmq
```
Error: mismatch in content integrity
name: bafkreicbs2klwgxlq7cvmxyouhduhr2c6m76clz3uexdtfvlgffgvkahm4
data: bafkreigqgdm74bko7stv6i6nwfsxqyqazwy4jpmdz3zpypdmei6khxc73a
```

olizilla:
Currently, if we are unable to pin the full DAG on ipfs-cluster then that upload will not show as pinned in the UI, which is a start, but really, where the api recieves a CAR with 1 or more cid/block mismatches, we should fail loudly and alert the user that the upload should be re-attempted. 

For larger datasets, we could suggest adding to a local ipfs node and using the pinning api so that only the failed/missing blocks are re-attempted... a fancy resume mechanism thanks to the wonders of dags and content-addressing.

We could conceivably also see bitflips when sending data between the api and ipfs-cluster, or to the s3 backups, which need to be considered.

ribasushi:
Regarding bitflips protection in the pipeline, the workflow should be:
1. CF worker starts 2 independent "threads" while the content is streamed in from user:
1.1 md5sum (yes, md5) of the incoming bytes
1.2 check of individual cid+block pairs within the car file itself
2. Once above completes, add the `Content-MD5` header to the PUT request to S3
3. If s3 returns success you know with many-nines certainty that you have a valid, intact car in s3, and you can say 200 to the user

olizilla:
We already iterate all the blocks in the CAR here, so that seems like a good place to verify each multihash
[...]

Issue #922, opened 2022-01-28
Add logtail integration


vasco-santos:
We need to add logtail to web3.storage API. Similar to what we did for https://github.com/nftstorage/nft.storage/pull/860

adamalton:
This has now been deployed in #1325

PR #1068, opened 2022-03-04
fix: verify s3 backup integrity on upload

olizilla:
add `x-amz-checksum-sha256` header on s3 object put requests.

This is nice as we reuse our existing sha256 hash of the car as an integrity check. AWS will calculate the sha256 of the body it receives and send us a 400 BadDigest error if they dont match, to avoid storing CARs that had bits flipped in transit.

It was necessary to update to the latest version of the aws-sdk as support for sha256 integrity checking was not supported in the older one.

Some debug logging is left in for the s3 client in dev mode, as it was an absolute pain to figure out what it was doing otherwise.

see: #1036
see: https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html#AmazonS3-PutObject-request-header-ChecksumSHA256

License: (Apache-2.0 AND MIT)

### DIFF
--- a/packages/api/src/car.js
+++ b/packages/api/src/car.js
@@ -227,17 +227,37 @@ async function backup (blob, rootCid, userId, env, structure = 'Unknown') {
     return undefined
   }
 
-  const data = await blob.arrayBuffer()
-  const dataHash = await sha256.digest(new Uint8Array(data))
-  const keyStr = `raw/${rootCid.toString()}/${userId}/${toString(dataHash.bytes, 'base32')}.car`
+  const data = new Uint8Array(await blob.arrayBuffer())
+  const multihash = await sha256.digest(data)
+  const keyStr = `raw/${rootCid.toString()}/${userId}/${toString(multihash.bytes, 'base32')}.car`
+  // strip the multihash varint prefix to get the raw sha256 digest for aws upload integrity check
+  const rawSha256 = multihash.bytes.subarray(2)
+  const awsChecksum = toString(rawSha256, 'base64pad')
+
+  // see: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/modules/putobjectrequest.html
   const cmdParams = {
     Bucket: env.s3BucketName,
     Key: keyStr,
-    Body: blob,
-    Metadata: { structure }
+    Body: data,
+    Metadata: { structure },
+    // ChecksumSHA256 specifies the base64-encoded, 256-bit SHA-256 digest of the object, used as a data integrity check to verify that the data received is the same data that was originally sent.
+    // see: https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html#AmazonS3-PutObject-request-header-ChecksumSHA256
+    ChecksumSHA256: awsChecksum
   }
 
-  await env.s3Client.send(new PutObjectCommand(cmdParams))
+  try {
+    await env.s3Client.send(new PutObjectCommand(cmdParams))
+  } catch (err) {
+    if (err.name === 'BadDigest') {
+      // s3 returns a 400 Bad Request `BadDigest` error if the hash does not match their calculation.
+      // see: https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#RESTErrorResponses
+      // see: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/index.html#troubleshooting
+      console.log('BadDigest: sha256 of data recieved did not match what we sent. Maybe bits flipped in transit. Retrying once.')
+      await env.s3Client.send(new PutObjectCommand(cmdParams))
+    } else {
+      throw err
+    }
+  }
   return keyStr
 }
 
--- a/packages/api/src/env.js
+++ b/packages/api/src/env.js
@@ -121,6 +121,7 @@ export function envAll (req, env, ctx) {
     env.s3BucketRegion = env.S3_BUCKET_REGION
 
     env.s3Client = new S3Client({
+      // logger: console, // use me to get some debug info on what the client is up to
       endpoint: env.S3_BUCKET_ENDPOINT,
       forcePathStyle: !!env.S3_BUCKET_ENDPOINT, // Force path if endpoint provided
       region: env.S3_BUCKET_REGION,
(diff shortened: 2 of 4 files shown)


olizilla:
@ribasushi the `x-amz-checksum-sha256` header on puts to s3 ended up doing what we wanted... my previous failures were due to the older version of the aws-sdk just silently silently stubbing out that header when i tried to set it. With the latest sdk, we get a 400 BadDigest http error on put when we send them a sha256 that doesn't match.

olizilla:
it is notable that the latest aws-sdk is now automatically adding an additonal header `x-amz-content-sha256` which is similar to `x-amz-checksum-sha256` but appears to contain the sha256 of the body but in hex rather than base64... e.g

```
s3 request headers {
  'content-type': 'application/octet-stream',
  'x-amz-checksum-sha256': 'q3g2SxTkn2avonv2fDG4AIrZFA5vB0jCmWetFl/Jkjs=',
  'x-amz-meta-structure': 'Complete',
  'content-length': '55513',
  host: 's3.us-east-2.amazonaws.com',
  'x-amz-user-agent': 'aws-sdk-js/3.53.1 os/other lang/js md/browser/unknown_unknown api/s3/3.53.1',
  'amz-sdk-invocation-id': '3f2ef788-dc39-48fa-90e4-7cfc07b76b76',
  'amz-sdk-request': 'attempt=1; max=3',
  'x-amz-date': '20220304T112718Z',
  'x-amz-content-sha256': 'ab78364b14e49f66afa27bf67c31b8008ad9140e6f0748c29967ad165fc9923b',
}
POST /car 200 OK (1410.77ms)
```

The old version was not setting that header. It is likely that we are duplicating the check, but I have no control over `x-amz-content-sha256` header and we can manipulate `x-amz-checksum-sha256` and show that if it's wrong we get a useful error, so I'm happy with it.

alanshaw (approved):

packages/api/src/env.js
```diff
@@ -129,5 +130,17 @@ export function envAll (req, env, ctx) {
...
+      env.s3Client.middlewareStack.add(
+        (next, context) => async (args) => {
+          console.log('s3 request headers', args.request.headers)
+          return next(args)
+        },
+        {
+          step: 'finalizeRequest'
+        }
+      )
+    }
```
vasco-santos, line 144: This is not needed anymore right?

packages/api/src/car.js
```diff
@@ -227,17 +227,37 @@ async function backup (blob, rootCid, userId, env, structure = 'Unknown') {
...
 
-  await env.s3Client.send(new PutObjectCommand(cmdParams))
+  try {
+    await env.s3Client.send(new PutObjectCommand(cmdParams))
+  } catch (err) {
+    if (err.name === 'BadDigest') {
+      // s3 returns a 400 Bad Request `BadDigest` error if the hash does not match their calculation.
+      // see: https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#RESTErrorResponses
+      // see: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/index.html#troubleshooting
+      console.log('BadDigest: sha256 of data recieved did not match what we sent. Maybe bits flipped in transit. Retrying once.')
```
vasco-santos, line 255:
```suggestion
```

can be removed?

vasco-santos (commented):
Awesome 🙌🏼

packages/api/src/car.js
olizilla: This was intentional, as I'd be curious to see how many times this appears in the logs, but I could be pursuaded to remove it as I can't say I'd be checking the logs for it (if we had any)

packages/api/src/env.js
olizilla: This was left in to save others the pain of trying to figure out what the hell s3 is up to, as it took a while for me to figure out that i needed to add logging middleware to do it, so I figured it captures a bit of knowledge that we might benefit from again in the future.

packages/api/src/car.js
vasco-santos: We need https://github.com/web3-storage/web3.storage/issues/922 for this. With `console.log` we won't be able to see any logs

packages/api/src/env.js
vasco-santos: I think it is fine to add this as log debug level

packages/api/src/car.js
```diff
@@ -227,17 +227,37 @@ async function backup (blob, rootCid, userId, env, structure = 'Unknown') {
...
   const cmdParams = {
     Bucket: env.s3BucketName,
     Key: keyStr,
-    Body: blob,
-    Metadata: { structure }
+    Body: data,
+    Metadata: { structure },
+    // ChecksumSHA256 specifies the base64-encoded, 256-bit SHA-256 digest of the object, used as a data integrity check to verify that the data received is the same data that was originally sent.
+    // see: https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html#AmazonS3-PutObject-request-header-ChecksumSHA256
+    ChecksumSHA256: awsChecksum
```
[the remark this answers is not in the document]
olizilla: alas no. it does not.

`x-amz-content-sha256` is required by aws for it's signed uploads feature, and uses a chunked "tree-hash" strategy (a merkel hash!) for larger files. We can't control the value of `x-amz-content-sha256` and if we could we'd set it to the wrong value for larger files. It's unrelated to the checksum headers.

see: https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html

`x-amz-checksum-sha256` seems to be a simpler sha256 hash of the body.

see: https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html

olizilla:
I tried some local uploads of 112MB file and compared the times against `main` with s3 backup enabled

| method | run 1 | run  2 | vs main
|---------|------|-------|--------
| chunked w3 put | 30.7s | 16.5s | 25.17s
| single curl put /upload | 125s | 121s | 130s
| single curl put /car | 79.94s | 73s | 106s


with a local ipfs cluster, **all** of the variance comes from s3 uploads... some of the 10MB chunks would go up to s3 in 10s, and some in 1s. Some of that could be my network connection but it was suprising to see that level of variance

```
POST /car 200 OK (12959.15ms)
POST /car 200 OK (13501.06ms)
POST /car 200 OK (13551.99ms)
POST /car 200 OK (3577.20ms)
POST /car 200 OK (3023.86ms)
POST /car 200 OK (3394.19ms)
POST /car 200 OK (2991.47ms)
POST /car 200 OK (3508.17ms)
POST /car 200 OK (3230.08ms)
POST /car 200 OK (2672.82ms)
POST /car 200 OK (3422.13ms)
POST /car 200 OK (2378.45ms)
POST /car 200 OK (797.40ms)
```

packages/api/src/car.js
olizilla: @vasco-santos want me to remove it? or leave it in ahead of #922

vasco-santos: can leave, #922 already has this linked, so we just need to remember to change it

PR #1069, opened 2022-03-04
fix: check bytes match cid

olizilla:
Verify the hash of a blocks bytes matche its proposed cid multihash for every block in a CAR.

see: #1036

## TODO
- [x] validate the blocks
- [x] add tests for an invalid block and a non-sha256 hashed block.

License: (Apache-2.0 AND MIT)


packages/api/src/car.js
```diff
@@ -273,6 +273,12 @@ async function carStat (carBlob) {
     if (blockSize > MAX_BLOCK_SIZE) {
       throw new InvalidCarError(`block too big: ${blockSize} > ${MAX_BLOCK_SIZE}`)
     }
+    if (block.cid.multihash.code !== 0x12) {
```
alanshaw, line 276:
```suggestion
    if (block.cid.multihash.code !== sha256.code) {
```

packages/api/src/car.js
```diff
@@ -273,6 +273,12 @@ async function carStat (carBlob) {
...
+      throw new InvalidCarError(`block with unsupported hash function: ${block.cid.multihash.code} for ${block.cid.toString()}`)
+    }
+    if (!isValidBlock(block)) {
+      throw new InvalidCarError(`block data does not match cid for ${block.cid.toString()}`)
```
alanshaw, line 280:
```suggestion
      throw new InvalidCarError(`block data does not match CID for ${block.cid.toString()}`)
```

packages/api/src/car.js
```diff
@@ -273,6 +273,12 @@ async function carStat (carBlob) {
     if (blockSize > MAX_BLOCK_SIZE) {
       throw new InvalidCarError(`block too big: ${blockSize} > ${MAX_BLOCK_SIZE}`)
     }
+    if (block.cid.multihash.code !== sha256.code) {
+      throw new InvalidCarError(`block with unsupported hash function: ${block.cid.multihash.code} for ${block.cid.toString()}`)
+    }
+    if (!isValidBlock(block)) {
```
olizilla, line 279: yikes. this is async, so here we test that the return is truthy, and the non-null promise always returns true. Thank goodness for tests. Fixing now.

packages/api/src/car.js
```diff
@@ -273,6 +273,13 @@ async function carStat (carBlob) {
     if (blockSize > MAX_BLOCK_SIZE) {
       throw new InvalidCarError(`block too big: ${blockSize} > ${MAX_BLOCK_SIZE}`)
     }
+    if (block.cid.multihash.code !== sha256.code) {
```
vasco-santos, line 276: One of the cases where is not good to support everything. While it is the default, we support other hash functions, including by `ipfs-car` itself: https://github.com/web3-storage/ipfs-car/blob/main/src/pack/index.ts#L21 

Considering we recommend `ipfs-car` and people can pack their own CARs, as well as we don't inform anywhere to the users that we only support in web3.storage sha256 we need to be careful here.

What do you think @olizilla ? Should we support at least the hash functions that are in multiformats repo and properly document it? Or just sha256?

packages/api/test/car.spec.js
```diff
@@ -208,4 +208,66 @@ describe('POST /car', () => {
...
+      },
+      body: new Blob(carBytes)
+    })
+
+    assert.strictEqual(res.ok, false)
+    const { message } = await res.json()
+    assert.strictEqual(message, `Invalid CAR file received: block with unsupported hash function: 0x${sha512.code.toString(16)} for ${cid.toString()}`)
+  })
+
+  it.only('should throw for CAR with a block where the bytes do match the CID', async () => {
```
vasco-santos, line 243:
```suggestion
  it('should throw for CAR with a block where the bytes do match the CID', async () => {
```

packages/api/src/car.js
olizilla: From discussion with @alanshaw and @vasco-santos we're going to

- adjust this PR so that we don't error if we see a non-sha256 hash
- land this PR so that sha256 hashed blocks (the vast majority) *are* validated
- open an issue to see what other hashes we want to pre-validate _(ideally many, but bundle size and perf of hashing functions other than sha2 need to be considered given that the impl is built in to subtle.crypto and may be hardware optimised)_

vasco-santos (approved):
LGTM

PR #1072, opened 2022-03-05
fix: Pagination reset

MichaelPhan:
### Changes
- Fixing query param reference in `useQueryParams`
- Reseting `page` query `order` parameter on change
- Fixing misc build errors that recently popped up


orvn (changes requested):
This is throwing a Type error @MichaelPhan 

```js
./components/tokens/tokenCreator/tokenCreator.js:61:9
Type error: Object is possibly 'null'.

  59 | 
  60 |   const handleInputValueChange = () => {
> 61 |     if (inputRef.current.value !== '') {
     |         ^
  62 |       if (!inputHasValue) {
  63 |         setInputHasValue(true);
  64 |       }
```

orvn (approved):

PR #1080, opened 2022-03-07
feat: app ui component visual feedback enhancements

joannaong:


orvn (changes requested):
@joannaong, it looks like the checkboxes in the dashboard all load in a checked state. 

(Expected behaviour is for them to load in unchecked)

orvn (changes requested):
@MichaelPhan works when running locally as `npm run dev`, but a type error is thrown on `build`, from `modules/zero/components/pagination/pagination.js`

```js
  40 |   const setCurrentPage = useCallback((page) => {
  41 |     queryParam ? setQueryValue(page) : setActivePage(page)
> 42 |     const scrollToElement= document.querySelector(scrollTarget);
     |                                                   ^
  43 |     scrollToElement?.scrollIntoView(true);
  44 |     
  45 |   }, [queryParam, setQueryValue, setActivePage, scrollTarget])
```

orvn (approved):

orvn:
Full list of fixes from this PR:

- [x] Checkmark appears when an action is performed, clearly indicating that something changed on the table
- [x] Fix to modal skew effect, which was being obscured by overflow
- [x] Ellipsis on upload now blink, indicating progress, so the user doesn't think the app is stuck on large chunked uploads
- [x] A toggle appears of the number of storage providers > 5
- [x] Storage providers are linked to Filfox and have link styles
- [x] Improvements to scrolling after pagination, or when loading a page with GET params, in a variety of cases
- [x] Improvements for IE11 (there are still a lot of issues)
- [x] Browser update banner for IE
- [x] Reduced an unintended layout-shif/blinking effect when the table was refreshed by adding a min height

PR #1083, opened 2022-03-08
fix: pin status syncing

alanshaw:
When the DB was updated to use the IPFS node peer ID instead of the Cluster node peer ID the pin status syncing stopped working.


ribasushi:
Ahhh this is why my pin updates stopped rolling in 😿 
I thought it has to do with the badger-calypse, is why I didn't say anything 🤦

Issue #1082, opened 2022-03-08
Upsert pins failing with null value for pin_location_id


vasco-santos:
We are seeing some issues in DB logs in https://github.com/web3-storage/web3.storage/blob/main/packages/api/src/utils/pin.js#L155 . We don’t have locationId here, as OkPins is response from Cluster and not DB. It should likely use the rpc function instead of the supabase client

```
[AQUA] [37-1]  sql_error_code = 23502 ERROR:  null value in column \"pin_location_id\" of relation \"pin\" violates not-null constraint
```

PR #1088, opened 2022-03-09
fix: use rpc for pins upsert

GaryHomewood:
Revert to using a DB function to upsert pins so that we have the location id.

fixes https://github.com/web3-storage/web3.storage/issues/1082


packages/db/postgres/functions.sql
```diff
@@ -214,6 +214,44 @@ BEGIN
...
+AS
+$$
+DECLARE
+  pin_data json;
+  pin_location_result_id BIGINT;
+  pin_result_id BIGINT;
+BEGIN
+  FOREACH pin_data IN array json_arr_to_json_element_array(data -> 'pins')
+  LOOP
+    -- Add to pin_location table if new
```
flea89, line 230: Could we be calling `upsert_pin` here instead? 
We'd avoid some duplication. There's an inconsistency with the data inputs, in one function we use `content_cid` and in the other we use`cid`, but I think that can be easily changed? And we'd benefit from a bit more consistency?

packages/db/index.js
```diff
@@ -452,20 +452,25 @@ export class DBClient {
...
+          }
+        }))
+      }
+    })
 
     if (error) {
       throw new DBError(error)
     }
+
+    return data
```
flea89, line 473: If I'm not mistaken the rpc call is currently returning void, so there's no actual useful data to return here.
It'd be probably nice to return an array of the inserted pin ids, to be consistent with `upsertPin`. If not we get rid of the return statement here.

flea89 (commented):
This should fix it. 🤞 
I've left a couple of comments.

packages/db/index.js
vasco-santos: Yeah, I think we could be consistent and return Pin Ids

packages/db/index.js
```diff
@@ -452,20 +452,25 @@ export class DBClient {
...
-      })), { count: 'exact', returning: 'minimal' })
+    const { data, error } = await this._client.rpc('upsert_pins', {
+      data: {
+        pins: pins.map((pin) => ({
+          status: pin.status,
+          cid: pin.cid,
+          location: {
+            peer_id: pin.location.peerId,
+            peer_name: pin.location.peerName,
+            region: pin.location.region
```
vasco-santos, line 463: Be aware this will change soon (today) to have one more thing: https://github.com/web3-storage/web3.storage/pull/1093/files

So let's wait on that PR to get in and update this

vasco-santos (commented):
Thanks for putting this together @GaryHomewood Overall looking good, let's wait for #1093 to get in to have `ipfsPeerId` column in location. And please look into @flea89's good suggestions

packages/db/postgres/functions.sql
GaryHomewood: Corrected to re-use `upsert_pin` and to return an array of pin Ids.
Also changed to use content_cid consistenly.

packages/db/index.js
GaryHomewood: Merged latest to include ipfs peer id.

packages/db/postgres/migrations/004-add-upsert-pins-function.sql
```diff
@@ -0,0 +1,59 @@
...
+  -- Add to pin_location table if new
+  insert into pin_location (peer_id, peer_name, ipfs_peer_id, region)
+  values (data -> 'pin' -> 'location' ->> 'peer_id',
+          data -> 'pin' -> 'location' ->> 'peer_name',
+          data -> 'pin' -> 'location' ->> 'ipfs_peer_id',
+          data -> 'pin' -> 'location' ->> 'region')
+  ON CONFLICT ( peer_id ) DO UPDATE
+          SET "peer_name" = data -> 'pin' -> 'location' ->> 'peer_name',
+              "region" = data -> 'pin' -> 'location' ->> 'region'
+  returning id into pin_location_result_id;
```
vasco-santos, line 25: Let's do the same as @alanshaw did in https://github.com/web3-storage/web3.storage/pull/1101

vasco-santos (changes requested):

packages/db/postgres/migrations/004-add-upsert-pins-function.sql
GaryHomewood: Can do. Why is it `SELECT 42` ?

```
 -- Add to pin_location table if new
  INSERT INTO pin_location (peer_id, peer_name, ipfs_peer_id, region)
    SELECT * FROM (
      SELECT data -> 'pin' -> 'location' ->> 'peer_id' AS peer_id,
             data -> 'pin' -> 'location' ->> 'peer_name' AS peer_name,
             data -> 'pin' -> 'location' ->> 'ipfs_peer_id' AS ipfs_peer_id,
             data -> 'pin' -> 'location' ->> 'region' AS region
    ) AS tmp
    WHERE NOT EXISTS (
      SELECT 42 FROM pin_location WHERE peer_id = data -> 'pin' -> 'location' ->> 'peer_id'
    );
```

GaryHomewood: Ok now doing the same.

vasco-santos: We talked about this sync.

packages/db/postgres/migrations/004-add-upsert-pins-function.sql
```diff
@@ -0,0 +1,59 @@
+DROP FUNCTION IF EXISTS upsert_pin;
```
vasco-santos, line 1: This files is not needed anymore right? We already have a migration 004

packages/db/postgres/migrations/005-add-upsert-pins-function.sql
```diff
@@ -0,0 +1,60 @@
...
+    ) AS tmp
+    WHERE NOT EXISTS (
+      SELECT 42 FROM pin_location WHERE peer_id = data -> 'pin' -> 'location' ->> 'peer_id'
+    );
+
+  -- Add to pin table if new
+  insert into pin (content_cid, status, pin_location_id, updated_at)
+  values (data ->> 'content_cid',
+        (data -> 'pin' ->> 'status')::pin_status_type,
+        pin_location_result_id,
```
vasco-santos, line 32: Why this change? It is not in `functions.sql`, so I guess we don't need it?

With that in mind, we are not changing `upsert_pin` and only `upsert_pins` should be in the migration?

vasco-santos (changes requested):
Nice, it looks much cleaner now 👌🏼 
Left some comments on migration scripts to wrap thiss up

packages/db/postgres/migrations/004-add-upsert-pins-function.sql
GaryHomewood: That's true, have removed.

packages/db/postgres/migrations/005-add-upsert-pins-function.sql
GaryHomewood: Also true, have removed. Migration is just the new function.

vasco-santos (approved):
LGTM

Please add a comment in release PR for API to not forget about running migration script

GaryHomewood:
Please ensure migration is run `005-add-upsert-pins-function.sql`

alanshaw:
The cron job has changed in the PR and it now relies on a function that currently does not exist in the production DB. So for 2 days now the pins job has been failing because it can't update pins.

flea89:
@alanshaw good catch.

**TL;DR**
As I'm sure you already figure out the fix is releasing [this](https://github.com/web3-storage/web3.storage/pull/1107) and running the migration that has been added there. 

However, this raises some questions for me.

How can we avoid this in the future? 
It doesn't seem documentation or processes make it really clear that any cron code that lands in `main` is automatically in production, and there's no safe-guarding against it, isn't it?
Even if you do know about it, as this PR has proved, it's really easy to forget? 

I wonder if we should have a release step for cron code as well?
ie. Running the cron from a "prod" branch and automating that through release-please could be a way?
This is just the first idea that pop to mind, I'm sure there are better ones if we think about it.

alanshaw:
Extracting from the convo here: https://filecoinproject.slack.com/archives/C02BZPRS9HP/p1647599946603429

> I'm open to suggestions for making this better so it doesn't happen again.
> 
> We could maybe do something where cron is included in a release please process and the yaml is copied (and committed) from the project dir to the root on release?

@flea89 as we both suggested, adding it to release please might help. I think it's unexpected for changes to cron to immediately be in production when that is not the case for the rest of the packages in the repo. Any chance you can open an issue so we can track? 🙏🙏🙏

flea89:
Sure, [here](https://app.zenhub.com/workspaces/web3storage-product-6106d25b58fccc00105aedcb/issues/web3-storage/web3.storage/1129) it is!

Issue #1089, opened 2022-03-09
Validate Cluster Pin status response


vasco-santos:
We currently accept any status response from `ipfs-cluster` and try to insert it in the DB. This might result in problems if expected response values are not there

We got an outage today as its response did not include `ipfsPeerId` in one of the entries of the peerMap.

We should look into loosen the DB schema, and allow peerId to be null. Given we don't overwrite it on the `on conflict` clause it should be fine, unless new peers are added and they don't have peerId specified.

vasco-santos:
We decided to add to `PinLocation` a column `ipfsPeerId` and revert change of using `PeerId` with PeerId to `clusterId`.

PR #1090, opened 2022-03-09
fix: tmp workaround for lack of ipfs peer id on status response

vasco-santos:
While we don't fix https://github.com/web3-storage/web3.storage/issues/1089 this was added to production

### DIFF
--- a/packages/api/src/utils/pin.js
+++ b/packages/api/src/utils/pin.js
@@ -98,12 +98,19 @@ export function toPins (peerMap) {
   // Note: `clusterPeerId` is an internal id, and is only used for cluster admin.
   // The `ipfsPeerId` which we rename to `peerId` can be  used to connect to the underlying ipfs node
   // that stores a given pin, by passing it to `ipfs swarm connect <peerid>`.
-  return Object.entries(peerMap).map(([clusterPeerId, { peerName, status, ipfsPeerId: peerId }]) => {
+  const toPins Object.entries(peerMap).map(([clusterPeerId, { peerName, status, ipfsPeerId: peerId }]) => {
     return {
       status: toPinStatusEnum(status),
       location: { peerId, peerName }
     }
   })
+  
+  const filteredToPins = toPins.filter(pin => pin.location?.peerId)
+  if (toPins.length !== filteredToPins.length) {
+    console.log('Dropping pins that do not have "peerId" on it', toPins.filter(pin => !!pin.location?.peerId))
+  }
+
+  return filteredToPins
 }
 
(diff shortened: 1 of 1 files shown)


packages/api/src/utils/pin.js
```diff
@@ -98,12 +98,19 @@ export function toPins (peerMap) {
   // Note: `clusterPeerId` is an internal id, and is only used for cluster admin.
   // The `ipfsPeerId` which we rename to `peerId` can be  used to connect to the underlying ipfs node
   // that stores a given pin, by passing it to `ipfs swarm connect <peerid>`.
-  return Object.entries(peerMap).map(([clusterPeerId, { peerName, status, ipfsPeerId: peerId }]) => {
+  const toPins Object.entries(peerMap).map(([clusterPeerId, { peerName, status, ipfsPeerId: peerId }]) => {
```
olizilla, line 101:
```suggestion
  const toPins =  Object.entries(peerMap).map(([clusterPeerId, { peerName, status, ipfsPeerId: peerId }]) => {
```

packages/api/src/utils/pin.js
```diff
@@ -98,12 +98,19 @@ export function toPins (peerMap) {
...
+  const toPins = Object.entries(peerMap).map(([clusterPeerId, { peerName, status, ipfsPeerId: peerId }]) => {
     return {
       status: toPinStatusEnum(status),
       location: { peerId, peerName }
     }
   })
+
+  const filteredToPins = toPins.filter(pin => pin.location?.peerId)
+  if (toPins.length !== filteredToPins.length) {
+    console.log('Dropping pins that do not have "peerId" on it', toPins.filter(pin => !!pin.location?.peerId))
```
alanshaw, line 110: We need the `ipfsPeerId` we have no way to re-add the pins records after the upload.

alanshaw (changes requested):

packages/api/src/utils/pin.js
olizilla: yes, but we currently fail an upload that is pinned on some cluster nodes if any one of them comes back with a missing `ipfsPeerId` property. Hector is taking a look on the cluster side, but we need to do something here too.

vasco-santos:
Synced with @alanshaw we need to get this in and fix consequences later on.

alanshaw:
The plan is to revert to using the cluster peer ID and _also_ store the IPFS peer ID. This will take a while to get together so lets get this merged and released and we'll have to follow up to add pin objects for the uploads later.

PR #1097, opened 2022-03-10
perf: improve pins cron

alanshaw:
* Batch CIDs for status queries with cluster (significantly faster than 1 by 1)
* Add an index on `inserted_at` in `pin_sync_request` (19s -> 6s query time improvement)
* Increase `LIMIT` in query to fetch pins to sync (made possible by the index)
* Remove the IPFS proxy client (unused)


packages/db/postgres/tables.sql
```diff
@@ -246,6 +246,7 @@ CREATE TABLE IF NOT EXISTS pin_sync_request
 );
 
 CREATE INDEX IF NOT EXISTS pin_sync_request_pin_id_idx ON pin_sync_request (pin_id);
+CREATE INDEX IF NOT EXISTS pin_sync_request_inserted_at_idx ON pin_sync_request (inserted_at);
```
vasco-santos, line 249: Should we add migration script?

vasco-santos (approved):
LGTM

Issue #1099, opened 2022-03-10
DB locking in create_upload


olizilla:
For the last couple of days we've seen the number of open connections on the db hit ~500 after which requests from the api start hitting timeouts e.g 

```
"Database connection error. Retrying the connection., details: connection to server at \"ec2-54-81-248-174.compute-1.amazonaws.com\" (54.81.248.174), port 5432 failed: FATAL:  remaining connection slots are reserved for non-replication superuser connections\n, hint: undefined, code: "
```

Looking at the queries that have been running for over > 5 mins we see

```
WITH pgrst_source AS (WITH pgrst_payload AS (SELECT $1::json AS json_data), pgrst_body AS ( SELECT CASE WHEN json_typeof(json_data) = 'array' THEN json_data ELSE json_build_array(json_data) END AS val FROM pgrst_payload), pgrst_args AS ( SELECT * FROM json_to_recordset((SELECT val FROM pgrst_body)) AS _("data" json) )SELECT "public"."create_upload"("data" := (SELECT "data" FROM pgrst_args LIMIT 1)) AS pgrst_scalar) SELECT null::bigint AS total_result_set, pg_catalog.count(_postgrest_t) AS page_total, coalesce(string_agg(to_json(_postgrest_t.pgrst_scalar)::text, ','), 'null')::character varying AS body, nullif(current_setting('response.headers', true), '') AS response_headers, nullif(current_setting('response.status', true), '') AS response_status FROM (SELECT "pgrst_source".*FROM "pgrst_source"    ) _postgrest_t
```

which is `create_upload` wrapped in a bunch of postgREST noise.

In the db logs we see this kind of query timing out with:

```
2022-03-10 15:30:01.000 [db] postgres.1836004 [AQUA] [276-1]  sql_error_code = 00000 LOG:  process 1073010 still waiting for ExclusiveLock on tuple (8079,25) of relation 16893 of database 16385 after 1008.361 ms
```

see: https://logtail.com/team/52899/tail?s=110548&a=1646926201000.af316c92-8a55-4211-ac62-9f860e1b911a

The current theory is we are doing needless locking of the db to maybe insert a new pin location (a rare occasion) for every pin of every new root cid.

https://github.com/web3-storage/web3.storage/blob/77af7da8e1ff6b6be155b156656ac863b73e4216/packages/db/postgres/functions.sql#L227-L235

PR #1101, opened 2022-03-10
fix: pin_location lock contention

alanshaw:
For every upload we lock every `pin_location` record to update it with values that never change. This PR changes that behaviour to only insert into the table if the row does not exist.

It results in two `SELECT`s for the same row on the `pin_location` table but the idea is that this will be better than locking the records.

fixes https://github.com/web3-storage/web3.storage/issues/1099 🤞


vasco-santos (approved):
LGTM 🤞

PR #1104, opened 2022-03-11
chore: remove lingering faunadb references

adamalton:
FaunaDB was removed in #610, but there were still various things referencing it. This just removes those references and/or replaces them with the Postgres equivalents.


adamalton:
I prefixed the branch with `fix/`, but it's probably more of a chore really.

vasco-santos (changes requested):
Can you remove db package lock?

packages/api/lib/utils.js
```diff
@@ -0,0 +1,17 @@
+import { DBClient } from '@web3-storage/db'
+
+/**
+ * Create a new Fauna DB client instance from the passed environment variables.
```
olizilla, line 4: Adds a new Fauna DB reference!

packages/report/index.js
```diff
@@ -183,22 +183,17 @@ function renderCSV ({ users, userUploads }, dest, options = {}) {
 }
 
 /**
- * Create a new Fauna DB client instance from the passed environment variables.
+ * Create a new DB client instance from the passed environment variables.
```
olizilla, line 186: This package assumes it can pass graphql queries to the DB client as it was written in a time of Fauna. There is a good chunk more work to do to make this package work with postgres, and it's not clear if we still need it now we have a /metrics endpoint for prometheus metric tracking.

@alanshaw should `report` be refactored to work with postgres or removed as no longer needed?

olizilla (changes requested):
Thanks for sweeping up... some more tweaks to do. Feedback inline.

 Also not that the `db/package-lock.json` shows as changed but there is no change to the `db/package.json` so that should be reverted. Only change the dep tree if you mean to change the dep tree.

packages/api/lib/utils.js
adamalton: copy-paste-o! 🤦

packages/report/index.js
adamalton: This is a good point. The same is true of `add-mock-deals.js` and `get-auth-token.js`. Presumably both of those want to be kept (and therefore refactored)?

adamalton:
@olizilla re. `db/package-lock.json`, I know I haven't changed `db/package.json` (because it was updated in the original PR - #610), but "faunadb" was still listed in `db/package-lock.json`, hence I committed the change to remove it.

It seems that `db/package-lock.json` gets updated by the GitHub Actions Bot when it does a release, but it doesn't seem to have ever removed the "faunadb" entry. It's as if the bot has a persistent environment, and each time it does a release it runs `npm install` and commits the changes to `package-lock.json`, but it doesn't _uninstall_ any packages which are no longer required and so they're still persisting.

packages/api/lib/utils.js
adamalton: Fixed.

packages/api/scripts/add-mock-deals.js
```diff
@@ -2,9 +2,10 @@
  * Add some mock deals to your database for your existing content.
```
alanshaw, line 2: Fauna exposed the GraphQL interface, this file can be removed entirely - it does not work.

packages/api/scripts/get-auth-token.js
```diff
@@ -1,24 +1,24 @@
 /**
  * Temporary utility to obtain an auth token for using the API.
```
alanshaw, line 2: Can remove this also.

packages/report/index.js
alanshaw: Just remove this whole directory.

alanshaw (changes requested):

olizilla:
@adamalton for this PR to land you need to remove the files that @alanshaw listed in comments, and update the branch to fix the conflicts.

adamalton:
Thanks @olizilla, I'd somehow missed the notification for those more recent comments from Alan. I've now removed those files and resolved the conflicts.

So the one remaining thing is the `db/package-lock.json` file. Did you see my comments about this above? Vasco says remove, you say revert, I say update! I'm thinking that we want to keep it so that we've always got a reproducible last-known working version. So given that we've removed `faunadb` from the tree, we want to keep it with the change, right?

olizilla:
Please keep chagesets focused on a single issue. This PR doesn't need to touch the website readme. It's not a big deal for this particular PR but it's a good habit, and gets more meaningful when doing a `fix:` or `feat:` that the release-please bot would see as reason to bump the version of all packages touched by a PR.

packages/website/README.md
```diff
@@ -51,7 +51,7 @@ cd packages/website
 npm run dev
 ```
 
-Yarn: 
+Yarn:
```
olizilla, line 54: plz revert the changes to this file

adamalton:
Closed in favour of #1187

PR #1106, opened 2022-03-11
chore: make 'cron' package use the same ENV vars for its DB settings

adamalton:
This makes it behave the same as the other packages in the repo and therefore makes the env setup simpler.

### DIFF
--- a/packages/cron/src/lib/utils.js
+++ b/packages/cron/src/lib/utils.js
@@ -22,24 +22,13 @@ export function getCluster (env) {
  * @param {Record<string, string|undefined>} env
  */
 export function getDBClient (env) {
-  let token, endpoint
-  if (env.ENV === 'production') {
-    if (!env.PG_REST_JWT) throw new Error('missing PG_REST_JWT environment var')
-    if (!env.PG_REST_URL) throw new Error('missing PG_REST_URL environment var')
-    token = env.PG_REST_JWT
-    endpoint = env.PG_REST_URL
-  } else if (env.ENV === 'staging') {
-    if (!env.STAGING_PG_REST_JWT) throw new Error('missing STAGING_PG_REST_JWT environment var')
-    if (!env.STAGING_PG_REST_URL) throw new Error('missing STAGING_PG_REST_URL environment var')
-    token = env.STAGING_PG_REST_JWT
-    endpoint = env.STAGING_PG_REST_URL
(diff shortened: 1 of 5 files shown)


packages/cron/src/lib/utils.js
```diff
@@ -22,24 +22,13 @@ export function getCluster (env) {
...
-    endpoint = env.DEV_PG_REST_URL
-  } else {
-    throw new Error(`unsupported environment ${env.ENV}`)
+  const token = env.PG_REST_JWT
+  const endpoint = env.PG_REST_URL
+  if (!token) {
+    throw new Error('missing PG_REST_JWT environment var')
+  }
+  if (!endpoint) {
+    throw new Error('missing PG_REST_URL environment var')
```
vasco-santos, line 31: We can't do this change because we use PROD and STAGING in our cron jobs. Example: https://github.com/web3-storage/web3.storage/blob/main/.github/workflows/cron-pins.yml#L32

vasco-santos: We need to look into staging and production. Dev and Prod will have same namings though

vasco-santos (changes requested):

packages/cron/src/lib/utils.js
adamalton: Yes, I was just about to convert this to a draft, as I'd realised there was clearly a prod/staging aspect that I'd missed.

Is this whole thing just not worth it? Should I put the PR in the bin? 😀

vasco-santos: Well, I initially liked the idea of uniforming it. But, also don't like that we use same name for Dev and Prod. 

If you have a good idea to work around that let me know, otherwise yes maybe put it in the bin

adamalton: Ah, I see. So maybe a good solution is:

* For dev use `PG_REST_...` with no prefix.
  - This removes the need for a duplicate set of vars in `.env` / `.env.tpl`.
* For prod, use the `PROD_` prefix.
  - This means that in `cron-pins.yml` we would no longer be changing the names of vars as we map them.

Profit and glory?!

Is there anywhere else where these are referenced that I've missed?

vasco-santos: Yeah go ahead, that should work. Thanks

vasco-santos (approved):
LGTM Thanks

PR #1109, opened 2022-03-14
chore: improv api testing

### LINKED ISSUE #642: Improve set-up for API testing (for package/db)
Improvements to make:
* Flush DB at every test. There is currently no easy way to do this. It'd be quite useful to allow for an easy way to clear data in order to keep our tests nice and encapsulated.
When & how?
  * Between different tests
  * Suites or every "describe" block
  * Should it be a custom decorator?
* Expose DB utils so that we can share factories to populate DB from client packages 
* A nasty bug where multiple results are returned from .single because of inconsistent versions of fetch. (one used in the APIs and the one used when calling a factory from tests). TL;DR accept header was sent twice and the second instance was ignored by PostgRest.
* Move randomCid and, refactor to use `Math.random`

flea89:
Resolves #642 

In this PR
- flush DB at every test
- expose DB utils so that we can share factories to populate DB from client packages 
- fixed a nasty bug where multiple results were returned from .single because of inconsistent versions of fetch. (one used in the APIs and the one used when calling a factory from tests). TL;DR accept header was sent twice and the second instance was ignored by PostgRest.
- moved randomCid and, refactored to use `Math.random`

### DIFF
--- a/packages/api/test/scripts/helpers.js
+++ b/packages/api/test/scripts/helpers.js
@@ -8,7 +8,6 @@ import { keys } from 'libp2p-crypto'
 import * as JWT from '../../src/utils/jwt.js'
 import { JWT_ISSUER } from '../../src/constants.js'
 import { SALT } from './worker-globals.js'
-import crypto from 'crypto'
 import { sha256 } from 'multiformats/hashes/sha2'
 import * as pb from '@ipld/dag-pb'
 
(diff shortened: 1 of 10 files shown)


packages/api/test/scripts/helpers.js
```diff
@@ -52,3 +55,13 @@ export async function updateNameRecord (privKey, existingRecord, newValue) {
 export function getTestJWT (sub = 'test-magic-issuer', name = 'test-magic-issuer') {
   return JWT.sign({ sub, iss: JWT_ISSUER, iat: 1633957389872, name }, SALT)
 }
+
+/**
+ * @param {number} code
+ * @returns {Promise<string>}
+ */
+export async function randomCid (code = pb.code) {
+  const bytes = crypto.randomBytes(10)
```
adamalton, line 64: Minor thing, but this is using a cryptographically secure random algorithm, which isn't really necessary for tests and will probably be slow. If it's called a lot of times then it might be worth swapping this out for a more basic thing using `Math.random()`, just to speed things up.

alanshaw (approved):

Issue #1098, opened 2022-03-10
Don't keep track of Pins with remote status


vasco-santos:
We currently keep all the Pins, even when remote status. This does not scale as number of peers in cluster gets bigger

alanshaw:
Often when calling "status" immediately after adding the statuses will ALL be "Unpinned" so it won't work for most uploads.

I seem to remember hector telling me that the pin allocations actually come back in the response from a call to `/add` but I don't think that's the case. I will ask and update here.

alanshaw:
Hector wants to fix this, I've opened an issue to track https://github.com/ipfs/ipfs-cluster/issues/1598

dchoi27:
can we close this?

flea89:
I've picked this up given we have a few hundred millions "Remote" Pins in prod that should be cleaned up.

Deleting the Pins should sensibly improve performance across the board and especially help with storage notification job.

flea89:
Follow-up PR #1619 to update db type once this is merged, release, and cron have been run.

PR #1110, opened 2022-03-14
fix: add missing statuses

### LINKED ISSUE #1102: Remote pin ends up in "invisible"  queued state
I am trying to pin 60GB of Ukrainian Wikipedia Mirror (https://github.com/ipfs/distributed-wikipedia-mirror/pull/120).

## Problem

I've pinned `bafybeibiqlrnmws6psog7rl5ofeci3ontraitllw6wyyswnhxbwdkmw4ka`  via CLI `ipfs pin remote add` command, received no error, but there is no new pin listed  in web interface, nor in CLI:  `ipfs pin remote ls --service=web3 --status=queued,pinning,failed,pinned` shows no "queued" pins.

## Debugging the "invisible" queued pin

To ensure there is no bug in go-ipfs, I sent raw HTTP request with bare minimum:
```console
$ curl -X POST -H "Authorization: Bearer [secret]"  -H "Content-Type: application/json" -d '{"cid": "bafybeibiqlrnmws6psog7rl5ofeci3ontraitllw6wyyswnhxbwdkmw4ka", "name": "wikipedia_uk_all_maxi_2022-03.zim__via-curl"}' "https://api.web3.storage/pins"  | jq

{
  "requestId": "bf6f216a-6f89-40db-a67c-9879074fc4b2",
  "status": "queued",
  "created": "2022-03-10T20:25:39.554+00:00",
...
```

Ok, no error, got requestId and the pin is supposed to be in queued state.

And here is where it gets interesting: this queued pin is "invisible".

- :anger: There is no UI in web interface for tracking "remote pin states": queued,pinning,pinned or failed.
  - I believe this  is already tracked in https://github.com/web3-storage/web3.storage/issues/1077 so won't go into it here
- :+1:  Doing direct GET for pin status via `https://api.web3.storage/pins/bf6f216a-6f89-40db-a67c-9879074fc4b2` returns correct response, we can see the pin is still in 'queued' state
- :anger: Doing GET for pins in queued state via `https://api.web3.storage/pins?status=queued` returns no results: `{"count":0,"results":[]}` (missing `bf6f216a-6f89-40db-a67c-9879074fc4b2` one)

So I guess the pin is in "queued"  state – which is fine, nobody expects things to be pinning to start immediately – but there is a wide UX problem: no visibility into pin state via web GUI (tracked in #1077), but what is worse, also on CLI and the low level API itself (`/pins?status=queued` shows no "queued" pins).

I am happy to provide more details if needed.

flea89:
Resolves #1102

While I'm pretty confident the fix in this PR should solve the bug reported in #1102, reviewing this made me suspect the logic for filtering PSA pin requests is buggy.

Currently, [here](https://github.com/web3-storage/web3.storage/blob/main/packages/api/src/utils/psa.js#L14)'s the logic we use to convert from Cluster to Psa status does not match the way we filter by those statuses.

We're currently using `query.in('content.pins.status', opts.statuses)` to filter psa requests, but given we different nodes might report different statuses, the same requests might be filtered in different statuses not exclusively.

Here's a description of the bug:
if for a specific CID, 1 node reports `Pinned` and one `Pinning`, the same request would show up in 
`/?status=pinned` and `/?status=pinning`

Unfortunately fixing that bug requires us to write some not trivial remote procedure, and require more time. 
For this reason, I think we should tackle it as a separate ticket?
Would you agree? If yes, I'll create a separate ticket

### DIFF
--- a/packages/api/src/utils/psa.js
+++ b/packages/api/src/utils/psa.js
@@ -41,7 +41,7 @@ export const getEffectivePinStatus = (pins) => {
 
 const psaStatusesToDBStatusesMap = {
   pinned: ['Pinned'],
-  queued: ['PinQueued'],
+  queued: ['PinQueued', 'Unpinned', 'Remote'],
   pinning: ['Pinning'],
   failed: [
     'UnexpectedlyUnpinned',
--- a/packages/api/src/utils/psa.js
+++ b/packages/api/src/utils/psa.js
@@ -28,7 +28,6 @@ export const getEffectivePinStatus = (pins) => {
   // The cluster will return "Remote" if it has been queued for pinning by another node.
   // A cluster status of "Unpinned", "Remote" or "PinQueued" is equivalent to a "queued" PSA status.
   if (pinStatuses.includes('PinQueued') ||
-      pinStatuses.includes('Remote') ||
       pinStatuses.includes('Unpinned')) {
     return 'queued'
   }
@@ -41,7 +40,7 @@ export const getEffectivePinStatus = (pins) => {
 
 const psaStatusesToDBStatusesMap = {
   pinned: ['Pinned'],
-  queued: ['PinQueued', 'Unpinned', 'Remote'],
+  queued: ['PinQueued', 'Unpinned'],
   pinning: ['Pinning'],
   failed: [
     'UnexpectedlyUnpinned',
--- a/packages/api/test/fixtures/init-data.sql
+++ b/packages/api/test/fixtures/init-data.sql
@@ -52,7 +52,8 @@ VALUES (3, 'bafybeifnfkzjeohjf2dch2iqqpef3bfjylwxlcjws2msvdfyze5bvdprfm', 'bafyb
        (3, 'bafybeiaiipiibr7aletbbrzmpklw4l5go6sodl22xs6qtcqo3lqogfogy4', 'bafybeiaiipiibr7aletbbrzmpklw4l5go6sodl22xs6qtcqo3lqogfogy4', 'Car', '2021-07-14T19:27:14.934572Z', '2021-07-14T19:27:14.934572Z');
 
 INSERT INTO pin_location (peer_id, peer_name, ipfs_peer_id, region)
-VALUES ('12D3KooEL1Jc', 'who?', '12D3KooWR1Js', 'where?');
+VALUES ('12D3KooEL1Jc', 'who?', '12D3KooWR1Js', 'where?'),
+       ('12D3KooEL1Jd', 'whoTwo?', '12D3KooWR1Jd', 'whereTwo?');
 
 INSERT INTO pin (status, content_cid, pin_location_id, inserted_at, updated_at)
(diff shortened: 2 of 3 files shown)


packages/api/src/utils/psa.js
```diff
@@ -41,7 +41,7 @@ export const getEffectivePinStatus = (pins) => {
 
 const psaStatusesToDBStatusesMap = {
   pinned: ['Pinned'],
-  queued: ['PinQueued'],
+  queued: ['PinQueued', 'Unpinned', 'Remote'],
```
vasco-santos, line 44: Interesting. So, `Remote` will always be remote and not change https://github.com/web3-storage/web3.storage/blob/api-v5.4.1/packages/api/src/utils/psa.js#L28 Either when it is in queue for other node, or or when the other node actually pinned the content.

So I think we should filter out Remote status completely. Also consider we want to get rid of tracking these https://github.com/web3-storage/web3.storage/issues/1098

vasco-santos: So, from your debugging, you suspect Unpinned is the root of this issue?

flea89: > So I think we should filter out Remote status completely. Also consider we want to get rid of tracking these https://github.com/web3-storage/web3.storage/issues/1098

Ok, I thought that Remote status was a transient state. 
(Are these statuses documented somewhere, we should add a link to that documentation if that's the case).

So assuming 3 nodes in the cluster
`Remote Remote Remote` --> `failed`
`Remote PinError PinError` --> `failed`
etc

If that is the case we should get rid of it from [here](https://github.com/web3-storage/web3.storage/blob/api-v5.4.1/packages/api/src/utils/psa.js#L28) and [here](https://github.com/web3-storage/web3.storage/pull/1110/files/16cddf8f1a92a4ffc3a302f25fd7e827cb1c8a9b#diff-da2c8b3756f3bdc198b20d150e214a5ef6fd6d43b851f1707b189a7b37b9a274R44), isn't?

> So, from your debugging, you suspect Unpinned is the root of this issue?

Not having `Unpinned` here is defo a problem given, the cluster may return "Unpinned" as the initial status.
That results in having that specific PSA request to return `queued` when hitting GET pin/{id} directly (because of [this logic](https://github.com/web3-storage/web3.storage/blob/api-v5.4.1/packages/api/src/utils/psa.js#L32)) but not appear in the list when filtering by `queued`. When I checked the pins for the request Lidel reported all the pins were `Unpinned` indeed.

vasco-santos: @flea89 could only find docs in https://cluster.ipfs.io/documentation/guides/pinning/#filtering-results

> If that is the case we should get rid of it from [here](https://github.com/web3-storage/web3.storage/blob/api-v5.4.1/packages/api/src/utils/psa.js#L28) and [here](https://github.com/web3-storage/web3.storage/pull/1110/files/16cddf8f1a92a4ffc3a302f25fd7e827cb1c8a9b#diff-da2c8b3756f3bdc198b20d150e214a5ef6fd6d43b851f1707b189a7b37b9a274R44), isn't?

Yeah, I think so.

flea89: Removed.

I think we should get this in and I'll create a separate issue for the issue I described [here](https://github.com/web3-storage/web3.storage/pull/1110#issue-1168108666).

vasco-santos (approved):
LGTM

PR #1132, opened 2022-03-18
feat: add rate limiting to client library

### LINKED ISSUE #1050: Add rate limits to client libs
We currently use `p-retry` for the client to handle rate limits from the HTTP API, which exponentially backs off when it gets a rate limit error from the API. Because we know our rate limits, though, we should just incorporate rate limiting into the client, which will be faster than retrying.

Let's add to JS for sure, and go if it's easy.

adamalton:
Fixes #1050 (although doesn't implement it for Go)

### DIFF
--- a/packages/client/src/lib.js
+++ b/packages/client/src/lib.js
@@ -35,11 +35,8 @@ const DEFAULT_CHUNK_SIZE = 1024 * 1024 * 10 // chunk to ~10MB CARs
 const MAX_BLOCK_SIZE = 1048576
 const MAX_CHUNK_SIZE = 104857600
 // These need to match what is configured in the Cloudflare dashboard
-// The limit is 30, but throttledQueue has an off-by-one bug!
-const RATE_LIMIT_REQUESTS = 29
-// We deduct a 100ms buffer from the Cloudflare time period to ensure we stay *under* the limit, as
-// as we get punished with a 30 second wait if we hit it
-const RATE_LIMIT_PERIOD = 10 * 1000 - 100
+const RATE_LIMIT_REQUESTS = 30
+const RATE_LIMIT_PERIOD = 10 * 1000
 
 /** @typedef { import('./lib/interface.js').API } API */
 /** @typedef { import('./lib/interface.js').Status} Status */
--- a/packages/client/src/lib.js
+++ b/packages/client/src/lib.js
@@ -34,7 +34,7 @@ const MAX_CONCURRENT_UPLOADS = 3
 const DEFAULT_CHUNK_SIZE = 1024 * 1024 * 10 // chunk to ~10MB CARs
 const MAX_BLOCK_SIZE = 1048576
 const MAX_CHUNK_SIZE = 104857600
-// These need to match what is configured in the Cloudflare dashboard
+// These match what is enforced server-side
 const RATE_LIMIT_REQUESTS = 30
 const RATE_LIMIT_PERIOD = 10 * 1000
 
@@ -52,15 +52,13 @@ const RATE_LIMIT_PERIOD = 10 * 1000
 /** @typedef { import('./lib/interface.js').Web3Response} Web3Response */
 
 /**
+ * Creates a rate limiter which limits at the same rate as is enforced
+ * server-side, to allow the client to avoid exceeding the requests limit and
+ * being blocked for 30 seconds.
  * @returns {RateLimiter}
  */
-export function createRateLimiter (
-  // These defaults match the limit which is enforced server-side.
-  // They are configurable to allow a shorter timer period in the tests.
-  requestsLimit = RATE_LIMIT_REQUESTS,
-  timePeriod = RATE_LIMIT_PERIOD
-) {
-  const throttle = throttledQueue(requestsLimit, timePeriod)
+export function createRateLimiter () {
+  const throttle = throttledQueue(RATE_LIMIT_REQUESTS, RATE_LIMIT_PERIOD)
   return () => throttle(() => {})
 }
 
--- a/packages/client/src/lib.js
+++ b/packages/client/src/lib.js
@@ -55,6 +55,8 @@ const RATE_LIMIT_PERIOD = 10 * 1000
  * @returns {RateLimiter}
  */
 export function createRateLimiter (
+  // These defaults match the limit which is enforced server-side.
+  // They are configurable to allow a shorter timer period in the tests.
   requestsLimit = RATE_LIMIT_REQUESTS,
   timePeriod = RATE_LIMIT_PERIOD
 ) {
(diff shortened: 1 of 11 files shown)


packages/client/src/lib.js
```diff
@@ -33,6 +34,12 @@ const MAX_CONCURRENT_UPLOADS = 3
 const DEFAULT_CHUNK_SIZE = 1024 * 1024 * 10 // chunk to ~10MB CARs
 const MAX_BLOCK_SIZE = 1048576
 const MAX_CHUNK_SIZE = 104857600
+// These need to match what is configured in the Cloudflare dashboard
+// The limit is 30, but throttledQueue has an off-by-one bug!
+const RATE_LIMIT_REQUESTS = 29
+// We deduct a 100ms buffer from the Cloudflare time period to ensure we stay *under* the limit, as
+// as we get punished with a 30 second wait if we hit it
+const RATE_LIMIT_PERIOD = 10 * 1000 - 100
```
alanshaw, line 42: The limit is enforced if requests _exceed_ this limit - it should be fine for it to be ste to `10 * 1000`.

packages/client/src/lib.js
```diff
@@ -33,6 +34,12 @@ const MAX_CONCURRENT_UPLOADS = 3
 const DEFAULT_CHUNK_SIZE = 1024 * 1024 * 10 // chunk to ~10MB CARs
 const MAX_BLOCK_SIZE = 1048576
 const MAX_CHUNK_SIZE = 104857600
+// These need to match what is configured in the Cloudflare dashboard
+// The limit is 30, but throttledQueue has an off-by-one bug!
+const RATE_LIMIT_REQUESTS = 29
```
alanshaw, line 39: I think it was actually the test that was counting a request that was released on or after the current period elapsed.

packages/client/src/lib.js
```diff
@@ -43,9 +50,28 @@ const MAX_CHUNK_SIZE = 104857600
...
+/** @typedef { import('./lib/interface.js').RateLimiter } RateLimiter */
 /** @typedef { import('./lib/interface.js').UnixFSEntry} UnixFSEntry */
 /** @typedef { import('./lib/interface.js').Web3Response} Web3Response */
 
+/**
+ * @returns {RateLimiter}
+ */
+export function createRateLimiter (
+  requestsLimit = RATE_LIMIT_REQUESTS,
+  timePeriod = RATE_LIMIT_PERIOD
```
alanshaw, line 62: I did not expose these in the public API as users should not change them. They cannot give themselves more requests!

packages/client/test/get.spec.js
```diff
@@ -1,14 +1,15 @@
...
 
 describe('get', () => {
   const { AUTH_TOKEN, API_PORT } = process.env
   const token = AUTH_TOKEN || 'good'
   const endpoint = new URL(API_PORT ? `http://localhost:${API_PORT}` : '')
+  const rateLimiter = createRateLimiter(Infinity, 1000)
 
   it('get a CAR', async () => {
-    const client = new Web3Storage({ token, endpoint })
+    const client = new Web3Storage({ token, endpoint, rateLimiter })
```
alanshaw, line 12: How come this is this being set in every test? Each new instance should receive a new limiter and 30 requests in 10 seconds is likely enough for every test.

alanshaw (changes requested):

alanshaw:
@gobengo woud you mind adding you conventional commit PR title lint check here also?

packages/client/src/lib.js
adamalton: Ah yes, so obvious when I look at it the test again now! Thanks.

packages/client/src/lib.js
adamalton: 👍

packages/client/test/ratelimit.spec.js
```diff
@@ -0,0 +1,23 @@
...
+    const elapsedTime = () => (new Date()) - start
+    // Check how many times the rate limiter returns within the time period
+    let numRequests = 0
+    await (async () => {
+      while (elapsedTime() < timePeriod) {
+        await rateLimiter()
+        numRequests++
+      }
+    })()
+    // throttledQueue has an off-by-one error, hence the plus one
```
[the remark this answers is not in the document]
adamalton: Updated mine to match. Thanks!

packages/client/src/lib.js
adamalton: I made these into parameters so that I can override them in the test. Otherwise the test has to take 30 seconds, which is sloooooow. (Users of the public API could override these defaults if they wanted to, but it would only be to their own detriment, as CloudFlare will then punish them with a 30 penalty when they exceed the limit!)

adamalton: Sorry, 10 seconds, but it still seems like a lot longer than we want a single test to take.

packages/client/test/get.spec.js
adamalton: Ah, yes, I added this because the test suite was taking more than twice as long to run with the rate limiting added. My thinking was: add it to every test so that we ensure we're not being unwittingly bitten by the rate limiter, and so that we don't _become_ unwittingly bitten by it when we extend a test to make more API cals. But actually it's only this ['adds big files' test](https://github.com/web3-storage/web3.storage/blob/3dc51309d76360dbb6b764ec4df87a5990841aee/packages/client/test/put.spec.js#L112-L147) which is being limited. So I'll just leave it on that one and remove the rest.

adamalton: Done

packages/client/src/lib.js
alanshaw: We shouldn't change our public API to accommodate tests, and I think it could be misleading to users to allow these parameters to be changed. IMHO we should either take the hit and wait 10 seconds for the test or move this into a separate file that is not exposed publically and call that function from here (and use the non-public file in the tests).

gobengo:
@alanshaw FWIW the check is now added, and I lowercased the 'Feat' -> 'feat' on this PR to make the check pass.

packages/client/src/lib.js
adamalton: Ok, I've refactored this, so the public API no longer has those parameters. For the test of the rate limiter itself, we take the 10 second hit, but for the `'adds big files'` test (which is where the real slowness was) we now pass in a mock ratelimiter to avoid being throttled.

Ready for re-review.

packages/client/test/ratelimit.spec.js
```diff
@@ -0,0 +1,30 @@
...
+// these must be the same as the values used by createRateLimiter
+const RATE_LIMIT_REQUESTS = 30
+const RATE_LIMIT_PERIOD = 10 * 1000
+
+describe('rate limiter', () => {
+  it('limits to correct rate', async function () {
+    this.timeout(RATE_LIMIT_PERIOD * 2)
+    const rateLimiter = createRateLimiter()
+    const start = new Date()
+    const elapsedTime = () => (new Date()) - start
```
alanshaw, line 14: Nit: I get linter warnings for using `Date` objects like this...

alanshaw (approved):
LGTM

PR #1148, opened 2022-03-21
chore: move dockerized services in cron tests

### LINKED ISSUE #1113: Update cron jobs package test setup
The [cron/package](https://github.com/web3-storage/web3.storage/blob/main/packages/cron/package.json#L13) is still using smoke to mock PostRest and Cluster. With all the challenges that comes with it.

We should be using dockerised instances of both instead and match the behavior in the [api package](https://github.com/web3-storage/web3.storage/blob/main/packages/api/test/hooks.js#L28).

flea89:
Resolve #1113 

## Context

Working on the limit quota epic, we are writing som cron jobs to deal with email sending etc.
With the current cron test setup was really hard to write tests, hence we refactored this to be consistent with API test setup.


GaryHomewood (approved):
Great work. Much cleaner.

PR #1151, opened 2022-03-21
feat: minor update to copy in SP tooltips

orvn:
Minor updates to copy in service provider column tooltips, which appear in the files table.

### DIFF
--- a/packages/website/content/pages/app/account.json
+++ b/packages/website/content/pages/app/account.json
@@ -117,8 +117,8 @@
           "storage_providers": {
             "label": "Storage Providers",
             "tooltip": {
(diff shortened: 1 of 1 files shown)


dchoi27 (changes requested):
Thanks! One nit - the hover-over button for `Storage Providers` is spaced differently than the rest of the columns (maybe because it's a two-line title, but I would think it should be aligned vertically with the second row `Providers` and spaced horizontally similarly to the other two columns
image

dchoi27:
Looks like the latest commit didn't come through? Just LMK when it does!

dchoi27 (approved):
Thank you!

Issue #1153, opened 2022-03-22
Storage limit - Add MailChimp email service


adamalton:
### Overarching user needs / context:
* Users need to receive warning emails from support@web3.storage about their storage quota with specific messages and some dynamically generated data on how much they are using and what their current quota is.
* Admins (support@web3.storage) need to receive an email when a user has gone over their allowed storage quota.

### Scope of this ticket:
This is a subtask of #990. We want to take the `EmailService` class that's been stubbed out in that ticket and make it actually connect to MailChimp and send the emails.

* We'll use the _templated_ transactional emails, so that we can use the visual template editor in MailChimp and therefore don't have to deal with the nightmare of trying to create HTML emails that will work across different email clients; instead our API calls are just `"Send 'template_x' to 'user@example.com' with variables { ... }"`.
* We'll make this as a mini sub-package inside the "cron" package, but it will be entirely generic, so could be shifted out to somewhere else to make it reusable by the other packages.
* Calls to send emails will go through the `EmailSerivce` class, and MailChimp will be plugged in behind this, so in future we could switch to a different email sending provider without changing any of the rest of the application code.

### Acceptance criteria:
- [ ] Once we run this service (on-demand), it needs to take all users returned by the checks and send them the appropriate email with the correct details on their storage usage and how to request an increase.

adamalton:
## Revised architecture

As we've built more of the storage limit and account restriction features, it's become apparent that there are several different cases where we want to send emails. So far:
* To a user when they reach a certain threshold of storage usage.
* To a user when their account is restricted (or de-restricted).
* To admins notifying them of users who are reaching/exceeding their storage limits.

And there will probably end up being more cases in the future for different features in the application.

This means that these emails are not just being triggered from inside the `cron` package, but from other places too (currently the `db` package). This means that:

* This `EmailService` should probably move out of `packages/cron`, probably into its own `packages/email`.
* There will emerge a circular dependency between the `db` package and the `email` package. (The `email` package relies on `db` to store a record of which emails have been sent, and `db` relies on `email` to send emails when certain user tags are updated.) Ideally we want to avoid that circular dependency.

So as well as moving the `EmailService` into a new `packages/email`, I propose that we also:

* Create some kind of "signals" or "events" which one package can publish and another package can listen to.
* The `db` package can publish a "user tag updated" event, which the `email` package can listen to to trigger the appropriate emails.
    - This will mean that `email` depends on `db`, but not the other way round.

[...]

PR #1156, opened 2022-03-23
feat: storage quota notifications

### LINKED ISSUE #990: Storage limit - Framework to get users and record emails sent

GaryHomewood:
Resolves #990

Note: email service integration on a separate ticket #1153


olizilla:
what is 990? If it's an issue number pls just link to it in the PR description

packages/tools/docker/cluster/docker-compose.yml
```diff
@@ -47,7 +47,7 @@ services:
         
     cluster0:
       container_name: cluster0
-      image: ipfs/ipfs-cluster:latest
```
flea89, line 50: I assume the latest version introduces some issue or is buggy?
Do we have that context? 
If yes I wonder if we could add a comment and maybe a follow-up ticket to move this back to latest if needed?

packages/db/utils.js
```diff
@@ -106,3 +106,11 @@ const DEAL_STATUS = new Set([
   'published',
   'active'
 ])
+
+export function parseTextToNumber (numberText) {
+  const number = Number(numberText)
+  if (!Number.isSafeInteger(number)) {
+    throw Error('Invalid number')
```
flea89, line 113: Nit.
I wonder if we could add a more specific Error.
I know it is a general util function and not tied to a specific piece of code, but suggesting a possible cause for this error could be the Postgres returning BigInts?
Just imagining a dev in the future looking at logs and struggling to figure this out?

packages/db/index.d.ts
```diff
@@ -38,7 +39,10 @@ export class DBClient {
   getMetricsValue (key: string): Promise<{ total: number }>
   upsertUser (user: UpsertUserInput): Promise<UpsertUserOutput>
   getUser (issuer: string): Promise<UserOutput>
-  getUsedStorage (userId: number): Promise<number>
+  getUsedStorage (userId: number): Promise<{ uploaded: number, pinned: number, total: number }>
```
flea89, line 42: Shall we have all these input/output types to be defined in `db-client-types.d.ts`?
That would avoid having duplication here and in index.js? To have the types a bit easier to mantain.

packages/db/index.js
```diff
@@ -193,17 +198,119 @@ export class DBClient {
...
+        from_percent: fromPercent,
+        to_percent: toPercent
+      })
+
+    if (error) {
+      throw new DBError(error)
+    }
+
+    return data.map((user) => {
+      const quota = (user.storage_quota)
```
flea89, line 241: nit.
Why not inlining this in the return object?

packages/db/db-client-types.ts
```diff
@@ -330,3 +330,25 @@ export type ListPsaPinRequestResults = {
 export type NameItem = {
   record: definitions['name']['record']
 }
+
+export type UsedStorageItem = {
+  uploaded: string,
+  pinned: string,
```
flea89, line 336: Given `pinned` is an extremly ambitous words ( also uploaded content is pratically pinned on cluster) and it is a common source of confusione, can we be explicit here and use something like:
`psa_requests`
`psa_content`
`psa_pinned`
to avoid any confusion?

It'd be nice to have it renamed across the board. 

If it is easier, given this is not functional issue, this can be done in a follow-up ticket.

packages/db/postgres/functions.sql
```diff
@@ -250,6 +250,95 @@ BEGIN
...
+    (
+      SELECT COALESCE((
+        SELECT COALESCE(SUM(c.dag_size), 0)
+        FROM psa_pin_request psa_pr
+        JOIN content c ON c.cid = psa_pr.content_cid
+        JOIN pin p ON p.content_cid = psa_pr.content_cid
+        JOIN auth_key a ON a.id = psa_pr.auth_key_id
+        WHERE a.user_id = query_user_id::BIGINT
+        AND psa_pr.deleted_at is null
+        AND p.status = 'Pinned'
```
flea89, line 292: I'm sure you're aware, but adding just as a reminder.
This suffers from the problem highlighted by Oli in this [comment](https://github.com/web3-storage/web3.storage/pull/1044#discussion_r834375035) which you already fix and that we'll be soon in master.

packages/db/postgres/functions.sql
```diff
@@ -250,6 +250,95 @@ BEGIN
...
+  pinned        BIGINT;
+  total         BIGINT;
+BEGIN
+  uploaded := 
+    (
+      SELECT COALESCE(SUM(c.dag_size), 0)
+      FROM upload u
+      JOIN content c ON c.cid = u.content_cid
+      WHERE u.user_id = query_user_id::BIGINT
+      AND u.deleted_at is null
```
flea89, line 279: Probably for a separate issue/PR, but here we're counting also "Uploads" that aren't and possibly will never be actually pinned.
We should take the same approach we're taking for "pinned".
In fact, the same query should be re-used for both.

packages/db/postgres/functions.sql
```diff
@@ -250,6 +250,95 @@ BEGIN
...
+  default_quota BIGINT := 1099511627776;
+BEGIN
+  RETURN QUERY
+    SELECT  u.id::TEXT                                      AS id,
+            u.name                                          AS name,
+            u.email                                         AS email,
+            COALESCE(ut.value::BIGINT, default_quota)::TEXT AS storage_quota,
+            (user_used_storage(u.id)).total                 AS storage_used
+    FROM public.user u
+    FULL OUTER JOIN user_tag ut ON u.id = ut.user_id
```
flea89, line 334: I wonder if we should have a LEFT JOIN instead here to disregard users that have been deleted and that still have a tag.

packages/db/postgres/functions.sql
```diff
@@ -250,6 +250,95 @@ BEGIN
...
+    WHERE (ut.tag IS NULL OR (ut.tag = 'StorageLimitBytes' AND ut.deleted_at IS NULL))
+    AND (user_used_storage(u.id)).total::BIGINT >= (from_percent::NUMERIC/100) * COALESCE(ut.value::BIGINT, default_quota)
```
flea89, line 336: I think with the query written this way, `user_used_storage` stored procedure might be run multiple times for the same user. We should make sure it is ran once per user.

flea89: It could be something like this 
```
RETURN QUERY
    SELECT *
    FROM (
        SELECT  u.id::TEXT                                AS id,
          u.name                                          AS name,
          u.email                                         AS email,
          COALESCE(ut.value::BIGINT, default_quota)::BIGINT AS storage_quota,
          (user_used_storage(u.id)).total::BIGINT                   AS storage_used
      FROM public.user u
      LEFT JOIN user_tag ut ON u.id = ut.user_id
      WHERE (ut.tag IS NULL OR (ut.tag = 'StorageLimitBytes' AND ut.deleted_at IS NULL))
    ) sq
    WHERE sq.storage_used >= (from_percent::NUMERIC/100) * sq.storage_quota
    AND (to_percent IS NULL OR sq.storage_used < (to_percent::NUMERIC/100) * sq.storage_quota)
    ORDER BY sq.storage_used DESC;
END
```

packages/db/postgres/functions.sql
```diff
@@ -250,6 +250,95 @@ BEGIN
...
+  SELECT  uploaded::TEXT,
+          pinned::TEXT,
+          total::TEXT
+  INTO    used_storage;
+
+  return used_storage;
+END
+$$;
+
+CREATE OR REPLACE FUNCTION users_by_storage_used(
```
flea89, line 308: My main concern here is performance :(

While this is ran as a cron job, I wonder if running this for a high number of users, and with users that  potentially have lots of files is going to be a problem.

I guess the former could be mitigated by sharding the operation, similarly to what we do in other cron jobs, not sure if there's an easy way to mitigate the latter?

I guess given we could monitor this and see the actual impact perfornance in prod?
Thoughts?

I wonder if makes us subject to DOs attacks?
ie. I create lots of really small files, I could be making this query be extremely expensive and have an impact to overal perf of DB?

packages/db/postgres/tables.sql
```diff
@@ -275,6 +296,8 @@ CREATE TABLE IF NOT EXISTS psa_pin_request
   updated_at      TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL
 );
 
+CREATE INDEX IF NOT EXISTS psa_pin_request_content_cid_idx ON psa_pin_request (content_cid);
```
flea89, line 299: While this sounds like a sensible index to have, can you give the rationale why this was added as part of this PR?

packages/db/postgres/tables.sql
```diff
@@ -299,17 +322,35 @@ CREATE TABLE IF NOT EXISTS metric
...
+
+CREATE TABLE IF NOT EXISTS email_notification_history 
+(
+  id              BIGSERIAL PRIMARY KEY,
+  -- the id of the user being notified
+  user_id         BIGINT NOT NULL REFERENCES public.user (id),
+  email_type      email_type NOT NULL,
+  -- the email service messages unique id
+  message_id      TEXT NOT NULL,
+  sent_at         TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL
```
flea89, line 355: I wonder if it be nice the store the type of services used to send the email. 
I'm sure, thinking if this is needed or not, can be part of  @adamalton work on the email client.

packages/db/test/user-storage.spec.js
```diff
@@ -0,0 +1,201 @@
...
+  it('uses user quota instead of default if one has been specified', async () => {
+    const users = await dbClient.getUsersByStorageUsed({
+      fromPercent: 90
+    })
+    assert.strictEqual(users.length, 2, 'Users with non-default quota included')
+    assert.strictEqual(users[0].email, 'test4@email.com')
+    assert.strictEqual(users[1].email, 'test3@email.com')
+  })
+
+  it('checks email log before sending', async () => {
```
flea89, line 173: Wouldn't it be better to have the email related tests in their own file?

packages/api/test/user.spec.js
```diff
@@ -28,7 +28,7 @@ describe('GET /user/account', () => {
     })
     assert(res.ok)
     const data = await res.json()
-    assert.strictEqual(data.usedStorage, 0)
+    assert.deepStrictEqual(data.usedStorage, { uploaded: 0, pinned: 0, total: 0 })
```
flea89, line 31: Shouldn't we be updating the website to reflect this changes?

packages/cron/test/storage.spec.js
```diff
@@ -0,0 +1,139 @@
...
+    })
+    const cid = 'bafybeibvuy3vcepqxy4plr34twv22vvxol2jjhmjxcrcvuhea5226whpsm'
+    await createUpload(dbClient, Number(user4._id), Number(authKey4), cid, {
+      dagSize: largeFileSize
+    })
+  })
+
+  it('can be executed', async () => {
+    const { stderr: emailLog1 } = await execa('./src/bin/storage.js', { env })
+    assert.match(emailLog1, /storage:checkStorageUsed 🗄 Checking users storage quotas/)
```
flea89, line 129: Can we be testing actual data or function calls instead of stderr?

packages/cron/src/lib/email.js
```diff
@@ -0,0 +1,89 @@
...
+  Used75PercentStorage: 'Used75PercentStorage',
+  Used80PercentStorage: 'Used80PercentStorage',
+  Used85PercentStorage: 'Used85PercentStorage',
+  Used90PercentStorage: 'Used90PercentStorage',
+  UsedOver100PercentStorage: 'UsedOver100PercentStorage'
+}
+
+const log = debug('email:EmailService')
+
+export class EmailService {
```
flea89, line 13: Ignoring this for now, since it's likely going to be tweaked and updated by @adamalton

packages/cron/src/lib/email.js
```diff
@@ -0,0 +1,89 @@
+import debug from 'debug'
+
+export const emailType = {
```
flea89, line 3: Given this represents an postgres type it should be part of the DB package and imported by client packages.
What you think?

packages/db/utils.js
adamalton: Yes, given that this is inside the `db` package, we can probably at least give a _suggestion_ that it might be BIGINT related.

packages/cron/src/lib/email.js
```diff
@@ -0,0 +1,89 @@
...
+  Used75PercentStorage: 'Used75PercentStorage',
+  Used80PercentStorage: 'Used80PercentStorage',
+  Used85PercentStorage: 'Used85PercentStorage',
+  Used90PercentStorage: 'Used90PercentStorage',
+  UsedOver100PercentStorage: 'UsedOver100PercentStorage'
```
adamalton, line 8: In my mind, all of these are implicitly "over" whatever the number is (or more precisely, greater than or equal to the number). So in my head the 100% one doesn't need an explicit "over". (I get that it's the top one, but in theory we could add ones for 110% and 120% as well (depending on whether/how/when we restrict them), so it's not necessarily the top one.)

packages/db/postgres/migrations/006-users-by-storage-used.sql
```diff
@@ -0,0 +1,115 @@
...
+        'Used80PercentStorage',
+        'Used85PercentStorage',
+        'Used90PercentStorage',
+        'UsedOver100PercentStorage'
+      );
+  END IF;
+END
+$$;
+
+CREATE TABLE IF NOT EXISTS email_notification_history
```
adamalton, line 105: My instinct is to call this table `email_history`, rather than `email_notification_history`, just because it keeps it as a more generic thing. I.e. not all emails are _necessarily_ notifications (they could be "please click this link to verify your email address", or "click this link to log in", or "here's the receipt for your account charges this month").

packages/db/utils.js
GaryHomewood: Added some documentation to make use explicit.

packages/db/index.d.ts
GaryHomewood: Yep that's a good shout. Have also renamed a bit for clarity/consistency.

packages/db/postgres/functions.sql
GaryHomewood: Good to have a reminder, have merged from main.

packages/api/test/user.spec.js
GaryHomewood: Yes, resolved by #1183

packages/cron/src/lib/email.js
GaryHomewood: That makes sense. Have removed the "over"

packages/db/postgres/migrations/006-users-by-storage-used.sql
GaryHomewood: Also makes sense. Made more generic.

packages/cron/src/lib/email.js
GaryHomewood: Agree. Have moved to db client types.

packages/db/postgres/functions.sql
GaryHomewood: Thanks this is super helpful. Have implemented.

packages/db/postgres/functions.sql
GaryHomewood: Yep, have included.

packages/db/db-client-types.ts
GaryHomewood: I agree, its definitely a source of confusion. I feel like it should be done as a separate ticket.

packages/db/postgres/functions.sql
GaryHomewood: Sounds like a good candidate for another ticket definitely.

packages/db/postgres/functions.sql
GaryHomewood: Agree, this is definitely something we should monitor in production to see if there are performance issues.

packages/db/postgres/tables.sql
GaryHomewood: This was added for #1043 to improive perfomance when getting pinned storage amopunt. Seems to have been removed in a merge from main, so its being reintstated here.

packages/db/test/user-storage.spec.js
GaryHomewood: Agree, moved to `email.spec.js` in the `db` package.

packages/db/test/utils.js
```diff
@@ -21,6 +23,30 @@ export async function createUser (dbClient, options = {}) {
...
+ */
+export async function createUserTag (dbClient, userId, options = {}) {
+  const { data } = await dbClient._client
+    .from('user_tag')
+    .upsert({
+      user_id: userId,
+      tag: options.tag || '',
+      value: options.value || '',
+      reason: options.reason || 'test reason'
+    })
```
adamalton, line 42: In order to be inline with the way the `user_tags` table is intended to work, this should really always do an insert, and set `deleted_at` on the previously-active row.

adamalton: Actually, we could also move this function into `packages/db/index.js` and then it would achieve this for us at the same time: #1195

GaryHomewood: Have moved to the `db` package, now that its needed elsewhere.

packages/cron/test/storage.spec.js
GaryHomewood: Raised as a follow-up improvement ticket #1203

packages/db/utils.js
flea89: Ta.

packages/db/index.js
```diff
@@ -1,10 +1,18 @@
...
+  normalizeUpload,
+  normalizeContent,
+  normalizePins,
+  normalizeDeals,
+  normalizePsaPinRequest,
+  parseTextToNumber
 } from './utils.js'
 import { ConstraintError, DBError } from './errors.js'
 
+import { emailType } from './constants.js'
```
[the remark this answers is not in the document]
GaryHomewood: Yes, have uppercased.

packages/cron/src/jobs/storage.js
```diff
@@ -0,0 +1,74 @@
...
+  })
+
+  // Send email every 7 days
+  users = await db.getUsersByStorageUsed({
+    fromPercent: 85,
+    toPercent: 90
+  })
+  await emailService.sendEmails({
+    users,
+    email: EMAIL_TYPE.Used90PercentStorage
```
flea89, line 52: Can you please double check all the email types in this file?

packages/cron/test/storage.spec.js
flea89: cool

packages/db/index.d.ts
flea89: thanks

packages/cron/src/jobs/storage.js
GaryHomewood: Good catch, have corrected and added to the test.

adamalton:
One other thought on this is that I'm wondering whether we should skip the call to `this.db.logEmailSent()` in [email.js](https://github.com/web3-storage/web3.storage/pull/1156/files#diff-1e19ad462334a3b1a9069a436c1943da0f6d37d79ad1025d7992c3301e5a744a) for now, because it will currently log that it's sent an email even though it hasn't.

So as as we do #1178 before we actually set this going as a cron job then I guess it's fine. But it might be good to comment those lines out for safey?

packages/cron/src/jobs/storage.js
```diff
@@ -0,0 +1,74 @@
...
+  })
+
+  users = await db.getUsersByStorageUsed({
+    fromPercent: 75,
+    toPercent: 80
+  })
+  await emailService.sendEmails({
+    users,
+    email: EMAIL_TYPE.Used75PercentStorage
+  })
```
adamalton, line 71: Should these have `numberOfDays: 7`?

GaryHomewood: The method has `numberOfDays = 7` as the default.

adamalton: Oh yeah, fair enough!

PR #1158, opened 2022-03-23
fix: create stream when needed

alanshaw:
Replaces the import candidate `content` property with a getter function that allows the stream to be created (and file descriptor opened) only when it is accessed.

The issue is that when there are many thousands of files in a directory, we'll attempt to open that many file descriptors eagerly, which can cause `EMFILE` errors on some systems.

refs https://github.com/nftstorage/nft.storage/issues/1672
refs https://github.com/nftstorage/nft.storage/pull/1693


olizilla (approved):
👍

PR #1196, opened 2022-04-01
fix: update incorrect dag sizes (#1059)

### LINKED ISSUE #1059: Inaccurate claimed DAG sizes
DAG sizes for some upload types are incorrect (reporting smaller than actual). The delta between size_claimed and size_actual can get quite big.

This issue is a sister issue to: https://github.com/nftstorage/nft.storage/issues/1427, but higher prio here, given that we are tracking upload sizes for account limits.

@mbommerez going to add this to the shortlist! It somewhat blocks the account limit restrictions.

flea89:
Resolves #1059

## Details
- Created a remote procedure to update content.dag_size for all instances that are either null or are different from cargo
- To be able to write tests effectively using cargo, I copied the dag carbo schema. I had to update some tests to have everything working with new schema.
- Added a manual action, where we can input a date to run cron to migrate existing data.

## Todo
- Once released, we can manually run the job to update historical data.


packages/db/test/dagSizes.spec.js
```diff
@@ -0,0 +1,128 @@
...
+      const u = await getUpload(dbClient, c[0], user._id)
+      assert.strictEqual(u.dagSize, c[1])
+      assert.strictEqual(allUploadsBefore.find((u) => u.cid === c[0]).updated,
+        allUploadsAfter.find((u) => u.cid === c[0]).updated, `Updated date of ${c[0]}`)
+    }))
+  })
+
+  // Skipping this since for some reason the updated_at field  gets back as unchanged
+  // Testing this manually shows the value is actually updated. Not sure what is wrong.
+  it.skip('updates the updated_at field', async () => {
```
flea89, line 117: Am I missing something obvious here?

.github/workflows/cron-dagcargo-sizes.yml
```diff
@@ -23,8 +28,9 @@ jobs:
...
           ENV: ${{ matrix.env }}
-          STAGING_PG_CONNECTION: ${{ secrets.STAGING_PG_CONNECTION }}
-          STAGING_RO_PG_CONNECTION: ${{ secrets.STAGING_PG_CONNECTION }} # no replica for staging
-          PROD_PG_CONNECTION: ${{ secrets.PROD_PG_CONNECTION }}
-          PROD_RO_PG_CONNECTION: ${{ secrets.PROD_RO_PG_CONNECTION }}
+          PROD_PG_REST_JWT: ${{ secrets.PROD_PG_REST_JWT }}
+          STAGING_PG_REST_JWT: ${{ secrets.STAGING_PG_REST_JWT }}
+          PROD_PG_REST_URL: ${{ secrets.PROD_PG_REST_URL }}
+          STAGING_PG_REST_URL: ${{ secrets.STAGING_PG_REST_URL }}
+          AFTER: ${{ inputs.after }}
```
alanshaw, line 35: We moved to using a direct connection to the DB because:

1. Queries are not subject to the 30s timeout heroku dyno imposes
2. Queries do not add load to the production postgrest dyno
3. Queries can be sent to the replica database (RO - read only) where they do not add load to the production DB

alanshaw (changes requested):

.github/workflows/cron-dagcargo-sizes.yml
alanshaw: Please, if making a big change to the way cron jobs or other part of the API is working then seek concensus before doing the work.

flea89: Thanks for the context @alanshaw. 

I wasn't aware this was the case and given the `DbClient` is exposed in the `cron` package and used in other cron jobs I didn't question if wasn't ok to be used in this one as well. 
I guess from my perspective I wasn't conscious I was introducing a big change given It was a pattern I've seen used in the package. ( Don't get me wrong, now that you pointed out is quite an obvious fundamental change indeed).

I wonder if we should add some context, and documentation to make this more obvious? 
I might add something.

That said, I agree it makes a lot more sense to use a direct connection and leverage the replica given the task is quite heavy on reads.

I'll update the PR.

packages/db/postgres/docker/docker-compose.yml
```diff
@@ -14,6 +14,7 @@ services:
       PGRST_DB_ANON_ROLE: postgres
       PGRST_JWT_SECRET: super-secret-jwt-token-with-at-least-32-characters-long
       PGRST_LOG_LEVEL: info
+      PGRST_DB_SCHEMAS: public, cargo
```
flea89, line 17: Despite this not being not required, I left it here since it could be useful in the future if we want to query cargo schema directly fom DbClient locally.

packages/cron/src/bin/dagcargo-sizes.js
```diff
@@ -4,14 +4,18 @@ import { updateDagSizes } from '../jobs/dagcargo.js'
 import { envConfig } from '../lib/env.js'
 import { getPg } from '../lib/utils.js'
 
-const oneMonthAgo = () => new Date().setMonth(new Date().getMonth() - 1)
+/**
+ *
+ * @param {number} days
+ */
+const xDaysAgo = (days) => new Date().setDate(new Date().getDate() - days)
```
flea89, line 11: Not sure we had a whole month here?

I've updated to take just a week into consideration.
It could actually be even shorter given the cron runs every 38 minutes.

.github/workflows/cron-dagcargo-sizes.yml
flea89: @alanshaw Is the replica pointing to a cargo replica as well?

alanshaw: yes, all our usage is on a dagcargo replica (albeit the same one everywhere)

alanshaw:
Is this ready for another review yet?

flea89:
@alanshaw it is, yes. 

I've asked @adamalton to give it a review before passing it back to you.
But I think it is more efficient if you just go for it. 
Thanks 🙏

packages/cron/test/cargo.spec.js
```diff
@@ -0,0 +1,161 @@
...
+    await Promise.all(wrongSizes.map(async c => {
+      const u = await getUpload(dbClient, c[0], user._id)
+      assert.strictEqual(u.dagSize, c[2], `Not updated ${c[0]}`)
+    }))
+    const toBeUpdated = contentItems.filter((c) => getToBeUpdatedNull(c) || getToBeUpdatedWrong(c))
+
+    assert.strictEqual(toBeUpdated.length, updatedCids.length)
+  })
+
+  // Skipping this since for some reason the updated_at field  gets back as unchanged
```
alanshaw, line 148: One or all of them? We need to resolve this - it sounds like a bug.

packages/cron/test/cargo.spec.js
```diff
@@ -0,0 +1,161 @@
...
+  // Testing this manually shows the value is actually updated. Not sure what is wrong.
+  it.skip('updates the updated_at field', async () => {
+    const { allUploadsBefore, allUploadsAfter } = await updateDagSizesWrp({ user })
+
+    const nullOrZeroSizes = contentItems.filter((c) => getToBeUpdatedNull(c) || getToBeUpdatedWrong(c))
+    await Promise.all(nullOrZeroSizes.map(async c => {
```
alanshaw, line 154: This `Promise.all` is unnecessary.

alanshaw (changes requested):

packages/cron/src/jobs/dagcargo.js
```diff
@@ -37,45 +26,39 @@ UPDATE public.content
...
-    const cids = contents.map((c) => c.cid)
-    const { rows: sizes } = await roPg.query(FIND_DAG_SIZES, [cids])
-
     /* eslint-disable camelcase */
-    for (const { cid_v1, size_actual } of sizes) {
-      log(`💪 ${cid_v1} ${size_actual} bytes`)
-      await rwPg.query(UPDATE_CONTENT_DAG_SIZE, [size_actual, cid_v1])
+    for (const { cid, size_actual } of contents) {
+      log(`💪 ${cid} ${size_actual} bytes`)
+      await rwPg.query(UPDATE_CONTENT_DAG_SIZE, [size_actual, cid])
```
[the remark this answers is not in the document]
flea89: Absolutely!
I actually started doing it as part of this PR  but, as you suggested, I agree it's better to keep it separate.
Created a ticket [here](https://app.zenhub.com/workspaces/web3storage-product-6106d25b58fccc00105aedcb/issues/web3-storage/web3.storage/1263)

packages/cron/test/cargo.spec.js
flea89: Good catch, I probably refactored the tests and left this over.

packages/cron/test/cargo.spec.js
flea89: This is actually passing now, got rid of the comment and `.skip`.

packages/cron/src/jobs/dagcargo.js
```diff
@@ -37,45 +26,39 @@ UPDATE public.content
...
-    log(`ℹ️ ${offset + contents.length} of ${total} processed in total`)
-    offset += limit
+    updatedCids.push(...contents.map(r => r.cid))
+    log(`ℹ️ Updated ${updatedCids.length} in current iteration`)
   }
 
+  log(`ℹ️ Updated ${updatedCids.length} in total`)
+
   log('✅ Done')
+  return updatedCids
```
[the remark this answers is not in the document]
flea89: That is fair point.
At the same time, I quite liked the idea that the function would return the updated cids, given that its main purpose is to update content.
I'm not too opinionated and I've updated this to return void instead.

.github/workflows/cron-dagcargo-sizes.yml
flea89: thanks!

packages/cron/src/bin/dagcargo-sizes.js
flea89: Thanks

packages/db/test/dagSizes.spec.js
flea89: This is now resolved.

packages/api/test/fixtures/pgrest/get-user-uploads.js
```diff
@@ -11,7 +11,7 @@ export default [
         dataCid: 'bafybeiek5gau46j4dxoyty27qtirb3iuoq7aax4l3xt25mfk2igyt35bmf',
         dataModelSelector: 'Links/0/Links',
         dealId: 2424133,
-        expiration: '2023-02-03T14:11:00+00:00',
+        expiration: '2021-09-11T14:11:00+00:00',
```
alanshaw, line 14: Why are these changing?

alanshaw (changes requested):
Changes look good, just a couple of queries.

flea89:
@alanshaw for both your comments, the reason is now we are using a closer cargo "setup" to the one we use in production.
`cargo.sql` is now basically a copy of the one used by cargo.

That's debatable, here's the rationale behind it: 
it was really hard to write tests for this with only the table that were imported, I was at a crossroad where I had to decide if keep adding "fake" representation of cargo schema or just use the real one.

I opted for the second one because testing something closer to production gives us a bit more peace of mind our tests are effective.
As usual, that comes at a cost, having a more complicated `cargo` schema here to maintain and slightly harder fixture to write.

At the same time I feel we should be moving to writing functions to populate data in our DB (we already started) and that should mitigate the issue.

Let me know what you think!

packages/api/test/fixtures/init-data.sql
```diff
@@ -74,17 +74,27 @@ VALUES ('Pinned', 'bafybeifnfkzjeohjf2dch2iqqpef3bfjylwxlcjws2msvdfyze5bvdprfm',
...
-(2424133, 'bafybeiek5gau46j4dxoyty27qtirb3iuoq7aax4l3xt25mfk2igyt35bmf', 'f144zep4gitj73rrujd3jw6iprljicx6vl4wbeavi', 'f0678914', 'queued', 1102102, 2570902, '2021-09-09 16:30:52.252233+00', '2021-09-10 00:45:50.408956+00', 'containing sector active as of 2021-09-10 00:36:30 at epoch 1097593', '2021-09-11 14:11:00+00', 1097593, '2021-09-10 00:36:30+00', '2023-02-03 14:11:00+00');
+INSERT INTO cargo.aggregate_entries (aggregate_cid, cid_v1, datamodel_selector) VALUES
+('bafybeiek5gau46j4dxoyty27qtirb3iuoq7aax4l3xt25mfk2igyt35bme', 'bafybeifnfkzjeohjf2dch2iqqpef3bfjylwxlcjws2msvdfyze5bvdprfm', 'Links/0/Links'),
+('bafybeiek5gau46j4dxoyty27qtirb3iuoq7aax4l3xt25mfk2igyt35bmf', 'bafybeica6klnrhlrbx6z24icefykpbwyypouglnypvnwb5esdm6yzcie3q', 'Links/0/Links');
+
+INSERT INTO cargo.clients (client, filp_available) VALUES
+('f144zep4gitj73rrujd3jw6iprljicx6vl4wbeavi', 1);
+
+INSERT INTO cargo.providers (provider) VALUES
+('f0678914');
```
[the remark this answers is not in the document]
alanshaw: The snippet in my previous comment shows that `cargo.providers` is not imported.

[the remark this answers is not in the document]
flea89: Sorry I meant in `cargo.testing.sql`.

Comment still applies and it's very much still debatable  :P 
I work on this so long ago that I don't even recall the very specific need I had but it related to painful testing.

But the high level (and debatable) rationale is in [comment](https://github.com/web3-storage/web3.storage/pull/1196#issuecomment-1113137906)

olizilla:
@flea89 what's the status of this?

flea89:
@olizilla Unfortunately this has been stuck in the review queue for a while :( 

But I had a chat with @alanshaw yesterday about this.
The only real open point is around `cargo.testing.sql` and the fact in this PR I've ported the whole schema from cargo.
We agreed with Alan it'd be great to keep it simple (to make it easier to maintain, easier to read and understand etc).

I remember the original `cargo.testing.sql` was making testing slightly painful, but I honestly don't remember why. 
I'll have a look tomorrow and we'll try to simplify that part.

packages/api/test/fixtures/init-data.sql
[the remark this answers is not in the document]
flea89: Resolving this one, superseded by [this](https://github.com/web3-storage/web3.storage/pull/1196#issuecomment-1128796482)

packages/api/test/fixtures/pgrest/get-user-uploads.js
flea89: Not relevant after reverting back to simplier `cargo.testing.sql`

flea89:
@alanshaw this is now back to using simpler `cargo.testing.sql`.

What's the action plan for running this in prod?
Are you going to be accountable for that or do you expect me to take care of it?
Let me know.

packages/cron/test/cargo.spec.js
```diff
@@ -0,0 +1,163 @@
...
+    // cid, public.content.dag_size, cargo.dag.size_actual
+    contentItems = [
+      [await randomCid(), 120, 120],
+      [await randomCid(), 105, 1000], // To be updated
+      [await randomCid(), 110, 110],
+      [await randomCid(), 125, 90], // To be updated
+      [await randomCid(), 0, 90], // To be updated
+      [await randomCid(), 100, null],
+      [await randomCid(), null, null],
+      [await randomCid(), 0, 0]
```
alanshaw, line 87: More code but these tests would be much more readable if these were objects like this for example:

```suggestion
      [{ cid: await randomCid(), dagSize: 0, dagSizeActual: 0 }]
```

...you could even `map` these arrays into that format if you wanted.

alanshaw (approved):

alanshaw:
> What's the action plan for running this in prod?
> Are you going to be accountable for that or do you expect me to take care of it?

Once merged we should just be able to run it from the github actions interface, with the date parameter. I'm fine with you running that :)

packages/cron/test/cargo.spec.js
flea89: Really good point, not sure why I ended using an array here.
Updated.

PR #1201, opened 2022-04-02
feat: add storybook to website

joannaong:
Add storybook to website. This is a first pass with setup and configuration and since stories on common components. We'll continue to add more later.


orvn:
Going to evaluate whether it's worth adding Storybook to the site, for the purpose of documenting all existing components in a separate ticket. This PR only has initial components documented.

PR #1205, opened 2022-04-04
feat: docs troubleshooting

### LINKED ISSUE #1052: Docs troubleshooting note on updating to latest client if facing upload speed issues
We need to add a note to the Troubleshooting section in the docs:
https://docs.web3.storage/how-tos/troubleshooting

> **I am experiencing slowness in uploads to web3.storage.**
> 
> Please make sure that you are using the latest version of web3.storage client: 3.5.6. There is an issue in the latest Node version (https://github.com/nodejs/node/pull/42117) and we reverted using native version of Blob in latest release until that fix is shipped.

---

Background info from Discord

GaryHomewood:
Resolves #1052


flea89 (approved):
LGTM!

Issue #1206, opened 2022-04-04
Use IANA media type application/vnd.ipld.car


lidel:
See https://github.com/ipfs/in-web-browsers/issues/192#issuecomment-1087506324

olizilla:
[ipfs-car@0.7.0](https://github.com/web3-storage/ipfs-car/releases/tag/v0.7.0) is released with the new iana media type set to `application/vnd.ipld.car` when packing to a Blob

#1215 updates our api and client to use the new mediat type. That PR will need to be merged and the client and api released before we can update the `w3` cli and website transitive dep on ipfs-car@0.6.2 via the previous release of the web3.storage client. A follow up PR will be submitted once `web3.storage` is updated on npm with the change.

PR #1215, opened 2022-04-05
feat: use IANA media type application/vnd.ipld.car

olizilla:
We now have an official IANA media type for CARs `application/vnd.ipld.car` so use it here.

wip: web3-storage/web3.storage#1206
needs: https://github.com/web3-storage/ipfs-car/pull/119
see: https://www.iana.org/assignments/media-types/application/vnd.ipld.car

License: (Apache-2.0 AND MIT)


alanshaw (approved):
LGTM

Issue #1048, opened 2022-02-28
Docs - Generalize answer in docs troubleshooting for any import error to use prebuilt module


dchoi27:
Right now, in https://docs.web3.storage/how-tos/troubleshooting#i-need-to-use-webpack-4
we recommend folks use the prebuilt module for Webpack 4

However, it's a generally good solution for import errors (most common one that comes up is `Can't resolve 'ipfs-car/blockstore/memory'`). We should change the question title to:
"I'm getting import errors when importing the Javascript client library"
And response to say: Some common situations where this might happen include:
- Using Webpack 4 (e.g., for create-react-app)
- Getting an error like `Can't resolve 'ipfs-car/blockstore/memory'`
- (would be good to do 10 min of research through old issues and the community channel to see what other situations this pops up in)

@mbommerez - @yusefnapora is pretty overloaded this and next week, mind if we get this in your queue?

mbommerez:
Thanks @dchoi27 . Yea we can take that one 👍

PR #1216, opened 2022-04-05
feat: update troubleshooting docs

alexandrastoica:
Closes https://github.com/web3-storage/web3.storage/issues/1048

### DIFF
--- a/packages/docs/docs/how-tos/troubleshooting.md
+++ b/packages/docs/docs/how-tos/troubleshooting.md
@@ -14,6 +14,12 @@ Some common situations where this might happen include:
 - Using Webpack 4 (e.g., for `create-react-app`)
 - Getting an error like `Can't resolve 'ipfs-car/blockstore/memory'`
 
+Add the following import to the top of your `.js` scripts:
+
+```javascript
+import { Web3Storage } from ‘web3.storage/dist/bundle.esm.min.js’
+```
+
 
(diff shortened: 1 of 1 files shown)


yusefnapora (approved):
This looks good to me :) Thanks for picking it up, @alexandrastoica

packages/docs/docs/how-tos/troubleshooting.md
```diff
@@ -7,14 +7,20 @@ description: A collection of common issues for new developers, and solutions to
...
+Some common situations where this might happen include:
+
+- Using Webpack 4 (e.g., for `create-react-app`)
+- Getting an error like `Can't resolve 'ipfs-car/blockstore/memory'`
+
+To get around this issue, add the following import to the top of your `.js` scripts to import the pre-webpack-bundled version of Web3.Storage:
 
 ```javascript
 import { Web3Storage } from 'web3.storage/dist/bundle.esm.min.js'
 ```
```
alanshaw, line 21:
````suggestion
```javascript
import { Web3Storage } from 'web3.storage/dist/bundle.esm.min.js'
```

Alternatively, you can try upgrading to Webpack 5 or using a bundler that supports [`exports` in `package.json`](https://nodejs.org/api/packages.html#subpath-exports).
````

alanshaw (approved):
LGTM, but IMO we should also note the real fix in addition to the band aid.

PR #1219, opened 2022-04-05
fix: Return 400 on malformed form-data

### LINKED ISSUE #1217: malformed content-disposition header should not cause 5xx
Context https://sentry.io/organizations/protocol-labs-it/issues/3084035333/?project=5864302

Gozala:
fixes #1217


vasco-santos (changes requested):
Thanks, just needs lint fix below

vasco-santos (approved):

Issue #1184, opened 2022-03-30
Upgrade ipfs-cluster to v1


olizilla:
Cluster v1 has breaking changes, and significant perf improvements

Before we can use it we need to:

- [x] Deploy a staging v1 cluster. https://github.com/protocol/bifrost-infra/issues/1791
- [x] Update and release the cluster client library https://github.com/nftstorage/ipfs-cluster/pull/40
- [x] Update and release the cluster client librart with support for https://github.com/ipfs/ipfs-cluster/pull/1626 
  - client PR https://github.com/nftstorage/ipfs-cluster/pull/42
- [x] Update web3.storage api with the new client: https://github.com/web3-storage/web3.storage/pull/1227
- [x] Update staging deployment config to use staging v1 cluster.
- [ ] Update pin cron job with new client: https://github.com/web3-storage/web3.storage/pull/1228

From @hsanjuan: Updating web3.storage ipfs-cluster to v1.0.0-rc3 will take approximately 2 hours, but the part where writes are disabled should be short. The suggested deployment steps are:

- [x] Stop cron
- [x] Stop writes to cluster
- [ ] Upgrade cluster
- [x] Deploy web3.storage API with updated cluster client https://github.com/web3-storage/web3.storage/pull/1225
- [ ] Merge cron job https://github.com/web3-storage/web3.storage/pull/1234
- [ ] Enable writes to v1 cluster
- [ ] Restart cron
- [ ] Sit down and wait to see how it goes _(potentially downgrade if we don't like how things look)_

olizilla:
Currently the staging api writes to the prod ipfs-cluster. This was convenient, but to be able to deploy an api change to cluster and test it, we are now going to need a staging cluster that we can update to v1 while leaving prod on v0.x

olizilla:
@thattommyhall is setting up a staging cluster for us https://github.com/protocol/bifrost-infra/issues/1791

thattommyhall:
ETA today!

PR #1227, opened 2022-04-07
feat: api support for ipfs-cluster@1.0

olizilla:
The IPFS Cluster v1 release has much improved perf, and some breaking API changes.

In this PR we upgrade the ipfs-cluster client and local test environment to work with the v1 cluster API.

### Testing it

- Our integration tests have been updated to use the v1 cluster and now pass with this changeset. 
- This PR is also tested locally by pointing a dev api at the `nft.storage.ipfscluster.io` cluster which is now used for QA and is running ipfs-cluster@v1.0.0-rc.4
- The staging deployment should be tested when this PR is merged to main.

### Deploying it

The **staging API cluster credentials must be updated** on Cloudflare after this PR is merged to main.

To deploy to production we need to:
- Put the API into READ-ONLY (check, do we read from cluster in read-only? if so do we hit any api that has changed?
- Deploy ipfs-cluster@1.0.0-rc.4 to the production web3 cluster
- Merge the API release PR and wait for the API deployment to complete.
- Merge the https://github.com/web3-storage/web3.storage/pull/1228 PR, to update the affected cron jobs
- Put the API back into RW mode and monitor.
- FIN

see: #1184
see: https://github.com/nftstorage/ipfs-cluster/releases/tag/v5.0.0

License: (Apache-2.0 AND MIT)

### DIFF
--- a/packages/tools/docker/cluster/docker-compose.yml
+++ b/packages/tools/docker/cluster/docker-compose.yml
@@ -47,7 +47,7 @@ services:
         
     cluster0:
       container_name: cluster0
-      image: ipfs/ipfs-cluster:v0.14.5
+      image: ipfs/ipfs-cluster:v1.0.0-rc3
       depends_on:
         - ipfs0
       environment:
--- a/packages/tools/docker/cluster/docker-compose.yml
+++ b/packages/tools/docker/cluster/docker-compose.yml
@@ -47,7 +47,7 @@ services:
         
     cluster0:
       container_name: cluster0
-      image: ipfs/ipfs-cluster:v1.0.0-rc3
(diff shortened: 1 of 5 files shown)


packages/tools/docker/cluster/docker-compose.yml
```diff
@@ -47,7 +47,7 @@ services:
         
     cluster0:
       container_name: cluster0
-      image: ipfs/ipfs-cluster:v0.14.5
+      image: ipfs/ipfs-cluster:v1.0.0-rc3
```
vasco-santos, line 50:
```suggestion
      image: ipfs/ipfs-cluster:v1.0.0-rc4
```

vasco-santos (approved):
LGTM

PR #1230, opened 2022-04-07
feat: Gradient background optimization

svvimming:
This PR replaces the many gradient background images with a single image that is styled using scss (opacity, rotation, blur) to achieve the many variations throughout the site. The image used is a square for ease of use with rotational transformations. A blur effect is applied using the css `filter: blur()` property to smooth out pixelation, especially when stretching the image at very large screen sizes. Some measures for performance have also been taken, specifically a wrapper div with `overflow: hidden` reduces load on the browser's paint cycle by limiting the area which the filter has to process.


orvn (changes requested):
Looks great, except for an issue on Safari, for both iOS and MacOS, where the filter blends the white background at the edges, making a white gradient appear at the edges.

Can we fix that by scaling it up and clipping the overflow?

gradient edges on iOS

packages/website/components/gradientbackground/gradientbackground.scss
```diff
@@ -7,6 +7,7 @@
   z-index: -1;
   filter: blur(1rem);
   overflow: hidden;
+  transform: scale(1.05);
```
orvn, line 10: Aha, nice fix :)

orvn (approved):

orvn:
Added a fix that includes a background color if the image filter takes an extra moment to paint on older machines 

Paint performance is still quite good, but in some cases does take a second. This does not impact core web vitals metrics like LCP (actually the LCP is better after this PR than before).

Issue #1190, opened 2022-03-30
ci: should take screenshot of website homepage and upload as artifact, so that stakeholders can visualize any changes


gobengo:
Motivation:
* provide same review/testing automation as https://github.com/nftstorage/nft.storage/pull/1725/files#
* @yusefnapora told me it could be good here too

Acceptance Criteria
* when GitHub Actions builds PRs, the check should have upload artifacts containing website screenshots for firefox, chromium

PR #1231, opened 2022-04-07
ci: github actions store website homepage screenshot

gobengo:
Why?
* https://github.com/web3-storage/web3.storage/issues/1190

Decisions
* Is it better to add `@playwright/test` in the repo root or in packages/website?
  * originally I did the latter, but it felt awkward because at test-time it also needed packages/api running. Because the tests span packages, I put them in repo root. I don't feel very strongly about it either way. I think as more package-specific tests/configuration are added, it will be easy to colocate them in their appropriate packages and link to them from the root config

How to Test
* See artifacts from website workflow: https://github.com/web3-storage/web3.storage/actions/runs/2111847287

Possible Futures
* [x] Fix that it looks like the test is flaky. It seems to always [fail]() the first time, then succeed.
  * I think I fixed this by increasing the playwright timeout to 2min from 1 (it probably includes boot-up time). There are still [failish](https://github.com/web3-storage/web3.storage/actions/runs/2112017515) annotations on the checks page, but I think they are from older runs? the [newest runs are green](https://github.com/web3-storage/web3.storage/actions/workflows/website.yml)
* [x] Fix that the test-e2e jobs aren't red even though underlying tests are failing (as [pointed out by @vasco-santos](https://github.com/web3-storage/web3.storage/pull/1231#discussion_r847053775))
  * [example failure on test-e2e error](https://github.com/web3-storage/web3.storage/runs/5977853422?check_suite_focus=true)
* [x] Fix that github action test-e2e's are flaky or failing, e.g. [here](https://github.com/web3-storage/web3.storage/runs/5889916913?check_suite_focus=true#step:9:73)

Possible Futures
* more assertions using playwright test, e.g. the html <title> or main important content. (these can be feature-specific-things, so I just focused on laying the automation out for now)


package.json
```diff
@@ -21,6 +21,7 @@
     "api": "npm start -w packages/api",
     "web": "npm start -w packages/website",
     "test": "npm run lint && npm run test --workspaces",
+    "test:e2e": "npx playwright test",
```
vasco-santos, line 24: Shouldn't this live in website package json?

packages/website/tests/homepage.e2e.spec.ts
```diff
@@ -0,0 +1,31 @@
+import { test, expect, Page, TestInfo } from '@playwright/test';
+import * as fs from 'fs';
+import { extname } from 'path';
+
+test.beforeEach(async ({ page }) => {
+  await page.goto('/');
```
vasco-santos, line 6: Looks like tests are failing. Interesting that CI Job was not affected by it

vasco-santos (changes requested):
Thanks for porting this here @gobengo

package.json
gobengo: I logged this decision in the PR:
> * Is it better to add @playwright/test in the repo root or in packages/website?
>   * originally I did the latter, but it felt awkward because at test-time it also needed packages/api running. Because the tests span packages, I put them in repo root. I don't feel very strongly about it either way. I think as more package-specific tests/configuration are added, it will be easy to colocate them in their appropriate packages and link to them from the root config

packages/website/tests/homepage.e2e.spec.ts
gobengo: I forked where this was at into another branch and [tests pass](https://github.com/web3-storage/web3.storage/runs/5889791211?check_suite_focus=true).

In what way did it look like the tests are failing?
This was in orig PR:
> There are still [failish](https://github.com/web3-storage/web3.storage/actions/runs/2112017515) annotations on the checks page, but I think they are from older runs? the [newest runs are green](https://github.com/web3-storage/web3.storage/actions/workflows/website.yml)

packages/website/package.json
```diff
@@ -18,6 +18,7 @@
   },
   "dependencies": {
     "@magic-ext/oauth": "^0.7.0",
+    "@playwright/test": "^1.20.2",
```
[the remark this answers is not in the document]
gobengo: good catch. It should also be in same package that invokes it, which is currently the repo root (pending discussion of your other comment). For now I pushed https://github.com/web3-storage/web3.storage/pull/1231/commits/99c6b989b405f69bd6401d619b00bb6dc17e1b1e

packages/website/tests/homepage.e2e.spec.ts
vasco-santos: I was looking the logs of the workflow job: https://github.com/web3-storage/web3.storage/runs/5889916913?check_suite_focus=true#step:9:142

It is weird because it is green, but per the logs it keeps failing. Any ideas?

It looks like beforeEach always fails https://github.com/web3-storage/web3.storage/runs/5889916913?check_suite_focus=true#step:9:178

package.json
vasco-santos: I was expecting consistency on the scripts, so I would expect this to be in website, as part of its `test` script, which would be triggered by the `"web": "npm start -w packages/website",` of main package.

However, that is a good point that this is `e2e` testing and uses multiple packages. I also not have a strong preference, so would probably put it in line with how we do in nft.storage for consistency.

gobengo: @vasco-santos I'll try that

gobengo: @vasco-santos I've updated to move all the testing stuff back into packages/website.

packages/website/package.json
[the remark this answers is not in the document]
gobengo: https://github.com/web3-storage/web3.storage/pull/1231/commits/c00c6408d3dabd15016889c6b3502349456aac1e

packages/website/tests/homepage.e2e.spec.ts
gobengo: @vasco-santos The only thing I can think of is that maybe playwright didn't consider that there were any test failures because the failure was in `beforeEach`, none of the tests were actually run?

I tested my latest changes [here](https://github.com/web3-storage/web3.storage/runs/5977853422?check_suite_focus=true) to try to ensure that failed tests also fail the CI

vasco-santos: >  that there were any test failures because the failure was in beforeEach, none of the tests were actually run?

Probably the case. Mind opening an issue in playwright test?

.github/workflows/website.yml
```diff
@@ -23,7 +21,54 @@ jobs:
...
+          - packages/website/test-results
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node-version }}
+      - uses: bahmutov/npm-install@v1
+        env:
+          DEBUG: pw:install
```
vasco-santos, line 47:
```suggestion
```

Can we take it out?

vasco-santos (approved):
LGTM! Nice work @gobengo

.github/workflows/website.yml
gobengo: will do

yusefnapora (approved):
Nice!

packages/website/tests/homepage.e2e.spec.ts
gobengo: I'm not able to validate that theory with a reduced test case: https://github.com/gobengo/test-playwright/runs/5994068552?check_suite_focus=true

The error [we saw](https://github.com/web3-storage/web3.storage/runs/5889916913?check_suite_focus=true#step:9:143) involved an `NS_ERROR_ABORT`, which I think indicates an aborted request in firefox, but that could be for any number of reasons?
https://github.com/microsoft/playwright/search?q=NS_ERROR_ABORT&type=issues

Issue #1115, opened 2022-03-16
PSA filtering by status is buggy


flea89:

dchoi27:
@mbommerez this doesn't fit cleanly in a PI but it's probably a quick fix, so would be good to include in this sprint!

alexandrastoica:
Adding some notes from the investigation:

When we pin a cid, we get back a pin request containing among others a `pin[Object]` which contains pins and their status. The pin request object also contains an overall status (i.e. `pinning`), however the pins might have various statuses. 

Take for example: request id `ab62cf3c-c98d-494b-a756-b3a3fb6ddcab` has two pins – one with status `pinning`, other with status `pinned`.

![tables displaying the example above](https://user-images.githubusercontent.com/9439090/162456113-097af0ca-443d-4554-9396-8b5a041828b0.png)
_Note that the second table is the `pin` array_

When we query the db in [`listPsaPinRequests()`](https://github.com/web3-storage/web3.storage/blob/main/packages/db/index.js#L935), we receive an array with one or multiple statuses. Per the example, if we filter by `status=pinning` we’ll get back the request above and when we filter again by `status=pinned`, we get the same one back. This is inconsistent and confusing on the user end, as to them the same pin request will appear as `pinned` when filtered by `pinned` and as `pinning` when filtered by `pinning`. This may also be true for other statuses such as `queued` and `failed`. 

NB: The [current convention](https://github.com/web3-storage/web3.storage/blob/main/packages/api/src/utils/psa.js#L14) is to always return
`pinned` when at least one node is `pinned`,  ELSE
`pinning` when at least one node is `pinning`, ELSE
`queued` when at least one node is `queued`, ELSE
`failed` when at least one node is `failed`


The issue lies in the way [we apply the statuses filter](https://github.com/web3-storage/web3.storage/blob/main/packages/db/index.js#L954):

```
query = query.in('content.pins.status', opts.statuses)
```

Taking the example above, when `opts.statuses = [‘pinning’]`, the query will “match” the request id because at least one pin has status of `pinning` and keep only the pin objects that match:

![tables displaying the example above](https://user-images.githubusercontent.com/9439090/162456487-92814050-796e-409a-9589-967a69b98687.png)

We need to update the way we filter by statuses to not only look for a status, but also make sure the other statuses are not present so that when a user filters by a single status, a pin request will be returned for the appropriate status only.

francois-potato:
So far I tried:

[...]

PR #1232, opened 2022-04-08
feat: add pinned files to list

### LINKED ISSUE #1077: Add pinned files to /files/ page list in web3
User who have started using pinning services on web3.storage, currently do not see these files listed under their files in the web3.storage UI.

This needs a discussion and UX design input on how to best differentiate such files from files that the user uploaded directly into web3.storage (as opposed to just pinning existing CIDs).

GaryHomewood:
As a web3.storage user I would like to see content that I have pinned using the pinning service API 

Resolves #1077 

### Context 
Currently the DB package includes two separate methods

```javascript
/**
  * List uploads of a given user.
  *
  * @param {number} userId
  * @param {import('./db-client-types').ListUploadsOptions} [opts]
  * @returns {Promise<Array<import('./db-client-types').UploadItemOutput>>}
  */
 async listUploads (userId, opts = {}) {
```

Used by the APis to list uploaded files in the dashboard (and also exposed to REST API users).
The users file list at https://web3.storage/account/ is currently quite slow to ### load.

```javascript
/**
  * Get a filtered list of pin requests for a user
  *
  * @param {string} authKey
  * @param {import('./db-client-types').ListPsaPinRequestOptions} [opts]
  * @return {Promise<import('./db-client-types').ListPsaPinRequestResults> }> }
  */
 async listPsaPinRequests (authKey, opts = {}) {
```

Currently used by GET pins/ endpoint. 

### General assumptions

- We should minimise the number of API calls wherever possible.
- Pinning API users are just a small percentage of the total
- There's a preference in showing the “all” files in the same list
- Is there a requirement to differentiate between the two?

### Issues

The current implementation aggregates data from the 2 different tables (`upload`, `psa_pin_request`) by requesting them separately, resulting in 2 HTTP requests to posgREST.

We don’t have direct access to performance data so it’s hard for us to gauge the impact and overhead of multiple REST calls, but we’re concerned that this approach may impact the performance and experience of the users account page. This is true also for users that don’t use the pinning APIs, so the majority of them.

There is also an additional request to the `auth_key` table. The latter could/should be improved by joining to the user table and retrieving on user id, to get pinned files for any of the users' auth keys. ( Let’s ignore this for the sake of this discussion)

### Alternative Options

**1. List psa content only for `HasPsaAccess=true` Users**  

Pros

- Most users keep same performance as today
- We keep logic in JS land which is generally more maintainable and easy to understand by developers
- Generally speaking it sensibly quicker to implement

Cons

- `HasPsaAccess=true` Users will still have suboptimal experiences
- Users that had `HasPsaAccess` and then disabled it won’t see their historical data.

If we have a really small number of users using PSA APIs this might be a trade off we’re happy to make?

**2. Posgres RPC**

Create a DB function that combines uploads and pins in a SELECT…UNION 

This may mean the where and order by clause becomes complex:

```sql
order by (case when sort_column = 'column1' and sort_direction = 'ASC' then column1 end) asc,
         (case when sort_column = 'column1' and sort_direction = 'DESC' then column1 end) desc,
         (case when sort_column = 'column2' and sort_direction = 'ASC' then column2 end) asc,
         (case when sort_column = 'column2' and sort_direction = 'DESC' then column2 end) desc,
         . . .
```

Pros

- Best for performance, avoid multiple calls to postgREST

Cons

- Filtering logic, especially for PSA List, is quite cumbersome. Having it in SQL makes it harder to understand and maintain. 
- It would be more complex to implement.

**3.  Show uploads and PSA content in different tables in dashboard**

Pros

- Not a lot to implement on the Backend

Cons

- Break one of our assumptions (There's a preference in showing the “all” files in the same list)
- Need to figure out the UX for it
- It differs from NFT experience


packages/db/index.js
```diff
@@ -325,6 +325,98 @@ export class DBClient {
...
+    }
+
+    const cids = pins?.map((u) => u.content.cid)
+    const deals = await this.getDealsForCids(cids)
+    return pins?.map((pin) => ({
+      ...normalizePsaPinRequest(pin),
+      type: 'PinRequest',
+      ...pin.content,
+      deals: deals[pin.content.cid] || []
+    }))
```
adamalton, line 391: Should this have `|| []` on the end, so that the return type is always an array, even if `pins?.map` is undefined?

packages/db/index.js
```diff
@@ -325,6 +325,98 @@ export class DBClient {
...
+    const userFiles = [
+      ...uploadedUserFiles,
+      ...pinnedUserFiles
+    ]
+
+    // Sort applied to combined array
+    if (opts.sortBy === 'Name') {
+      userFiles.sort((a, b) => ((opts.sortOrder === 'Asc') ? (a.name > b.name) : (a.name < b.name)) ? 1 : -1)
+    } else {
+      userFiles.sort((a, b) => ((opts.sortOrder === 'Asc') ? (new Date(a.created).getTime() - new Date(b.created).getTime()) : (new Date(b.created).getTime() - new Date(a.created).getTime())))
```
adamalton, line 413: I _think_ this could be simplified to:

```js
const sortProperty = options.sortBy === 'Name' ? 'name' : 'created'
userFiles.sort((a, b) => (((opts.sortOrder === 'Asc') === (a[sortProperty] > b[sortProperty])) ? 1 : -1)
```

packages/db/index.js
```diff
@@ -325,6 +325,98 @@ export class DBClient {
...
+          )))`
+    const size = opts?.size || 10
+    // TODO get pinned for all of the users auth keys?
+    const authKeys = await this.listKeys(userId)
+    const authKeyIds = authKeys.map((authKey) => authKey._id)
+
+    let query = this._client
+      .from(psaPinRequestTableName)
+      .select(pinnedQuery)
+      .eq('content.pins.status', 'Pinned')
```
adamalton, line 362: I think this will probably suffer from the problem described in #1115. I don't think it's worth trying to fix it here, but maybe just worth a TODO comment here and a note on #1115 to fix this as part of that ticket?

adamalton (commented):
Aside from the few minor comments I've made, I think this looks good. I'll now leave it at the mercy of the discussion about latency / number of queries and what we want to do about that.

packages/db/index.js
GaryHomewood: Yes, that's better. Have added test as well.

adamalton: NIce.

alanshaw:
Personally I'd have a separate menu item and page for listing these.

"feat: add pinned files to list" suggests that this is a list of _pins_, but it's not entirely accurate, we're actually listing pin _requests_.

_Pin requests_ have different properties to _uploads_, for instance you can send a request to pin the same content multiple times with different metadata and you can delete _one_ of those pin requests and effectively still be "pinning" that CID.

If we were listing _pins_ and not _pin requests_ then it _might_ be ok to list them together...

For background, the NFT.Storage PSA was implemented such that a pin request for the same CID multiple times does not result in multiple instances/rows. They are also stored in the same table as uploads (just with a different `type`). IMO it was a mistake (of mine) in NFT.Storage to group these things together.

Given that users have to apply to use this API, the method of receiving the data is completely different and the way you have to think about management of these items is different I think it's more than reasonable to separate them from the list of uploads.

PR #1236, opened 2022-04-08
fix: safari checkmark is always checked

joannaong:
Fixing issue where checkmark in safari is always checked.

Safari does not support negative stroke-dashoffset so I've adjusted it to use positive value. However, this affects the transition animation. So in order to "fix" it, I've just made safari to animate opacity instead of "drawing tick".


orvn (approved):
Passed tests in MacOS Safari 15 and iOS Safari 15, lgtm!

PR #1242, opened 2022-04-11
fix: db delete should match content cid

### LINKED ISSUE #1202: car delete fails with 404
If I upload a car file using "POST /car", and then go to the "files" page and try to delete it, it fails with a 404 (the UI just fails, you have to look at the console to realize the API returned 404). This happens for any such car file, you can never delete any of them.

vasco-santos:
We were previously deleting uploads by source_cid, which does not match the CIDs we provide in all the API endpoints like list.

Closes #1202

### DIFF
--- a/packages/db/index.js
+++ b/packages/db/index.js
@@ -369,7 +369,7 @@ export class DBClient {
         updated_at: now
       })
       .match({
-        source_cid: cid,
+        content_cid: cid,
         user_id: userId
       })
       .is('deleted_at', null)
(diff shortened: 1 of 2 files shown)


alanshaw:
Eek, no the user should always use/see the `source_cid` - the CID they know the content by. The `content_cid` is for us.

olizilla:
@vasco-santos I think this PR is not needed... perhaps a new one to fix the CID used in list is. Am closing out.

Issue #1175, opened 2022-03-29
Provide dweb link to saved file with filename search parameter as before


gyuri-lajos:
I really appreciate web3.storage it is a dream come true. A true game changer.
The new look is nice.
I did trust that search will be added later. Thanks for that.

![](https://dweb.link/ipfs/bafkreifrfryyi67v2yzr2jfixxbljkhmk3ed6aw6cp6wljlx6igb6btioi?filename=Account%2520-%2520Web3%2520Storage%2520file%2520search.png)

I've been experimenting with integrating  web3.storage client example 
into IndyLab powered Inter Personal Knowledge Graph pilot
and extended it so that it 
get's the link for the uploaded file as a dweb link and displays it,
so that it can be added to documents.

![](https://bafkreib6y2rqx77z3ytsnpfstlcj36ipqsi67bhhzpe2u5t3mvber7re4m.ipfs.dweb.link/?filename=indylab%2520web3storage%2520experiment.png)

Before that I was able to get to the link via the listing, clicking the listed link and get the link for the saved file that way.

The current search interface only gives the CID for the upload, though.

it would be nice for people who do not know, or wish to know about CIDs
to give them  a dweb link to the actual file eventually including the filename as a search parameter as before

except for the files still being processed the search listing  have all the information
needed to provide a dweb link to the uploaded file

like so
https://bafkreib6y2rqx77z3ytsnpfstlcj36ipqsi67bhhzpe2u5t3mvber7re4m.ipfs.dweb.link/?filename=indylab%2520web3storage%2520experiment.png

dchoi27:
Thanks for flagging, tracking this in #1164 so closing this one!

PR #1243, opened 2022-04-12
feat: CID links in files table

### LINKED ISSUE #1164: Link to the content?
@dchoi27 edit for clarity: Currently, in the /account page of the website, the user's files table shows the CIDs of the file, but there are no links to actually viewing the content through a gateway.

Can we make the text in either the file name or CID column a link to `ipfs.io/ipfs/{cid}` for the given row?

cc @orvn 

---
Is there a link to my stored content? I feel like i'm missing it :(

orvn:
- CIDs that appear in the files table are now linked to a gateway URL
- All links open in a new tab

Closes #1164 and #1175

### DIFF
--- a/packages/website/components/account/filesManager/fileRowItem.js
+++ b/packages/website/components/account/filesManager/fileRowItem.js
@@ -127,7 +127,7 @@ const FileRowItem = props => {
         </span>
         <a
           className="cid-truncate underline medium-up-only"
-          href={`//ipfs.io/ipfs/${cid}`}
+          href={`https://dweb.link/ipfs/${cid}`}
           target="_blank"
           rel="noreferrer"
         >
--- a/packages/website/components/account/filesManager/fileRowItem.js
+++ b/packages/website/components/account/filesManager/fileRowItem.js
@@ -125,9 +125,14 @@ const FileRowItem = props => {
           <Tooltip content={fileRowLabels.cid.tooltip} />
           {fileRowLabels.cid.label}
         </span>
-        <span className="cid-truncate medium-up-only">
+        <a
+          className="cid-truncate underline medium-up-only"
+          href={`//ipfs.io/ipfs/${cid}`}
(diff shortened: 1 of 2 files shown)


dchoi27 (approved):
thanks! one point is that since things are wrapped in directories by default, it might be better to detect whether a file was uploaded with `wrapWithDirectory: true` and if so, have the link include the path to the file. but not sure if that's simple to detect - if not we can just merge.

alanshaw:
> thanks! one point is that since things are wrapped in directories by default, it might be better to detect whether a file was uploaded with `wrapWithDirectory: true` and if so, have the link include the path to the file. but not sure if that's simple to detect - if not we can just merge.

No you can't detect that from the CID alone. You'd need to download the block, decode it, inspect the entries, if only one then link to that item.

packages/website/components/account/filesManager/fileRowItem.js
```diff
@@ -125,9 +125,14 @@ const FileRowItem = props => {
           <Tooltip content={fileRowLabels.cid.tooltip} />
           {fileRowLabels.cid.label}
         </span>
-        <span className="cid-truncate medium-up-only">
+        <a
+          className="cid-truncate underline medium-up-only"
+          href={`//ipfs.io/ipfs/${cid}`}
```
olizilla, line 130: dweb.link is the preferred gateway domain these days, as it enforces origin isolation per cid by redirecting to a cid subdomain. It also deals with cidv0 to cidv1 rewriting as required.
```suggestion
          href={`https://dweb.link/ipfs/${cid}`}
```

olizilla (approved):
We can't easily do anything smarter than link to the CID on a gateway today, so let's get this PR merged so we at least have that link in the UI.

Just need to use dweb.link instead of ipfs.io and this is good to go

olizilla:
i'm gonna get this merged and deployed

PR #1247, opened 2022-04-12
fix: testimonial cards link behavior

svvimming:
A fix to update the index page testimonial cards with the correct behaviour outlined below:

If a card has a link in its description:

- Clicking anywhere on the card takes the user to that link (in a new tab)
- Hovering over the card shows the current glow effect
- The text that's linked should still be underscored (currently shown with a bottom border)

If the card does not have a link in its description:

- Hovering on the card does not produce the hover effect
- The card is not clickable

### DIFF
--- a/packages/website/components/card/card.js
+++ b/packages/website/components/card/card.js
@@ -67,10 +67,11 @@ export default function Card({ card, cardsGroup = [], index = 0, targetClass, on
 
   const navigateOnCardClick = useCallback(
     item => {
-      if (!card.cta && card.action === 'next-link') {
-        router.push(card.url)
+      if (!item.cta && item.action === 'next-link') {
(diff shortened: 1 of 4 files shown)


orvn (changes requested):
Looks good, but there's one issue where the `cursor: pointer` isn't working for some reason, maybe because the leaf nodes are being hovered on instead of the actual linked card

svvimming:
@orvn fixed now, it was actually a before element from the following section that was covering part of the cards

orvn:
@svvimming, I'm still seeing the issue? (tested on MacOS Firefox and Chrome)

orvn (approved):
Regression tested cross browser and this works in all instances.

Issue #851, opened 2022-01-10
Enforcing Web3 user storage capacity limits - first iteration


dchoi27:
The following describes the complete scope of the first iteration:

#### 1.  User notifications
_Notifying users who are close or over their storage limit (usually 1TB)_

- [X] Usage calculation and DB retrieval of all users > 75% usage of their quota
- [X] MailChimp ‘user quota reached’ email template with explanations on how to request more storage
  - [ ] To do: Adding tests for the MailChimp stuff & mock it
- [X] MailChimp ‘user quota exceeded’ email template with explanations on how to request more storage
  - [ ] To do: Adding tests for the MailChimp stuff & mock it
- [x] Daily automated (CRON) job to trigger to send email to all users 75-80%, 80-85%, 85-90%, 90-100%, 100%+ of their storage limit
  - [ ] To do: Uncommenting the PROD CRON job trigger - turned on for safety now

#### 2.  Admin notifications
_Notifying admins who are the recent heavy users and who is over the limit_

- [X] Usage calculation and DB retrieval of all users > 100% usage of their quota
- [x] MailChimp ‘user exceeded list’ email template
- [x] Daily automated (CRON) job to trigger to send 1 email listing all users 100%+ of their storage limit

#### 3.  Blocking upload access
_Admins blocking access to APIs for people over the limit._

- [x] Manual interface to set a restriction flag (admin.storage)
- [x] Implement filters on all APIs based on this flag

#### 4.  Request more storage
_Users requesting a higher storage limit with the reason_

- [x] Send an email to support@web3.storage
- [x] GitHub request

#### 5.  Grant more storage
_Admins granting a higher storage capacity to individual users_

- [x] Set a new user tag with storage in admin.storage

## Test plan

* Notifications
- [ ] Upload a 100MB file in Staging, adjust the user's storage limit and test the email notifications that come out.
* Blocking
- [ ] Verify if HasRestriction flag is being set in the database from admin.storage.
- [ ] See what happens with the pinning and upload APIs. User should not be able to upload anything and get an appropriate error message

JeffLowe:
Just to clarify @dchoi27 Is this referencing limits to a user's total allowed storage on their account, rate limiting a user's api, or something else?

PR #1248, opened 2022-04-14
feat: add github workflow cron job for storage limit email alerts

### LINKED ISSUE #1193: Storage limits - automate the email job
Related to #990

We need to automate the cron job that retrieves users who have gone over a certain treshold, and sends them an adapted email.

## Todo
- We need to figure out the best way to test this in staging and then release it to prod.

## Notes
- Setting this up as a manual action might help with testing (run it on staging).

adamalton:
Fixes #1193.

This adds the GitHub workflow configuration for running the cron job to send email notifications about users approaching or over their storage limits.

Notes:

* It's currently only configured for staging, because we want to test it there before adding it to production.
* I've configured it so that it can be run manually like [this](https://docs.github.com/en/github-ae@latest/actions/managing-workflow-runs/manually-running-a-workflow) (as well as running on its schedule).
* It requires the MailChimp API key to be added to the `MAILCHIMP_API_KEY` secret. I don't think I've got access to the right stuff in GitHub to be able to do this.

### DIFF
--- a/.github/workflows/cron-storage-limits.yaml
+++ b/.github/workflows/cron-storage-limits.yaml
@@ -37,9 +37,5 @@ jobs:
           STAGING_PG_REST_JWT: ${{ secrets.STAGING_PG_REST_JWT }}
           PROD_PG_REST_URL: ${{ secrets.PROD_PG_REST_URL }}
           STAGING_PG_REST_URL: ${{ secrets.STAGING_PG_REST_URL }}
-          CLUSTER_API_URL: ${{ secrets.CLUSTER_API_URL }}
-          CLUSTER_BASIC_AUTH_TOKEN: ${{ secrets.CLUSTER_BASIC_AUTH_TOKEN }}
-          CLUSTER_IPFS_PROXY_API_URL: ${{ secrets.CLUSTER_IPFS_PROXY_API_URL }}
-          CLUSTER_IPFS_PROXY_BASIC_AUTH_TOKEN: ${{ secrets.CLUSTER_IPFS_PROXY_BASIC_AUTH_TOKEN }}
           MAILCHIMP_API_KEY: ${{ secrets.MAILCHIMP_API_KEY }}
         run: npm run start:storage -w packages/cron
--- a/.github/workflows/cron-storage-limits.yaml
+++ b/.github/workflows/cron-storage-limits.yaml
@@ -1,0 +1,45 @@
+name: Storage Limit Email Notifications
+
+on:
+  schedule:
+    - cron: '0 5 * * *'
+  # Including 'workflow_dispatch' here allows the job to be triggered manually,
+  # as well as on the schedule.
+  workflow_dispatch:
+
+jobs:
+  send-notifications:
+    name: Send notifications
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # TODO: Update this to include production once it's been successfully
+        # tested on staging
+        # env: ['staging', 'production']
+        env: ['staging']
+    timeout-minutes: 100
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup node
+        uses: actions/setup-node@v1
+        with:
+          node-version: 16
+
+      - name: Install dependencies
+        uses: bahmutov/npm-install@v1
+
+      - name: Run job
+        env:
+          DEBUG: '*'
+          ENV: ${{ matrix.env }}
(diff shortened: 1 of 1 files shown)


.github/workflows/cron-storage-limits.yaml
```diff
@@ -0,0 +1,45 @@
...
+        # TODO: Update this to include production once it's been successfully
+        # tested on staging
+        # env: ['staging', 'production']
+        env: ['staging']
+    timeout-minutes: 100
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup node
+        uses: actions/setup-node@v1
```
olizilla, line 25: for consistency with our other workflows...
```suggestion
        uses: actions/setup-node@v2
```

.github/workflows/cron-storage-limits.yaml
```diff
@@ -0,0 +1,45 @@
...
+          DEBUG: '*'
+          ENV: ${{ matrix.env }}
+          PROD_PG_REST_JWT: ${{ secrets.PROD_PG_REST_JWT }}
+          STAGING_PG_REST_JWT: ${{ secrets.STAGING_PG_REST_JWT }}
+          PROD_PG_REST_URL: ${{ secrets.PROD_PG_REST_URL }}
+          STAGING_PG_REST_URL: ${{ secrets.STAGING_PG_REST_URL }}
+          CLUSTER_API_URL: ${{ secrets.CLUSTER_API_URL }}
+          CLUSTER_BASIC_AUTH_TOKEN: ${{ secrets.CLUSTER_BASIC_AUTH_TOKEN }}
+          CLUSTER_IPFS_PROXY_API_URL: ${{ secrets.CLUSTER_IPFS_PROXY_API_URL }}
+          CLUSTER_IPFS_PROXY_BASIC_AUTH_TOKEN: ${{ secrets.CLUSTER_IPFS_PROXY_BASIC_AUTH_TOKEN }}
```
olizilla, line 43: does it need credentials for all the things here?

olizilla (commented):
Looks good, but it's a little odd to see a PR that wants to merge into a branch but no associated PR for the target branch [feat/storage-quota-emails](https://github.com/web3-storage/web3.storage/tree/feat/storage-quota-emails), so I can't easily verify.

.github/workflows/cron-storage-limits.yaml
adamalton: Good point, I don't think it needs any of the cluster stuff. I'll remove that.

adamalton:
@olizilla I've now created a draft PR for the feature as a whole, #1273, which hopefully gives this some context!

.github/workflows/cron-storage-limits.yaml
```diff
@@ -0,0 +1,41 @@
+name: Storage Limit Email Notifications
+
+on:
+  schedule:
+    - cron: '0 5 * * *'
```
olizilla, line 5: just to check, this means run at 5am every day. Is the plan to just run this once or day or is it something we should run more regularly? Will it email a user that is at their limit every time it is run? Would a user that hits their limit at 5:01am only be informed 24hrs later?

https://crontab.guru/#0_5_*_*_*

adamalton: > just to check, this means run at 5am every day. 
Yep, kind of randomly chosen with the vague idea of not running _all_ our cron jobs at midnight and taking the platform down!

> Will it email a user that is at their limit every time it is run?
It will only email them once per day (or once per week, depending on the email type), so we could run the cron more often than this if we want to. Email frequency does not increase with cron frequency!

> Would a user that hits their limit at 5:01am only be informed 24hrs later?
Yes, and I think you raise a good point here. I don't really know what the typical pattern is for how (fast) a user tends to reach their limit, and therefore whether this is likely to be significant..?

> Is the plan to just run this once or day or is it something we should run more regularly?
Well, given the answers above, I think that comes down to what balance we want to strike between notifying users as quickly as possible vs. the potential strain it puts on the system (especially as we scale to more users). And maybe the unlikely possibility that there's a bug which causes us to spam everyone, so if we're feeling timid then we might want to deploy it set to once per day to start with and then crank it up. But given that we've got fairly good test coverage and we'll test it on staging before we stick it on prod, I'm not feeling timid!

How about every 6 hours?

flea89: Moving this conversation in the main first iteration PR #851 in order to have all the sub PRs merged and tidied up in one place.

PR #1252, opened 2022-04-17
feat: consolidate subnav account menu

svvimming:
Nav account dropdown changed to consolidate 'Dashboard' and 'Files' options to one item ('Manage Files') which navigates to the account page.


orvn (approved):
Looks good, just made sure that the case matches in all subnav items (there was a mix of lowercase and uppercase)

PR #1255, opened 2022-04-19
fix: return 400 on malformed form data

vasco-santos:
`HTTPError.throw(error.message, 400)` was not properly receiving status code, leading to non handling of https://sentry.io/organizations/protocol-labs-it/issues/3185331407/?project=5864302


olizilla (approved):
yikes. 👍

PR #1257, opened 2022-04-19
feat: storage quota admins email

### LINKED ISSUE #1189: Storage limit - Notify admins of users approaching/over storage limit
This is a an extension of #990, and depends on #1153, which is part of that.

We want to notify the Web3.Storage admins of the users who are approaching or over their storage limit.

- [ ]  We should do this as part of the `checkStorageUsed` function that's been added in #990.
- [ ] We only want to tell admins about users who have exceeded their storage quota (over 100%), we don't need to tell them about the other users.
- [ ] We should exclude any users whose accounts are restricted, as the admins don't need to worry about these. This has been done as #1194 
- [ ] Rather than send the admins a separate email for each individual user who has exceeded their limit, we should send a single email with a list of all the users who have exceeded their limit.
- [ ] For this we will use the `template_content` parameter to insert a block of content into a MailChimp template. See [API docs](https://mailchimp.com/developer/transactional/api/messages/send-using-message-template/).
- [ ] We should use the `EmailService` to send these emails.
- [ ] The admins probably don't want to receive this email more than once a day, so we could make use of the `daysSince` parameter of `EmailService.sendEmail` to ensure that admins don't get it more than once a day, even if we start running the cron job more frequently than that.
  - Note that for this we will want to set the `daysSince` to **less than** one day, because the cron job might be run daily but won't necessarily run on an exact schedule, so we probably want to set it to something like 23 hours. So we might want to change `daysSince` to `hoursSince`! NOTE: implemented as `secondsSinceLastSent`
- [ ] We'll need to create a new email type `AdminStorageExceeded` and a new template in MailChimp (and then use the "Send to Mandrill" option so that we can send it transactionally).
- [ ] We should probably call `sendEmail` with just the list of user details in `templateVars` parameter, and then leave it up to the `MailChimpEmailProvider` to decide how it wants to format that list into the template.

GaryHomewood:
Resolves #1189


adamalton:
Superseded by #1264

PR #1261, opened 2022-04-19
feat: Upload progress feedback

svvimming:
Animated count upward between chunked points in file upload progress. This count is displayed in the file uploader component next to uploading files to visually indicate to the user that the file is indeed uploading and not hanging.


svvimming:
The upload progress percentage has been pulled out into its own component called 'UploadProgress' found in the dropzone directory in the zero module. The dropzone calls in this component to handle two things: 

1. updating the percentage value at a set frame rate (currently 1 fps)
2. calculating a crude upload rate by looking at the interval of time and progress between upload chunks. The component then uses this upload rate to change the amount by which the progress percentage is incremented during animation.

orvn (approved):

Issue #1194, opened 2022-04-01
Exclude restricted accounts from quota usage notification emails


adamalton:
This is a follow-up to #990 (and so is dependent on it).

When we email users telling them that they are approaching or exceeding their storage quota, we should exclude accounts which are restricted (by checking the `user_tags` table).

The rationale is that if a user's account is restricted then there's no point in warning them about their usage level and that their account might become restricted. When their account becomes restricted that should trigger a separate email telling them about that (which is a separate task, not tied to this), and these warning emails are no longer relevant.

PR #1264, opened 2022-04-20
feat: admin emails for storage

### LINKED ISSUE #1189: Storage limit - Notify admins of users approaching/over storage limit

GaryHomewood:
Resolves #1189


packages/cron/test/storage.spec.js
```diff
@@ -56,6 +56,11 @@ describe('cron - check user storage quotas', () => {
   ]
 
   beforeEach(async () => {
```
flea89, line 58: A few thoughts on testing:
- It'd be nice to write tests to cover for the different scenarios on a more granular level. (ie. users that are restricted are not sent in the admin list)
- We should be testing the email client is called with the right inputs (stubbing it should be fine here)

packages/db/index.js
```diff
@@ -295,26 +316,26 @@ export class DBClient {
...
       emailType,
-      numberOfDays = 7
+      secondsSinceLastSent = 60 * 60 * 24 * 7
     } = email
 
-    const d = new Date()
-    d.setDate(d.getDate() - numberOfDays)
-    const numberOfDaysAgo = d.toISOString()
+    const now = new Date()
+    now.setSeconds(now.getSeconds() - secondsSinceLastSent)
```
flea89, line 330: nit.

Given your setting this to be back in the past, calling it `now` is a bit confusing.

packages/cron/src/lib/email.js
```diff
@@ -44,24 +44,24 @@ export class EmailService {
    * @param {object} opts
    * @param {import('@web3-storage/db/db-client-types').UserStorageUsedOutput} opts.user
    * @param {string} opts.email
-   * @param {number} opts.numberOfDays
-  * @returns void
+   * @param {number} opts.secondsSinceLastSent
+   * @returns void
    */
   async sendEmail ({
```
flea89, line 50: I wonder if we should be agreeing on the interface and how we could pass `context` to the email template here so that we could start creating it as part of the `PR`.

flea89 (changes requested):

packages/cron/src/jobs/storage.js
```diff
@@ -19,56 +47,30 @@ export async function checkStorageUsed ({ db }) {
...
-    users,
-    email: EMAIL_TYPE.Used90PercentStorage,
-    numberOfDays: 1
-  })
+    if (users.length) {
+      if (email.emailType === EMAIL_TYPE.User100PercentStorage) {
+        await emailService.sendAdminEmail({
+          users,
+          email: EMAIL_TYPE.AdminStorageExceeded,
+          secondsSinceLastSent: 60 * 60 * 23
```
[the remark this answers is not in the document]
GaryHomewood: Yes it should. Corrected as part of overall refactor.

packages/cron/src/jobs/storage.js
```diff
@@ -19,56 +47,30 @@ export async function checkStorageUsed ({ db }) {
...
-    fromPercent: 90,
-    toPercent: 100
-  })
-  await emailService.sendEmails({
-    users,
-    email: EMAIL_TYPE.Used90PercentStorage,
-    numberOfDays: 1
-  })
+    if (users.length) {
+      if (email.emailType === EMAIL_TYPE.User100PercentStorage) {
```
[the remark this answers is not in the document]
GaryHomewood: We did have a discussion about this, went with only sending admin email for storage quota exceeded.

packages/cron/src/lib/email.js
```diff
@@ -79,4 +79,42 @@ export class EmailService {
...
+
+  /**
+   * Send an email to admin with a list of all users that have exceeded their storage quota
+   * @param {object} opts
+   * @param {Array<import('@web3-storage/db/db-client-types').UserStorageUsedOutput>} opts.users
+   * @param {string} opts.email
+   * @param {number} opts.secondsSinceLastSent
+   * @returns void
+   */
+  async sendAdminEmail ({
```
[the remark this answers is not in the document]
GaryHomewood: Have refactored, email service now has just the one `sendEmail` method.

The `user` parameter has either:
- the ID of the user to be emailed
- the email address to be emailed (currently used for the admin email)

Have also moved logging out of the email service and into `storage.js` where it makes more sense anyway.

packages/db/index.js
GaryHomewood: True. Have renamed.

packages/cron/src/lib/email.js
```diff
@@ -79,4 +79,42 @@ export class EmailService {
...
+    users,
+    email,
+    secondsSinceLastSent
+  }) {
+    const adminUser = await this.db.getUserByEmail('admin@web3.storage')
```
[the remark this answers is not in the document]
GaryHomewood: Good point about the admin user, have added the user creation as a migration script so it doesn't get forgotten.

The `emailHasBeenSent` has been left as supporting just user IDs since we only have the one extra case for user by email address. Should be revisited in future.

packages/cron/src/jobs/storage.js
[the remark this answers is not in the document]
flea89: Agree, it should be enough! 
Would you mind updating the requirements in the ticket 😇 ?

packages/cron/test/storage.spec.js
GaryHomewood: - Test added to exclude restricted users as part of [#1194](https://github.com/web3-storage/web3.storage/issues/1194)
- Added test using `sinon` to ensure email client is called with the correct parameters.

packages/cron/src/jobs/storage.js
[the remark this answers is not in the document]
GaryHomewood: Have updated the ticket requirements #1189

packages/cron/src/jobs/storage.js
```diff
@@ -19,56 +47,30 @@ export async function checkStorageUsed ({ db }) {
...
-    toPercent: 80
-  })
-  await emailService.sendEmails({
-    users,
-    email: EMAIL_TYPE.Used75PercentStorage
-  })
+      await emailService.sendUserEmails({
+        users,
+        email: email.emailType,
+        ...email.secondsSinceLastSent && { secondsSinceLastSent: email.secondsSinceLastSent }
```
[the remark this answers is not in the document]
GaryHomewood: Agree, I have updated.

packages/cron/src/lib/email.js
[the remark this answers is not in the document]
flea89: Good idea creating the migration 👌

packages/cron/src/jobs/storage.js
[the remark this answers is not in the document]
flea89: Thanks!

packages/db/index.js
flea89: Ta!

packages/cron/src/jobs/storage.js
```diff
@@ -1,74 +1,99 @@
...
-    toPercent: 100
-  })
-  await emailService.sendEmails({
-    users,
-    email: EMAIL_TYPE.Used90PercentStorage,
-    numberOfDays: 1
-  })
+        const emailSent = await emailService.sendEmail(to, EMAIL_TYPE.AdminStorageExceeded, {
+          secondsSinceLastSent: email.secondsSinceLastSent,
+          templateVars: users
```
flea89, line 68:
```suggestion
          templateVars: { users }
```
I reckon `templateVars` should always be an object.

packages/cron/src/lib/email.js
```diff
@@ -16,67 +16,53 @@ export class EmailService {
...
-   * @param {import('@web3-storage/db/db-client-types').UserStorageUsedOutput} opts.user
-   * @param {string} opts.email
-   * @param {number} opts.numberOfDays
-  * @returns void
+   * @param {{id?: number, email: string, name: string}} user
+   * @param {string} emailType
+   * @param {object} [options]
+   * @param {number} [options.secondsSinceLastSent]
+   * @param {boolean} [options.failSilently]
+   * @param {object} [options.templateVars]
```
flea89, line 28:
```suggestion
   * @param {Object.<string, any>} [options.templateVars]
```

packages/cron/package.json
```diff
@@ -33,6 +33,7 @@
     "@types/node": "^16.3.1",
     "execa": "^5.1.1",
     "mocha": "^8.3.2",
-    "npm-run-all": "^4.1.5"
+    "npm-run-all": "^4.1.5",
+    "sinon": "^13.0.2"
```
flea89, line 37: Nice, we should make sure restore the sinon sandbox at each test as stated in the docs [here](https://sinonjs.org/releases/v13/general-setup/).

[Here](https://github.com/web3-storage/web3.storage/pull/1196/commits/372d9b487b2ea2317216f48203a1704bc5a4a5fd#diff-60767f1efe42b8a4a16c4fdd5283cd62c2572192bbb11aeb1196c0bb11ed062fR61)'s an example

packages/cron/test/storage.spec.js
flea89: Nice! I wonder if we could we write the test in a bit more declarative way. At the moment there's a lot of boilerplate to create the users, uploads etc and it hard to figure out what are the actual numbers per user etc.

Can it be simplified by having a function to create a user with a specific amount of bytes stored?
And define users/bytes as a array of objects?

packages/cron/package.json
GaryHomewood: Added to `hooks.js`

packages/cron/test/storage.spec.js
GaryHomewood: Added a helper method for the user with files data setup, easier to read in the `beforeAll` now

packages/cron/src/jobs/storage.js
```diff
@@ -1,74 +1,99 @@
...
+    emailType: EMAIL_TYPE.User90PercentStorage,
+    fromPercent: 90,
+    toPercent: 100,
+    secondsSinceLastSent: 60 * 60 * 23 // 1 day approx
+  },
+  {
+    emailType: EMAIL_TYPE.User85PercentStorage,
+    fromPercent: 85,
+    toPercent: 90,
+    secondsSinceLastSent: 60 * 60 * 24 * 7 // 1 week
```
adamalton, line 22:
```suggestion
    secondsSinceLastSent: 60 * 60 * 24 * 7 - 60 * 60 // 1 week, with flexibility for cron irregularity
```

I think that all of these need the ~1 hour flexibility on them. Otherwise, if the cron doesn't run exactly daily then these emails will sometimes get sent every 8 days rather than every 7.

packages/cron/src/jobs/storage.js
```diff
@@ -1,74 +1,99 @@
...
+  },
+  {
+    emailType: EMAIL_TYPE.User80PercentStorage,
+    fromPercent: 80,
+    toPercent: 85,
+    secondsSinceLastSent: 60 * 60 * 24 * 7 // 1 week
```
adamalton, line 28:
```suggestion
    // 1 week, with flexibility for cron irregularity
    secondsSinceLastSent: 60 * 60 * 24 * 7 - 60 * 60
```

packages/cron/src/jobs/storage.js
```diff
@@ -1,74 +1,99 @@
...
+  },
+  {
+    emailType: EMAIL_TYPE.User75PercentStorage,
+    fromPercent: 75,
+    toPercent: 80,
+    secondsSinceLastSent: 60 * 60 * 24 * 7 // 1 week
```
adamalton, line 34:
```suggestion
    secondsSinceLastSent: 60 * 60 * 24 * 7 // 1 week
```

packages/cron/src/jobs/storage.js
GaryHomewood: Have added the flex throughout.

packages/cron/src/jobs/storage.js
GaryHomewood: Ok done.

packages/cron/src/jobs/storage.js
GaryHomewood: Also done here.

PR #1265, opened 2022-04-20
feat: ignore restricted users for storage emails

### LINKED ISSUE #1194: Exclude restricted accounts from quota usage notification emails

GaryHomewood:
Resolves #1194

### DIFF
--- a/packages/db/test/user-storage.spec.js
+++ b/packages/db/test/user-storage.spec.js
@@ -6,7 +6,8 @@ import {
   createUserAuthKey,
   createUpload,
   createPsaPinRequest,
-  token
+  token,
+  randomCid
 } from './utils.js'
 
 describe('Users used storage', () => {
@@ -17,17 +18,9 @@ describe('Users used storage', () => {
     postgres: true
   })
 
-  let user1, user2, user3, user4
-  let authKey1, authKey2, authKey3, authKey4
-  const cids = [
-    'bafybeiczsscdsbs7ffqz55asqdf3smv6klcw3gofszvwlyarci47fgf111',
-    'bafybeiczsscdsbs7ffqz55asqdf3smv6klcw3gofszvwlyarci47fgf112',
-    'bafybeiczsscdsbs7ffqz55asqdf3smv6klcw3gofszvwlyarci47fgf113'
-  ]
-  const cidsPinned = [
-    'QmdA5WkDNALetBn4iFeSepHjdLGJdxPBwZyY47ir1bZGAK',
-    'QmNvTjdqEPjZVWCvRWsFJA1vK7TTw1g9JP6we1WBJTRADM'
-  ]
+  let user1, user2, user3, user4, user5
+  let authKey1, authKey2, authKey3, authKey4, authKey5
+  let cid
   const uploadSize = 439804651110
   const pinnedSize = 109951162800
   const largeFileSize = 3199023255550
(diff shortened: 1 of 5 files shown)


packages/db/test/user-storage.spec.js
```diff
@@ -113,17 +114,42 @@ describe('Users used storage', () => {
...
+      dagSize: largeFileSize
+    })
+
     // User               | upload  | pinned  | quota   | percentage of quota
     // ----------------------------------------------------------------------
     // test1@email.com    | 1       | 2       | 1TiB    | < 75%
     // test2@email.com    | 2       | 0       | 1TiB    | 80%
     // test3@email.com    | 2       | 1       | 1TiB    | 90%
     // test4@email.com    | 1 (XL)  | 0       | 2TiB    | > 90%
+    // test5@email.com    | 1 (XL)  | 0       | 2TiB    | ACCOUNT RESTRICTED
```
adamalton, line 152: Should there be a test added for the expected result of `getUsersByStorageUsed` with respect to user5?

packages/db/test/utils.js
```diff
@@ -119,3 +122,13 @@ export async function getUpload (dbClient, cid, userId) {
...
+
+/**
+ * @param {number} code
+ * @returns {Promise<string>}
+ */
+export async function randomCid (code = pb.code) {
+  const bytes = crypto.getRandomValues(new Uint8Array(10))
+  const hash = await sha256.digest(bytes)
+  return CID.create(1, code, hash).toString()
+}
```
adamalton, line 134: Minor thing, but we've already got a `randomCid` function in `db/test/pinning.spec.js`. So maybe that one can be removed and replaced with an import of this one?

Also, given that it's just for testing, and therefore doesn't need to be _actually_ random in a secure way, it could just use `Math.random()` which would be faster than the cryptographically-secure `crypto.getRandomValues`. Every little helps! 🤷  See [randomCid in api package](https://github.com/web3-storage/web3.storage/blob/main/packages/api/test/scripts/helpers.js#L62-L65).

packages/db/test/user-storage.spec.js
GaryHomewood: Good point, have added an explicit test.

packages/db/test/utils.js
GaryHomewood: Amended, and refactored `pinning-spec.js` to use the same.

packages/db/test/user-storage.spec.js
```diff
@@ -113,25 +114,50 @@ describe('Users used storage', () => {
...
   })
 
-  it('returns user details needed for email', async () => {
+  it.only('returns user details needed for email', async () => {
```
adamalton, line 155: Accidental change?

GaryHomewood: Yes, have removed.

packages/db/test/utils.js
adamalton: Sweet

PR #1268, opened 2022-04-22
chore: rename pinned to psaPinned

### LINKED ISSUE #1197: Rename users pinned content as psa_content
The amount of storage used for content that has been pinned by a user is currently referred to in the codebase as `pinned` (in `db/db-client-types.ts`). This is slightly confusing since the users uploaded content is also pinned to a cluster. 

We should rename throughout to `psa_content` so that the distinction is clear.

flea89:
resolves #1197 

🛑 🛑 🛑  
- Requires migration `packages/db/postgres/migrations/008-user-used-storage-update.sql`
- Requires to deploy API and Website.


alanshaw:
Do we need this one? It's not user facing and soon data will not actually be "pinned" in Cluster anymore so I think the distinction is sufficient.

flea89:
@alanshaw I'm not too precious about this but here's the rationale:

- Despite being not user-facing, "pinned" is a source of confusion for basically every new developer joining the project. So despite this not having an impact on users it does have an impact on us.
- we already prefix with `psa_` every Pinning Service Api-related data structure/method. That is and will be true despite the change to the cluster.
- I see your point about the change in the future, but when is that going to be actually released? Despite that `pinned` appears 507 times in 52 files, it might not go from everywhere 🤷 ?


That said, I'm not too opinionated.
Happy to drop it if you think those are not strong enough reasons.

packages/db/postgres/migrations/006-user-used-storage-update.sql
```diff
@@ -0,0 +1,39 @@
+DROP TYPE used_storage
```
alanshaw, line 1:
```suggestion
DROP TYPE used_storage;
```

packages/db/postgres/migrations/006-user-used-storage-update.sql
```diff
@@ -0,0 +1,39 @@
...
+CREATE TYPE used_storage AS (uploaded TEXT, psa_pinned TEXT);
+
+DROP FUNCTION user_used_storage(bigint);
+
+CREATE OR REPLACE FUNCTION user_used_storage(query_user_id BIGINT) 
+RETURNS used_storage
+LANGUAGE plpgsql
+AS
+$$
+DECLARE used_storage used_storage;
```
alanshaw, line 11: So, I'd either rename this variable to something different to the type or rename the type to be something more generic like `stored_bytes`.

alanshaw (changes requested):

packages/db/postgres/migrations/006-user-used-storage-update.sql
flea89: Good point. 
Updated

alanshaw (approved):

PR #1269, opened 2022-04-22
fix: Bump MAX_BLOCK_SIZE to 2MiB limit imposed by bitswap

TheFrozenFire:
go-ipfs's chunker produces blocks containing 1MiB of data, however the protobuf wrapping each block results in ~0.001MiB of overhead. This means that when uploading CAR files containing blocks that go-ipfs produces, sometimes there are well-formed blocks that exceed the current limit. I am proposing bumping the max block size accepted by web3.storage to the limit imposed by ipfs bitswap, which seems like it's more in line with the reasoning behind imposing a block size limit.

The 2MiB value selected is proposed by @Jorropo in the ipfs Discord, who claims that it's a part of the bitswap spec. This value is also mentioned in ipfs/go-ipfs#3104

### DIFF
--- a/packages/api/src/constants.js
+++ b/packages/api/src/constants.js
@@ -2,7 +2,7 @@ export const JWT_ISSUER = 'web3-storage'
 export const METRICS_CACHE_MAX_AGE = 10 * 60 // in seconds (10 minutes)
 export const DAG_SIZE_CALC_LIMIT = 1024 * 1024 * 9
 // Maximum permitted block size in bytes.
-export const MAX_BLOCK_SIZE = 1 << 20 // 1MiB
+export const MAX_BLOCK_SIZE = 1 << 21 // 2MiB
 export const UPLOAD_TYPES = ['Car', 'Blob', 'Multipart', 'Upload']
 export const PIN_STATUSES = ['PinQueued', 'Pinning', 'Pinned', 'PinError']
 export const USER_TAGS = {


Jorropo (approved):

alanshaw:
@TheFrozenFire the go-ipfs chunker produces blocks of size 262,144 bytes by default. I don't understand what the issue is here?

aschmahmann:
Not sure the exact context here, but for those looking for pointers around sizes:
- The Bitswap spec is being clarified that it should tolerate blocks up to 2MiB (https://github.com/ipfs/specs/pull/269)
- go-ipfs-chunker (and go-ipfs's `ipfs add`) has a max size of 1048576 (1MiB)  https://github.com/ipfs/go-ipfs-chunker/blob/468aeb15c71364467c7a01172944917776d4c241/parse.go#L18
- However, with protobuf wrapping for those not using raw leaves the block sizes can then slightly exceed 1MiB 😭. See more in the discussion around setting the chunk size limit https://github.com/ipfs/go-ipfs-chunker/pull/21#discussion_r369124879

olizilla (approved):
given
>  I've come across a number of blocks which are slightly greater in size than the current limit of web3.storage
> – @TheFrozenFire 

and
> with protobuf wrapping for those not using raw leaves the block sizes can then slightly exceed 1MiB 😭
> – @aschmahmann 

it sounds like we should increase the limit.

and with
> The Bitswap spec is being clarified that it should tolerate blocks up to 2MiB...

then this seems like the right fix! Thanks @TheFrozenFire

alanshaw (approved):

olizilla:
Tests pass locally.

PR #1270, opened 2022-04-25
chore: add log tail logging to API

### LINKED ISSUE #922: Add logtail integration

adamalton:
Fixes #922

Adds Logtail logging to the API.

Essentially a port of [nft 860](https://github.com/nftstorage/nft.storage/pull/860)


adamalton:
@olizilla are you able to help us get the `LOGTAIL_TOKEN` added to the Cloudflare Worker env for the web3.storage API, in anticipation of this being released? There's also a question of whether we want to use a different account/token for w3 vs. nft, or whether we want to just separate them by filtering in Logtail..?

packages/api/src/env.js
```diff
@@ -94,6 +98,20 @@ export function envAll (req, env, ctx) {
     release: env.SENTRY_RELEASE,
     pkg
   })
+
+  // Attach a `Logging` instance, which provides methods for logging and writes
+  // the logs to LogTail. This must be a new instance per request.
+  // Note that we pass `ctx` as the `event` param here, because that's what it
+  // actually is, we seem to have just misnamed it "ctx".
+  env.log = new Logging(req, ctx, {
```
flea89, line 106: Can we add log to the env type definition?

packages/api/src/index.js
```diff
@@ -109,11 +109,13 @@ function serverError (error, request, env) {
...
   async fetch (request, env, ctx) {
+    let response
     try {
-      return await router.handle(request, env, ctx)
+      response = await router.handle(request, env, ctx)
     } catch (error) {
-      return serverError(error, request, env)
+      response = serverError(error, request, env)
     }
+    return env.log.end(response)
```
flea89, line 118: As soon as I saw this the first thought I had was: why are we returning the logs here.
Obviously that's not the case, but still feels a bit un-natural `log.end` takes a response, mutates it and returns it.

Not a big deal and possibly not something for this PR, but I'd love to hear what you think?

packages/api/src/auth.js
```diff
@@ -30,7 +30,8 @@ export function withMagicToken (handler) {
     if (magicUser) {
       const userTags = await getUserTags(magicUser._id, env)
       request.auth = { user: magicUser, userTags }
-      env.sentry && env.sentry.setUser(magicUser)
+      // If env.log is not set, then the middlewares may be being run in the wrong order
+      env.log.setUser({ id: magicUser._id })
```
flea89, line 34: Another thought that probably goes beyond this PR.
Setting the user is duplicated acroos different middlewares and I wonder if it could live in its own?

flea89 (commented):
This looks great, I haven't digged deep into the logger code given it seems to be a straigh port from nft.

Left some minor comments.

vasco-santos (commented):
Overall looking good!

> help us get the LOGTAIL_TOKEN added to the Cloudflare Worker env for the web3.storage API, in anticipation of this being released? There's also a question of whether we want to use a different account/token for w3 vs. nft, or whether we want to just separate them by filtering in Logtail..?

I added the token to the repo secrets :) We have a logtail team for nft and other for web3. We already have other logtail sources for web3, like pogrest. So, just created a new source for the API specific to web3

packages/api/src/env.js
olizilla: @adamalton a todo!

packages/api/src/env.js
```diff
@@ -94,6 +98,20 @@ export function envAll (req, env, ctx) {
     release: env.SENTRY_RELEASE,
     pkg
   })
+
+  // Attach a `Logging` instance, which provides methods for logging and writes
+  // the logs to LogTail. This must be a new instance per request.
+  // Note that we pass `ctx` as the `event` param here, because that's what it
+  // actually is, we seem to have just misnamed it "ctx".
```
olizilla, line 105: It's nuanced. The `ctx` is "the object that has the `.waitUntil()` and `.passThroughOnException()` methods on it". 
Screenshot 2022-05-13 at 13 20 43
– https://developers.cloudflare.com/workers/runtime-apis/fetch-event/#parameters

> While an incoming HTTP request is still given the "fetch" name, a Module Worker does not surface the FetchEvent interface. Instead, Module Workers receive the [Request](https://developers.cloudflare.com/workers/runtime-apis/request/) and must reply with a [Response](https://developers.cloudflare.com/workers/runtime-apis/response/) directly.

in practice the worker recieves a `ctx` param that is in fact just the FetchEvent... IT'S BOTH!

olizilla (approved):
Looks good. Minor todos noted inline. LETS GET IT MERGED!

packages/api/src/env.js
adamalton: Ah, this is why I was slightly perplexed by it! I'm updating the comment. Thanks.

packages/api/src/auth.js
adamalton: I guess we could make it so that we always run the auth middleware first, and then the logging middleware afterwards, and then the logging middleware can know that the auth has already been done and so it can just look at `request.auth.user` to get the user. That way we wouldn't have to have the auth stuff calling the `setUser` method. But if we did it that way then:

* The auth middlewares wouldn't be able to do any logging (because the logging wouldn't yet be set up).
* We wouldn't be able to do `router.all('*', envAll)`, because we'd need that env stuff (or at least the logging part of it) to be run _after_ the auth middleware(s). So then I think our routing stuff would get quite complicated.

Maybe there's a third way, but having now thought that through, it makes me think that the current way is the least evil!

packages/api/src/env.js
adamalton: Good spot, thanks. Done.

packages/api/src/index.js
adamalton: Yeah, I agree that it seems a bit odd. Are you thinking that it would make more sense to do:

```javascript
await env.log.end(response)
return response
```

(The `end` method does this: `response.headers.set('Server-Timing', this._timersString())` so we need to `await` it before we return the response.)

flea89: Not sure what's best here, I guess my problem is mostly about `log.end` mutating the response itself.
 But to be honest, it is not a big deal.
So either what we have or what we're suggesting is ok. happy for you to go with what you prefer.

packages/api/src/auth.js
adamalton: For the record, we discussed this on Slack and concluded that it wasn't worth further discussion time for something so minor.

packages/api/src/index.js
adamalton: Ok, cool. I've gone for the slight tweak, because I agree that previously it looked like we were returning the logs!

PR #1271, opened 2022-04-25
chore: refactor auth middleware

### LINKED ISSUE #1162: Improve API auth middleware
Follow-up from the [account](https://github.com/web3-storage/web3.storage/pull/1053#discussion_r828323406) [restriction](https://github.com/web3-storage/web3.storage/pull/1053#pullrequestreview-918642375) PR

As we add more restrictions and requirements for API auth, we need a better way to combine middlewares and improve readability.

flea89:
Resolves #1162 

### First [comment](https://github.com/web3-storage/web3.storage/pull/1053#discussion_r828323406) in origin PR
As far as I can tell the implementation in `main` do not suffer from the problem of multiple requests to the db, given the `UserTags` are read once when fetching the user. That was already been revisited.

### Second [comment](https://github.com/web3-storage/web3.storage/pull/1053#pullrequestreview-918642375)
I do agree having a list of middlewares rather than composing the functions would make it more readable.
While refactoring it though, I've realised that it comes with some limitations, and in particular, it's less flexible when it comes to having the ability to run middlewares to modify the response, like in [addCorsHeaders](https://github.com/web3-storage/web3.storage/blob/main/packages/api/src/cors.js#L60).

While there are ways where we could still be able to use this approach while being able to control the order of the middlewares, I couldn't see a straightforward way that wouldn't require refactor/overengineer quite sensibly the way we define our routes and middlewares.

For that reason I instead just utility to allow us to "compose" our middlewares in a "linea" way.

### DIFF
--- a/packages/api/src/index.js
+++ b/packages/api/src/index.js
@@ -23,12 +23,27 @@ const router = Router()
 router.options('*', corsOptions)
 router.all('*', envAll)
 
+/**
+ * Takes any number of functions and invokes them all one after the other
+ *
+ * @param  {...function(...any):any} fns
+ * @returns
+ */
+const compose = (...fns) => x => fns.reduceRight((y, f) => f(y), x)
+
+/**
+ * It defines a list of "middlewares" that need to be applied for a given authentication mode.
+ * Each value takes an endpoint handler and returns a "composed" version.
+ *
+ * @type {Object.<string, function(...any):any>}
+ */
 const auth = {
-  '🤲': handler => withCorsHeaders(handler),
-  '🔒': handler => withCorsHeaders(withApiOrMagicToken(handler)),
-  '👮': handler => withCorsHeaders(withMagicToken(handler)),
-  '📌': handler => withPinningAuthorized(handler),
-  '🚫': handler => withAccountNotRestricted(handler)
+  '🤲': withCorsHeaders,
+  '🔒': compose(withCorsHeaders, withApiOrMagicToken),
+  '👮': compose(withCorsHeaders, withMagicToken),
+  '🚫': compose(withCorsHeaders, withApiOrMagicToken, withAccountNotRestricted),
+  '👀📌': compose(withCorsHeaders, withApiOrMagicToken, withPinningAuthorized),
+  '📝📌': compose(withCorsHeaders, withApiOrMagicToken, withAccountNotRestricted, withPinningAuthorized)
 }
 
 const mode = {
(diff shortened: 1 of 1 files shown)


olizilla (commented):
I think we can simplify things a little by pulling the `mode` / is read vs read-write checks up and linking directly to the http verbs:

```js
router.get('*', mode('r'))
router.head('*', mode('r'))
router.post('*', mode('rw'))
router.delete('*', mode('rw'))
```
we want to use the right verbs for the actions that mutate vs those that dont, so i am fine with removing flexibility there. see: https://github.com/kwhitley/itty-router#multi-route-upstream-middleware for "multi-route" middleware support example.

packages/api/src/index.js
```diff
@@ -23,12 +23,27 @@ const router = Router()
...
-  '🔒': handler => withCorsHeaders(withApiOrMagicToken(handler)),
-  '👮': handler => withCorsHeaders(withMagicToken(handler)),
-  '📌': handler => withPinningAuthorized(handler),
-  '🚫': handler => withAccountNotRestricted(handler)
+  '🤲': withCorsHeaders,
+  '🔒': compose(withCorsHeaders, withApiOrMagicToken),
+  '👮': compose(withCorsHeaders, withMagicToken),
+  '🚫': compose(withCorsHeaders, withApiOrMagicToken, withAccountNotRestricted),
+  '👀📌': compose(withCorsHeaders, withApiOrMagicToken, withPinningAuthorized),
+  '📝📌': compose(withCorsHeaders, withApiOrMagicToken, withAccountNotRestricted, withPinningAuthorized)
```
olizilla, line 46: Let's revisit these icons! I don't think they are making things clearer

- routes that _allow_ a *blocked* user to do a thing are definitly a _special case_ and so should visually stand out.

```suggestion
  '🌍': withCorsHeaders,                                        // world readable!                      
  '🔑⚠️': compose(withCorsHeaders, withApiOrMagicToken),        // any key will do & blocked users allowed!
  '👤': compose(withCorsHeaders, withMagicToken),               // must be a logged in user
  '🔑': compose(withCorsHeaders, withApiOrMagicToken, withAccountNotRestricted), // any key will do.

  '📌⚠️': compose(withCorsHeaders, withApiOrMagicToken, withPinningAuthorized), // needs PSA & blocked users allowed
  '📌': compose(withCorsHeaders, withApiOrMagicToken, withAccountNotRestricted, withPinningAuthorized)  // needs PSA
```

flea89 (commented):
Forgot to submit these comments in the first place.

olizilla:
@flea89 let's jump on a call? Could you update this PR to show the change that you'd **really** like to see? Let's refactor it good, rather than just less odd. We can switch it to words if that helps, tho i think we should simplify as well.

packages/api/src/index.js
flea89: Thanks, this is better. 
I still have a feeling this might be not obvious to someone new, but definitively a great improv from what we had.

Thanks Olly!

packages/api/src/index.js
```diff
@@ -37,37 +52,37 @@ const mode = {
...
-
-router.delete('/user/uploads/:cid',      mode['📝'](auth['👮'](userUploadsDelete)))
-router.post('/user/uploads/:cid/rename', mode['📝'](auth['👮'](userUploadsRename)))
-router.get('/user/tokens',               mode['👀'](auth['👮'](userTokensGet)))
-router.post('/user/tokens',              mode['📝'](auth['👮'](userTokensPost)))
-router.delete('/user/tokens/:id',        mode['📝'](auth['👮'](userTokensDelete)))
-router.get('/user/account',              mode['👀'](auth['👮'](userAccountGet)))
-router.get('/user/info',                 mode['👀'](auth['👮'](userInfoGet)))
+router.post('/user/login',          compose(mode['📝'], auth['🤲'])(userLoginPost))
+router.get('/status/:cid',          compose(mode['📝'], auth['🤲'])(statusGet))
```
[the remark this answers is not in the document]
flea89: Moved to using the multi-route approach, making this not relevant anymore.

packages/api/src/index.js
```diff
@@ -23,12 +23,27 @@ const router = Router()
...
-  '🤲': handler => withCorsHeaders(handler),
-  '🔒': handler => withCorsHeaders(withApiOrMagicToken(handler)),
-  '👮': handler => withCorsHeaders(withMagicToken(handler)),
-  '📌': handler => withPinningAuthorized(handler),
-  '🚫': handler => withAccountNotRestricted(handler)
+  '🤲': withCorsHeaders,
+  '🔒': compose(withCorsHeaders, withApiOrMagicToken),
+  '👮': compose(withCorsHeaders, withMagicToken),
+  '🚫': compose(withCorsHeaders, withApiOrMagicToken, withAccountNotRestricted),
+  '👀📌': compose(withCorsHeaders, withApiOrMagicToken, withPinningAuthorized),
```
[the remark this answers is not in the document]
flea89: Went for Olly's new icons! That makes this clearer.

packages/api/test/maintenance.spec.js
```diff
@@ -1,6 +1,5 @@
 /* eslint-env mocha */
```
flea89, line 1: I wonder if we should add some more integration-y tests, to make sure we test the modes are set up for the different verbs?
At the moment we are just testing the middleware itself

olizilla: yes it would be good to have at least one test that checks that the `*` style middlewear are actually being applied correctly, as I we've not used them before.

packages/api/src/index.js
```diff
@@ -23,54 +23,80 @@ const router = Router()
...
+router.post('*', withMode(READ_WRITE))
+router.delete('*', withMode(READ_WRITE))
+
+/**
+ * Takes any number of functions and invokes them all one after the other
+ *
+ * @param  {...function(...any):any} fns
+ * @returns
+ */
+const compose = (...fns) => x => fns.reduceRight((y, f) => f(y), x)
```
olizilla, line 37: minor: extract out to a utils/fn.js or similar. It's a one liner, but it distracts from the main purpose of this file – to init the thing.

olizilla:
@flea89 I'm pretty happy to land this as is. If you can add a test to show that the read/write modes are being applied correctly I would be even more happy, but if it's not trivial then we can check it manually this time.

olizilla:
@flea89 I'm pretty happy to land this as is. If you can add a test to show that the read/write modes are being applied correctly I would be even more happy, but if it's not trivial then we can check it manually this time.

packages/api/src/index.js
flea89: Good idea!

flea89:
Hey @olizilla, sorry I was working on something else. Thanks for following up on this.
Yeah, it's not exactly straightforward to test (or at least I can't think of a super simple way to do it). 
So given we have some other priorities now, I'd suggest tracking this on a [follow-up issue](https://github.com/web3-storage/web3.storage/issues/1297).

packages/api/test/maintenance.spec.js
flea89: Tracked in a separate [issue](https://github.com/web3-storage/web3.storage/issues/1297)

packages/api/src/index.js
```diff
@@ -18,59 +18,78 @@ import {
...
+router.head('*', withMode(READ_ONLY))
+router.post('*', withMode(READ_WRITE))
+router.delete('*', withMode(READ_WRITE))
+
+/**
+ * It defines a list of "middlewares" that need to be applied for a given authentication mode.
+ * Each value takes an endpoint handler and returns a "composed" version.
+ *
+ * @type {Object.<string, function(...any):any>}
+ */
```
olizilla, line 37: orphaned comment, plz delete
```suggestion
```

packages/api/src/maintenance.js
```diff
@@ -33,11 +33,10 @@ export const DEFAULT_MODE = READ_WRITE
  * r- = only needs read permission so enabled in read-only AND read+write modes.
  * rw = needs to read and write so only enabled in read+write mode.
  *
```
olizilla, line 35: Please add a comment here that this works by doing nothing if not enabled... as it's expected to be used from as an early catch-all middleware like `router.get('*', ...`

olizilla (approved):
minor tweaks, otherwise LETS GOOOOO!

packages/api/src/index.js
flea89: @olizilla It might not be the best comment/annotation, but it is still relevant for the auth map, isn't it?

olizilla: oh sorry, i was skimming and i thought this was the comment for compose function!

packages/api/src/maintenance.js
olizilla: This PR changes this handler so that it either does nothing or calls the maintenance handler. The suggestion was to add a comment about that, and explain that it's ok for this handler to do nothing, as it's expected to be run as middleware. But we all know how middleware works and it's super simple, so we can leave it. LETS GET THIS MERGED (I am assuming you will merge it when you are ready)

PR #1273, opened 2022-04-25
feat: send email notifications for storage quota usage

### LINKED ISSUE #851: Enforcing Web3 user storage capacity limits - first iteration

### LINKED ISSUE #990: Storage limit - Framework to get users and record emails sent

adamalton:
Fixes #851. 

This is an overall feature PR for #851, bringing together the work done in these sub-tasks:
* #1156 (which implements the core of this feature)
* #1264 (which implements #1189)
* #1282 (which implements #1153)
* #1248 (which implements #1193)

Now that those sub-tasks are done and merged to this branch, this is ready to be reviewed and merged as a complete feature. Note that potential follow-up improvements for this feature are listed in #1283.

Some of the requirements in the #851, are already in main or are part of [admin](https://github.com/nftstorage/admin.storage).

Specifically:

- Manual interface to set a restriction flag (in admin.storage)
- Set a new user tag with storage in admin.storage
- Implement restriction on all APIs/interface

### Outstanding discussion

There's an outstanding question of how often we want to run the cron job. Running it more frequently won't send the emails more frequently (each email type has its own frequency limit, so it will only be sent as often as necessary), but if we only run the cron every 24 hours then there's potentially up to a 24 hour delay between a user hitting a threshold and them receiving an email about it. If that delay is too long then we should run the cron more frequently. That needs to be balanced with not hammering the platform/DB too much (although I don't foresee a real issue here). My proposal is to increase the frequency to every 6 hours.
Original comment [here](https://github.com/web3-storage/web3.storage/pull/1248#discussion_r870439427)

### Oustanding tasks

* [ ] We need the API key from the Mailchimp account adding to the GitHub secrets as `MAILCHIMP_API_KEY`. I don't think any of us (Potato) have got access to the right stuff in GitHub to be able to do this.
* [ ] Testing this on staging before we deploy to production. (See below.)
* [ ] Enabling the cron job for production as well as staging in the GitHub workflow config file and update cron schedule to run every 6 hours.

### Testing this

Before we release this to prod, we should test it on staging. The plan for this is:

1. Deploy to staging. (We can actually deploy to prod as well if we want, as the cron job config is only set up to run it on staging at the moment - see above.)
2. Create some user accounts for ourselves and upload/pin some files.
3. Use admin.storage to set our quotas to something low enough to reach the 75%/80%/etc limit.
4. Run the cron job manually via the GitHub UI (see [instructions](https://docs.github.com/en/github-ae@latest/actions/managing-workflow-runs/manually-running-a-workflow)).
5. Reduce our quotas again so that we go over the limit.
6. Manually trigger the cron again.
7. Check that we’ve received the right emails and that the list of users over their quota has been sent to support@web3.storage.


packages/cron/src/lib/email/providers/mailchimp.js
```diff
@@ -0,0 +1,99 @@
...
+        merge_language: 'handlebars'
+        // There are various other options we can add in here, relating to tracking, IP address, etc
+        // https://mailchimp.com/developer/transactional/api/messages/send-using-message-template/
+      }
+    }
+    // It appears that the API returns a different type of response depending on the success or not.
+    // If there's an overall problem with the request, the response is an object, but if it's
+    // successful or if there's a problem with the message-level part of the request, it gives an
+    // array of responses, one for each recipient.
+    let response = await this.mailchimpTx.messages.sendTemplate(messageDef)
```
hugomrdias, line 63: can't we just use fetch here to call the api instead of requiring the full SDK, it seems to be just this code https://github.com/mailchimp/mailchimp-transactional-node/blob/master/src/ApiClient.js#L70-L104 running

hugomrdias (changes requested):
LGTM just left one inline comment

im good with cron running every 6 hours.

Are we just missing an Mailchimp key to test this in staging ?

hugomrdias (approved):

PR #1278, opened 2022-04-27
fix: only count pinned uploads for storage used

### LINKED ISSUE #1198: Users uploaded storage used should only count content that is pinned
In `db/postgres/functions.sql`where the users storage used is calculated for uploaded content, we are currently counting uploads that may never be actually pinned.

We should include a check to only include content that has a status of `Pinned`

GaryHomewood:
Resolves #1198


flea89:
Superseeded by #1430

PR #1281, opened 2022-04-27
feat!: add continuous deployment to cron package

### LINKED ISSUE #1129: Cron Package - Continuous Deployment

flea89:
Resolves #1129 

## 🚫  Before merging to main
- Create a tag, `cron-v1.0.0` to point to the latest main. This can be done on the merge commit of this PR in main?

## Rationale
Cron jobs should be running on the latest "released" cron package tag.
This is achieved by checking out the latest tag in the cron workflows.

## Nice To Have
The current implementation should work for existing jobs but it doesn't really prevent to have new jobs to be "missconfigured" (note checkouting latest tag).
It'd be nice to have a check in the scripts themselves, something like

`cron/src/index.js`
```
import { execSync } from 'child_process'
import { envConfig } from './lib/env.js'

export { dagCargoSizes } from './bin/dagcargo-sizes.js'
export { updatePins } from './bin/pins.js'
export { metrics } from './bin/metrics.js'

//Add new jobs here

checkOnLatestTag() {
  if(process.env.GITHUB_ACTIONS) {
   // check if on latest, if not raise
  }
}

envConfig()
checkOnLatestTag()

}

```

and then use [run-func](https://www.npmjs.com/package/run-func) to run the different jobs like this:

`cron/package.json`
```
...
scripts: {
  "start:pins": "run-func src/index.js updatePins",
  ...
}
```
What do you think?


flea89:
@alanshaw can you have a sanity check to see if this makes sense?

alanshaw (approved):
❤️ I love this!

flea89:
For context, this was discussed IRL with Alan, and we agreed this could go out without the Nice To have, given when adding a new Cron job devs generally take inspiration/copy and paste existing setup.

PR #1288, opened 2022-04-30
feat: Pinned/uploaded storage bar feedback

svvimming:
The storage bar in the storage manager section of the app UI has been split into different colours corresponding to uploaded and pinned files. There is also an accompanying legend.


orvn (changes requested):
@svvimming just one condition to add: if either the stored or pinned files is zero, we can hide the legend item referring to it.

orvn (approved):
Nice, looks good now 👍

PR #1291, opened 2022-05-02
style: Scale up app ui

svvimming:
Scale all rem values in the app scss files by 1.125.

### DIFF
--- a/packages/website/components/account/ctaCard/CTACard.scss
+++ b/packages/website/components/account/ctaCard/CTACard.scss
@@ -1,4 +1,4 @@
-$cta-card-padding-top: 4rem;
+$cta-card-padding-top: 4.5rem;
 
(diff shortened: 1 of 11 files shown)


orvn (changes requested):
@svvimming this looks great so far! I think scaling up the `rem` values this much works—we might even be able to do more, like `1.25`.

But also, we want to increase the actual width of the grid/page container. I think we'd minimally want to go with `1024px` (`64 rem`), but could also go higher to `1152px` (`72rem`).

Once this container width is changed for the whole app UI
- The increased `rem` sizes should look more organic and fit better
- The app UI responsive design might need some new rules to address the new sizes

svvimming:
closing in favour of https://github.com/web3-storage/web3.storage/pull/1300

Issue #1185, opened 2022-03-30
Default sort by should be newest to oldest


dchoi27:
Right now, default sorting is alphabetically, but that makes it confusing when a user uploads a file that isn't high up on the alphabet, and then they have no idea where to go (might have forgotten their file's name, etc.). Can we make the default sorting newest to oldest? Think it's fine to keep things in the current menu order, just change the default.

PR #1295, opened 2022-05-03
fix: Default sort by should be newest to oldest

joannaong:
fixes https://github.com/web3-storage/web3.storage/issues/1185


orvn (approved):

orvn:
Actually I noticed that although the order of the sort dropdown options were changed, the default sort was not `newest` (even though `newest` appeared first). Resolved that in the commit above.

Issue #1181, opened 2022-03-29
Change hover over description for "Queuing" in individual file rows


dchoi27:
Missed this since there's no parallel hover-over tooltip in NFT.Storage, but the hover-over information for Queuing is wrong - the data is queued for pinning.

![image.png](https://images.zenhubusercontent.com/5fa04bd64afb77249c77235b/b180ed44-dc12-4fef-9795-c180abd75d6b)![image.png](https://images.zenhubusercontent.com/5fa04bd64afb77249c77235b/2136a652-09f0-46d5-8e7b-60372932ec43)

Can we get rid of this tooltip description completely to match NFT.Storage?

Could we also change the hover-over tooltip description for the Status column to: "Reports the status of a file or piece of data stored on the IPFS Cluster. Status might not be fully up-to-date. Data is still available even when in Queued state." to match NFT.Storage?

cc @orvn

PR #1296, opened 2022-05-03
fix: Change hover over description for Queuing in individual file rows

joannaong:
fixes https://github.com/web3-storage/web3.storage/issues/1181


orvn (approved):
lgtm, works well on mobile too (although on super small screens, the tooltip tends to wrap a lot, I think this is acceptable)

PR #1302, opened 2022-05-09
feat: respond with unique error message when blocked API key is used

e-schneid:
When an API key is blocked in admin.storage, it sets deleted_at to the current time, and then sets the last auth_key_history item to have status 'Blocked.'  This PR just updates the code to look at the status so it can send a unique error to the client when an API key is deleted via a block.  Refers to #1382 in nft.storage.

### DIFF
--- a/packages/api/src/auth.js
+++ b/packages/api/src/auth.js
@@ -4,6 +4,7 @@ import {
   MagicTokenRequiredError,
   NoTokenError,
   PinningUnauthorizedError,
+  TokenBlockedError,
   TokenNotFoundError,
   UnrecognisedTokenError,
   UserNotFoundError
@@ -171,6 +172,17 @@ async function tryWeb3ApiToken (token, env) {
     // we have a web3 api token, but it's no longer valid
     throw new TokenNotFoundError()
   }
+
+  if (apiToken.isDeleted) {
+    const isBlocked = await checkIsTokenBlocked(apiToken, env)
+
+    if (isBlocked) {
+      throw new TokenBlockedError()
+    } else {
+      throw new TokenNotFoundError()
+    }
+  }
+
   return apiToken
 }
 
@@ -180,6 +192,10 @@ function findUserByIssuer (issuer, env) {
 
 function getUserTags (userId, env) {
   return env.db.getUserTags(userId)
+}
+
+function checkIsTokenBlocked (token, env) {
+  return env.db.checkIsTokenBlocked(token)
 }
 
 function verifyAuthToken (token, decoded, env) {
--- a/packages/api/src/errors.js
+++ b/packages/api/src/errors.js
@@ -58,6 +58,15 @@ export class TokenNotFoundError extends HTTPError {
 }
 TokenNotFoundError.CODE = 'ERROR_TOKEN_NOT_FOUND'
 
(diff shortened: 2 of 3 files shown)


packages/api/src/errors.js
```diff
@@ -58,6 +58,15 @@ export class TokenNotFoundError extends HTTPError {
 }
 TokenNotFoundError.CODE = 'ERROR_TOKEN_NOT_FOUND'
 
+export class TokenBlockedError extends HTTPError {
+  constructor (msg = 'API token is blocked') {
```
olizilla, line 62: @dchoi27 do we want to add something like `...please contact support@web3.storage` to the blocked token error message in case we ever block folks who are accidentally causing us problems rather deliberately ?

packages/api/src/auth.js
```diff
@@ -168,6 +169,17 @@ async function tryWeb3ApiToken (token, env) {
...
+  if (apiToken.isDeleted) {
+    const isBlocked = await checkIsTokenBlocked(apiToken, env)
+
+    if (isBlocked) {
+      throw new TokenBlockedError()
+    } else {
+      throw new TokenNotFoundError()
+    }
+  }
+
```
olizilla, line 182: The `checkIsTokenBlocked` check is good. Let's simplify this PR to keep it so that the db client `getKey` method doen't return deleted keys, and just add the blocked check once we determine that the key the user has sent us is not valid.

```suggestion
    // we have a web3 api token, but it's no longer valid
    const isBlocked = await checkIsTokenBlocked(apiToken, env)
    if (isBlocked) {
      throw new TokenBlockedError()
    } else {
      throw new TokenNotFoundError()
    }
  }

```

olizilla (changes requested):

packages/api/src/auth.js
jsdevel: @olizilla is there any performance benefit to filtering by isDeleted?  the lookup to fetch based on the token irrespective of whether or not it has been deleted i'm assuming is negligible.  the downside to your approach is that we have to read from the auth_key table twice in all scenarios.

packages/api/src/errors.js
jsdevel: probably.  we're going to add that verbiage in the UI soon so may as well return it in the API response.  @e-schneid can you add that?

packages/api/src/auth.js
olizilla: Ah, the bit I missed was we need the `auth_key` here to check if it is blocked. My propsal doesn't work. My only feels here is that It would be nice to not have to bring back all deleted keys here on the happy path, and the scheme feels a like it needs a refactor, but that is beyond the scope of this PR. I see the equivalent PR already got merged in nft.storage, so for the sake of consistency this should go in as is.

olizilla (approved):

packages/api/src/auth.js
jsdevel: @olizilla i may be wrong, but i believe the query in `dbClient.getKey` _only returns 1 key_ by virtue of `.eq('keys.secret', secret)`.

packages/api/src/errors.js
e-schneid: updated

PR #1305, opened 2022-05-09
feat: implement postgres optimization

francois-potato:
🛑 Requires db migration `packages/db/postgres/migrations/008-add-backup-urls.sql` to add new column
🛑 Requires db migration with the [following script](https://gist.github.com/yocontra/4ae719b42a54c4582ca28fcac74199d0)


packages/db/postgres/functions.sql
```diff
@@ -19,6 +19,11 @@ CREATE OR REPLACE FUNCTION json_arr_to_json_element_array(_json json)
   RETURNS json[] LANGUAGE sql IMMUTABLE PARALLEL SAFE AS
 'SELECT ARRAY(SELECT * FROM json_array_elements(_json))';
 
+-- transform a JSON array property into an array of SQL jsonb elements
+CREATE OR REPLACE FUNCTION json_arr_to_jsonb_element_array(_json json)
```
flea89, line 23: Isn't this a duplicate of [json_arr_to_json_element_array](https://github.com/web3-storage/web3.storage/pull/1305/files#diff-bf4100e6931ed75d8fbe33d506c141af632f120f4df66bfd0ff9c92d80da40c6R18)?

packages/db/index.js
```diff
@@ -430,21 +433,20 @@ export class DBClient {
    */
   async getBackups (uploadId) {
     /** @type {{ data: Array<definitions['backup']>, error: PostgrestError }} */
```
flea89, line 435: This type does not exist anymore, can you update?

packages/db/postgres/functions.sql
```diff
@@ -120,48 +125,39 @@ CREATE OR REPLACE FUNCTION create_upload(data json) RETURNS TEXT
...
             (data ->> 'inserted_at')::timestamptz,
-            (data ->> 'updated_at')::timestamptz)
+            (data ->> 'updated_at')::timestamptz,
+            json_arr_to_jsonb_element_array(data -> 'backup_urls'))
   ON CONFLICT ( user_id, source_cid ) DO UPDATE
     SET "updated_at" = (data ->> 'updated_at')::timestamptz,
         "name" = data ->> 'name',
-        "deleted_at" = null
+        "deleted_at" = null,
+        "backup_urls" = array_cat(upld.backup_urls, json_arr_to_jsonb_element_array(data -> 'backup_urls'))
```
flea89, line 158: I haven't had a whole lot of time to dig into this so please take what I'm going to say with a pinch of salt.

With this implementation, if a user uploads the same CID multiple times we might end up with multiple backup URLs that are the same, isn't it?


At the same time, we need to be able `concat` to handle chunked uploads. 
When a chunked upload (>100MB) is in progress we hit POST `/car`  multiple times, which means we call `db.createUpload` multiple times and I'd assume the expected S3 files and therefore urls in the db should be
`/raw/ROOT_CID1/u1/CAR1.car, upload_key_1`
`/raw/ROOT_CID1/u1/CAR2.car, upload_key_1`


I guess we should, if possible, concat if  the url doesn't exists already. Which mimics in a way what we had before with 

```
ON CONFLICT ( upload_id, url ) DO NOTHING;
```


I've also noticed this is not in NFT.storage, we might have to port there whatever we figure out here, isn't it?

packages/db/postgres/migrations/009-migrate-backup-urls.sql
```diff
@@ -0,0 +1 @@
+UPDATE "upload" SET "backup_urls" = (SELECT array_agg(jsonb_build_object('url', url, 'inserted_at', inserted_at)) FROM backup WHERE "upload_id" = "upload"."id");
```
flea89, line 1: Nit.
Add new line at the end of the file.

I've noticed this in other files, possibly something you want to set up in your editor.

flea89 (commented):
Great work @francois-potato!

Left some comments

packages/db/postgres/functions.sql
francois-potato: That make sense. I made some changes to make backup URLs unique (a JSON doc with the URL as a key).

packages/db/postgres/migrations/009-migrate-backup-urls.sql
```diff
@@ -0,0 +1 @@
+UPDATE "upload" SET "backup_urls" = (SELECT json_object_agg(url, jsonb_build_object('url', url, 'inserted_at', inserted_at)) FROM backup WHERE "upload_id" = "upload"."id");
```
vasco-santos, line 1: We have all the backup urls currently in Backup table, so I suppose we want to copy them in the migration?

@hugomrdias did you create a migration script for this?

packages/db/postgres/functions.sql
```diff
@@ -120,48 +120,39 @@ CREATE OR REPLACE FUNCTION create_upload(data json) RETURNS TEXT
...
             (data ->> 'inserted_at')::timestamptz,
-            (data ->> 'updated_at')::timestamptz)
+            (data ->> 'updated_at')::timestamptz,
+            data -> 'backup_urls')
   ON CONFLICT ( user_id, source_cid ) DO UPDATE
     SET "updated_at" = (data ->> 'updated_at')::timestamptz,
         "name" = data ->> 'name',
-        "deleted_at" = null
+        "deleted_at" = null,
+        "backup_urls" = (data -> 'backup_urls')::jsonb || upld.backup_urls
```
vasco-santos, line 153: This will replace the backups on chunked uploads, so I think we need to "merge" them here.

Can we add a test?

vasco-santos (changes requested):

packages/db/postgres/functions.sql
francois-potato: @vasco-santos `||` act as a merge on JSONB

> The || operator concatenates two JSON objects by generating an object containing the union of their keys, taking the second object's value when there are duplicate keys. 

Good catch, I added a test (please see `can backup chunked uploads` in `packages/db/test/backup.spec.js`).

vasco-santos: wow, that's pretty cool, didn't know || feature as JSONB! I see by the test it is working  😉

packages/db/postgres/migrations/009-migrate-backup-urls.sql
francois-potato: The migration script should be enough I think?

`UPDATE "upload" SET "backup_urls" = (SELECT json_object_agg(url, jsonb_build_object('url', url, 'created', inserted_at)) FROM backup WHERE "upload_id" = "upload"."id")`

will build the JSON structure e.g. `{"https://backup/":  { url: "https://backup/", created: '2022-05-18 10:30:00+00:00' }}` from the Backup table using `upload_id`.

vasco-santos: Yeah, this was mostly a call out that we need to setup a maintenance window. I think this works, even though we should test it in staging first.

packages/db/index.js
```diff
@@ -218,6 +218,17 @@ export class DBClient {
...
     const now = new Date().toISOString()
+
+    const backupsObjects = data.backupUrls.map(url => ({
+      url,
+      created: data.created || now
+    }))
+
+    const backupsObjectsByURL = {}
+    backupsObjects.forEach(
+      backup => (backupsObjectsByURL[backup.url] = backup)
```
vasco-santos, line 229: nit: Can we just do the forEach on `data.backupUrls` and mutate `backupsObjectsByURL` as needed by creating the object here?

packages/db/test/backup.spec.js
```diff
@@ -89,8 +89,56 @@ describe('backup', () => {
...
+    upload = await client.getUpload(cid, user._id)
+    assert(upload, 'second upload created')
+
+    const backups = await client.getBackups(upload._id)
+
+    assert.strictEqual(backups.length, 1, 'upload has a single backup')
+    assert(new Date(backups[0].created) < new Date(upload.updated), 'backup was created before the new upload')
+  })
+
+  it('can backup chunked uploads', async () => {
```
vasco-santos, line 118: Probably worth also checking if same URL is provided that it does not have both?

Can we also make this test be isolated? I see we rely on side effects from previous ones with uploads, but it would be better to have a different CID here and perform  `createUpload` function twice

vasco-santos (changes requested):
Great work @francois-potato 

Left two minor comments that should be addressed. Would also love to have @hugomrdias 's input before moving forward as this needs to be changed in nft side and would like to be aligned.

packages/db/postgres/migrations/009-migrate-backup-urls.sql
hugomrdias: the nft.storage migration script was this https://github.com/nftstorage/nft.storage/pull/1664#issuecomment-1082183567

packages/db/index.js
hugomrdias: why do we do this ? can't this be just a list of urls ?

hugomrdias: what if `backups_urls` is just an `text[]`, this function send only one url per call and the pg create upload function just appends to the array ?

hugomrdias (changes requested):

packages/db/index.js
vasco-santos: We should also guarantee URLs are unique

hugomrdias: yes, because it was not appending just overwriting

packages/db/test/backup.spec.js
francois-potato: Great improvement! I made some changes. The `creates unique backup urls` tests for duplicated URLs.

packages/db/postgres/migrations/009-migrate-backup-urls.sql
francois-potato: @vasco-santos, do you think I should port this script for web3.storage so we can batch the updates for the rollout?

packages/db/index.js
francois-potato: I went with a JSONB structure to keep the creation timestamp and handle unique URLs. This also adds the possibility of adding extra metadata to the backup in the future.

We could call `createUpload` multiple times with the same backup URL. I'm unsure how to enforce unique URLs when appending to the array (maybe with unnest and aggregation).

I'm new to the project, and I'm unsure why `createUpload` accepts an array for `backupUrls`. From an existing test, it looks like it is to support creating backups in different locations/regions. It could be refactored to accept only one URL.

packages/db/postgres/migrations/009-migrate-backup-urls.sql
hugomrdias: yes we should be able to migrate without downtime, ie. deploy to prod, start writing to the new column and run script to migrate from table to column.

packages/db/index.js
hugomrdias: We dont need any extra metadata about these urls in fact they will go away with uploads v2, we need them now just as a reference to be able to do migrations etc.

Create upload in the api and pg function should only accept one url, even when we chunk we just call create upload multiple times with the same root CID.

I get that jsonb merge is nice here, but we need to think about nft.storage and how we are going to fix it. In nft.storage we already have an array, the easiest migration path is to just append to it instead of migrating to a new structure.

Having only unique urls is nice but not critical because in any situation where we actually need to read those we can just select distinct from the array.

francois-potato: @hugomrdias, ok, thank you, this makes sense now. I've made the changes to use text[] instead.

packages/db/postgres/migrations/009-migrate-backup-urls.sql
```diff
@@ -0,0 +1 @@
+UPDATE "upload" SET "backup_urls" = (SELECT array_agg("url") FROM "backup" WHERE "upload_id" = "upload"."id");
```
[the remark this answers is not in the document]
francois-potato: Now that the schema is the same (`text[]` instead of `jsonb`), the [nft batch script](https://gist.github.com/yocontra/4ae719b42a54c4582ca28fcac74199d0) works without modifications. I tested it locally too.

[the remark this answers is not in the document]
vasco-santos: Yeah, I think batch script is best way to go. 

@hugomrdias how long did it take for you? We should probably schedule a maintenance windows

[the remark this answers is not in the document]
vasco-santos: awesome ok thanks, I will setup a migration window for this then.
@francois-potato we should be ready to merge here and do the maintenance next monday right?

vasco-santos (approved):

PR #1312, opened 2022-05-11
fix: upgrade miniflare to ^2.4.0

adamalton:
This gets us the fix to this: https://github.com/cloudflare/miniflare/issues/188 which is currently breaking all tests on the CI.


francois-potato (approved):

vasco-santos (approved):
🎉

Issue #1266, opened 2022-04-21
Complete the user blocking feature 


JeffLowe:
Upon review of Web3 website on 4.20.22, certain parts of the user blocking feature were not operational.  
The following needs to be completed on the Web3.storage site to complete the user blocking feature. 

- [x] display UI message upon blocked user login
SPEC:  LOGIN UI- upon login to web UI, blocked user gets a message explaining their account was blocked, and to reach out to support@ if they haven't already reached out).  Email in copy should be to support@web3.storage

Observation:  Does not display UI message upon blocked user login*

Example of NFT.Storage UI message
![uponLoginwhenBlocked.png](https://images.zenhubusercontent.com/5285300029fe34596000054e/f04e214e-3afa-4a6f-ab06-c84ff217eaac)

- [x] When user is blocked, deactivate "new api" key, preventing users from creating new keys (buttons found on https://web3.storage/tokens/ and on https://web3.storage/account/).  Note, there is also a nav option to create new token which leads to this route https://web3.storage/tokens/ that needs to be handled. 
SPEC:  API KEY - blocked user unable to create any new api key. The "+ New Key" button is greyed out and deactivated.

Observation:  Blocked user able to create new api key (button needs deactivating with UI tooltip)

- [x] Add tooltip to blocked user's new api key button https://web3.storage/tokens/
SPEC:  API KEY UI - Blocked user gets ui message in new api key button tooltip or click action explaining.  

Observation (same as above):  Blocked user able to create new api key (button needs deactivating with UI tooltip)

Example of UI message and button in NFT.Storage
![NewKeyAttemptWhenBlocked.png](https://images.zenhubusercontent.com/5285300029fe34596000054e/85be25b6-b928-482a-908a-5226416c30af)

- [x] For blocked users, Grey out and deactivate the "Upload" buttons in UI (https://web3.storage/account/ )
- [x] For blocked user, add ui tooltip to upload buttons

SPEC:  UPLOAD UI - Grey out and deactivate the "Upload" buttons in UI on /files and /new-file routes. gets UI feedback in upload file button tooltip and on click action for reason.

Observation: *Blocked user able to initiate an upload. User uploads fail silently in UI, only shows “Failed” in upload modal screen. Console does show error response of “This account is restricted, email support@web3.storage for more information. But upload button need’s to be deactivated

Example of button and tooltip in nft.storage
[...]

PR #1322, opened 2022-05-13
feat: add user blocking functionality

e-schneid:
Mimicking recent changes to nft.storage, we're adding functionality here to display a modal when a user's account is restricted as well as disabled buttons and tooltips

Completes #1266


jsdevel (approved):

jsdevel:
@e-schneid changes look good!

PR #1324, opened 2022-05-16
feat: js client accepts user provided fetch function

duxiaofeng-github:
I need to process some headers, like the cookie. because I post the request to my service and do some validation and then I create a proxy request to forward the request to your server. so, I need to add the `credentials: "include"` option to the `requestInit` option.

therefore, I add the `pre` middleware to achieve that, it will pass the requestInit to the user, and the user modifies and returns it to the client before the request is sent.

### DIFF
--- a/packages/client/src/lib.js
+++ b/packages/client/src/lib.js
@@ -23,7 +23,7 @@ import { CarReader } from '@ipld/car'
 import { filesFromPath, getFilesFromPath } from 'files-from-path'
 import throttledQueue from 'throttled-queue'
 import {
-  fetch,
+  fetch as _fetch,
   File,
   Blob,
   Blockstore
@@ -83,13 +83,13 @@ class Web3Storage {
    * const client = new Web3Storage({ token: API_TOKEN })
    * ```
    *
-   * @param {{token: string, endpoint?:URL, rateLimiter?: RateLimiter, customFetch: typeof fetch}} options
+   * @param {{token: string, endpoint?:URL, rateLimiter?: RateLimiter, fetch: typeof _fetch}} options
    */
   constructor ({
     token,
     endpoint = new URL('https://api.web3.storage'),
     rateLimiter,
-    customFetch,
+    fetch = _fetch,
   }) {
     /**
      * Authorization token.
@@ -107,10 +107,10 @@ class Web3Storage {
      */
     this.rateLimiter = rateLimiter || createRateLimiter()
     /**
-     * if a custom fetch function existed, will use it instead of the original fetch function
+     * Optional custom fetch function. Defaults to global fetch in browsers or @web-std/fetch on node.
      * @readonly
      */
-    this.customFetch = customFetch;
+    this.fetch = fetch;
   }
 
   /**
@@ -132,7 +132,7 @@ class Web3Storage {
    * @param {PutOptions} [options]
    * @returns {Promise<CIDString>}
    */
-  static async put ({ endpoint, token, rateLimiter = globalRateLimiter, customFetch }, files, {
+  static async put ({ endpoint, token, rateLimiter = globalRateLimiter, fetch = _fetch }, files, {
     onRootCidReady,
     onStoredChunk,
     maxRetries = MAX_PUT_RETRIES,
@@ -154,7 +154,7 @@ class Web3Storage {
       })
       onRootCidReady && onRootCidReady(root.toString())
       const car = await CarReader.fromIterable(out)
-      return await Web3Storage.putCar({ endpoint, token, rateLimiter, customFetch }, car, { onStoredChunk, maxRetries, maxChunkSize, name })
+      return await Web3Storage.putCar({ endpoint, token, rateLimiter, fetch }, car, { onStoredChunk, maxRetries, maxChunkSize, name })
     } finally {
       await blockstore.close()
     }
@@ -166,7 +166,7 @@ class Web3Storage {
    * @param {PutCarOptions} [options]
    * @returns {Promise<CIDString>}
    */
-  static async putCar ({ endpoint, token, rateLimiter = globalRateLimiter, customFetch }, car, {
+  static async putCar ({ endpoint, token, rateLimiter = globalRateLimiter, fetch = _fetch }, car, {
     name,
     onStoredChunk,
     maxRetries = MAX_PUT_RETRIES,
@@ -209,10 +209,10 @@ class Web3Storage {
       const res = await pRetry(
         async () => {
           await rateLimiter()
-          const request = await (customFetch||fetch)(url.toString(), {
+          const request = await fetch(url.toString(), {
             method: 'POST',
             headers,
-            body: carFile,
+            body: carFile
           })
           /* c8 ignore next 3 */
           if (request.status === 429) {
@@ -245,12 +245,12 @@ class Web3Storage {
    * @param {CIDString} cid
    * @returns {Promise<Web3Response | null>}
    */
-  static async get ({ endpoint, token, rateLimiter = globalRateLimiter, customFetch }, cid) {
+  static async get ({ endpoint, token, rateLimiter = globalRateLimiter, fetch = _fetch }, cid) {
     const url = new URL(`car/${cid}`, endpoint)
     await rateLimiter()
-    const res = await (customFetch||fetch)(url.toString(), {
+    const res = await fetch(url.toString(), {
       method: 'GET',
-      headers: Web3Storage.headers(token),
+      headers: Web3Storage.headers(token)
     })
     /* c8 ignore next 3 */
     if (res.status === 429) {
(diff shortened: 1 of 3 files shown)


duxiaofeng-github:
can anyone do a review for this pr? @olizilla

olizilla:
Can you share a link to your code where you need this? It seems reasonable, but I just want to see the calling code to better understand what you need.

olizilla:
For example, if we land something like this, then perhaps we should just let the calling code pass in a custom `fetch` function. I think that would still work for what you need, and handle whatever intercepting other users might want to do in the future as well.

olizilla:
@duxiaofeng-github what do you think to letting the user pass in their own fetch function which could have full control over all the arguments passed to fetch.

duxiaofeng-github:
> For example, if we land something like this, then perhaps we should just let the calling code pass in a custom `fetch` function. I think that would still work for what you need, and handle whatever intercepting other users might want to do in the future as well.

it's not about the fetch function, it is about the args passed to fetch function. but yup, it can be done by passing the fetch function.

duxiaofeng-github:
> @duxiaofeng-github what do you think to letting the user pass in their own fetch function which could have full control over all the arguments passed to fetch.

so, that's what i have done, the pre function accept the  params which needed by the web3.storage, the user can modify them in pre function and return a new one, the return value will put into the fetch function.

duxiaofeng-github:
@olizilla 
here's my code snippet:
```
function makeStorageClient(token: string) {
  return new Web3Storage({
    token,
    endpoint: new URL("http://localhost:8082"),
    pre: (init: RequestInit) => {
      init.credentials = "include";

      if (init.headers != null) {
        const headers = new Headers(init.headers);
        headers.delete("Authorization");
        init.headers = headers;
      }

      return init;
    },
  });
}
```

duxiaofeng-github:
why is the ci always failed? @orvn

olizilla (changes requested):
This is an interesting idea. If you have time to edit it so that a user may pass in a custom `fetch` function instead of a `pre`, and add a test then we can get this merged.

duxiaofeng-github:
ok, i can add a `customFetch` option, but what should I test for this case? should I just test the `customFetch` whether being executed?

duxiaofeng-github:
> This is an interesting idea. If you have time to edit it so that a user may pass in a custom `fetch` function instead of a `pre`, and add a test then we can get this merged.

@olizilla ok, i add a `customFetch` option already

packages/client/src/lib.js
```diff
@@ -148,7 +154,7 @@ class Web3Storage {
       })
       onRootCidReady && onRootCidReady(root.toString())
       const car = await CarReader.fromIterable(out)
-      return await Web3Storage.putCar({ endpoint, token, rateLimiter }, car, { onStoredChunk, maxRetries, maxChunkSize, name })
+      return await Web3Storage.putCar({ endpoint, token, rateLimiter, customFetch }, car, { onStoredChunk, maxRetries, maxChunkSize, name })
```
olizilla, line 157:
```suggestion
      return await Web3Storage.putCar({ endpoint, token, rateLimiter, fetch }, car, { onStoredChunk, maxRetries, maxChunkSize, name })
```

packages/client/src/lib.js
```diff
@@ -203,10 +209,10 @@ class Web3Storage {
       const res = await pRetry(
         async () => {
           await rateLimiter()
-          const request = await fetch(url.toString(), {
+          const request = await (customFetch||fetch)(url.toString(), {
```
olizilla, line 212: If we default fetch in the constructor, this line is neater
```suggestion
          const request = await fetch(url.toString(), {
```

packages/client/src/lib.js
```diff
@@ -239,12 +245,12 @@ class Web3Storage {
    * @param {CIDString} cid
    * @returns {Promise<Web3Response | null>}
    */
-  static async get ({ endpoint, token, rateLimiter = globalRateLimiter }, cid) {
+  static async get ({ endpoint, token, rateLimiter = globalRateLimiter, customFetch }, cid) {
```
olizilla, line 248:
```suggestion
  static async get ({ endpoint, token, rateLimiter = globalRateLimiter, fetch = _fetch }, cid) {
```

packages/client/src/lib.js
```diff
@@ -239,12 +245,12 @@ class Web3Storage {
...
     const url = new URL(`car/${cid}`, endpoint)
     await rateLimiter()
-    const res = await fetch(url.toString(), {
+    const res = await (customFetch||fetch)(url.toString(), {
       method: 'GET',
-      headers: Web3Storage.headers(token)
+      headers: Web3Storage.headers(token),
```
olizilla, line 253:
```suggestion
      headers: Web3Storage.headers(token)
```

packages/client/src/lib.js
```diff
@@ -269,12 +275,12 @@ class Web3Storage {
    * @param {CIDString} cid
    * @returns {Promise<Status | undefined>}
    */
-  static async status ({ endpoint, token, rateLimiter = globalRateLimiter }, cid) {
+  static async status ({ endpoint, token, rateLimiter = globalRateLimiter, customFetch }, cid) {
     const url = new URL(`status/${cid}`, endpoint)
     await rateLimiter()
-    const res = await fetch(url.toString(), {
+    const res = await (customFetch||fetch)(url.toString(), {
```
olizilla, line 281:
```suggestion
    const res = await fetch(url.toString(), {
```

packages/client/src/lib.js
```diff
@@ -302,17 +308,17 @@ class Web3Storage {
    * @param {{before: string, size: number}} opts
    * @returns {Promise<Response>}
    */
-    async function listPage ({ endpoint, token, rateLimiter = globalRateLimiter }, { before, size }) {
+    async function listPage ({ endpoint, token, rateLimiter = globalRateLimiter, customFetch }, { before, size }) {
```
olizilla, line 311:
```suggestion
    async function listPage ({ endpoint, token, rateLimiter = globalRateLimiter, fetch = _fetch }, { before, size }) {
```

packages/client/src/lib.js
```diff
@@ -302,17 +308,17 @@ class Web3Storage {
...
       const search = new URLSearchParams({ before, size: size.toString() })
       const url = new URL(`user/uploads?${search}`, endpoint)
       await rateLimiter()
-      return fetch(url.toString(), {
+      return (customFetch||fetch)(url.toString(), {
```
olizilla, line 315:
```suggestion
      return fetch(url.toString(), {
```

packages/client/src/lib/interface.ts
```diff
@@ -13,6 +13,7 @@ export interface Service {
   endpoint: URL
   token: string
   rateLimiter?: RateLimiter
+  customFetch?: typeof fetch
```
olizilla, line 16:
```suggestion
  fetch?: typeof fetch
```

packages/client/test/custom-fetch.spec.js
```diff
@@ -0,0 +1,65 @@
...
+    // @ts-ignore
+    let result = false;
+
+    function customFetch(...opts) {
+      result = true;
+
+      fetch(...opts);
+    }
+
+    const client = new Web3Storage({ endpoint, token, customFetch })
```
olizilla, line 38:
```suggestion
    const client = new Web3Storage({ endpoint, token, fetch: customFetch })
```

olizilla (changes requested):
Nice! Thanks for pushing this forward! As this will be part of the public api I'd like to rename `customFetch` to just `fetch`. I've provided some of the changes needed to land this as suggestions.

I think all that is left is to alias the fetch we import at the top of the file like

```js
import {
  fetch as _fetch,
  File,
  Blob,
  Blockstore
} from './platform.js'
```

olizilla:
@duxiaofeng-github Please run `npm run lint:fix` from the root of the project to fix the issues that Standard JS is reporting, and push the changes.

duxiaofeng-github:
@olizilla  it's done

packages/client/test/custom-fetch.spec.js
```diff
@@ -0,0 +1,64 @@
...
+    result = false
+    await client.put(files)
+    assert.is(result, true)
+
+    result = false
+    await client.get(cid)
+    assert.is(result, true)
+
+    result = false
+    await client.list()
```
olizilla, line 50: `list()` returns `AsyncIterable`. The return value needs to be iterated to trigger a call to fetch

duxiaofeng-github:
@olizilla fixed

olizilla:
you can run the `lint` and `test` steps locally by running

```console
$ npm run lint

$ npm run test -w packages/client
```

you need docker running for the tests to set up the needed containers, but otherwise it should just work

duxiaofeng-github:
@olizilla it should works this time

alanshaw (approved):

alanshaw:
@duxiaofeng-github this has been released in v4.3.0

PR #1335, opened 2022-05-17
feat: Message bar incident/maintenance name

svvimming:
This PR makes a change to how the message bar displays incident/maintenance messages. Rather than displaying the body of the message which can cause the message bar container to become very large, now only the name is displayed. Furthermore, the message is now wrapped in a link to more information about the issue.


orvn (changes requested):
@svvimming I think this was created before you had `main` pulled in locally so the Docusaurus site content (which has since been removed), is being PR'd here by accident. 

Can you do a cleanup and ensure you're only committing your new code against `main`?

orvn (approved):
Tested with an edit of my hosts file to replace the statuspage endpoint

PR #1338, opened 2022-05-18
fix: show custom storage quota to user

### LINKED ISSUE #1326: Display custom storage quota in account section
### Overarching user needs / context:

* Currently we can increase users' allowed storage quota from admin.storage, but there is no way for users to check on their current quota.
* If users have a 2 TiB quota, their account will still say
  > Storage: 0.9 TiB of 1 TiB used

  because the "1 TiB" is currently hardcoded

### Scope of this ticket:

* This ticket is solely about updating the copy to reflect the amount of storage attributed to everyone individually.
* The scope does not include changing the scale of the "progress bar" that visually reflects the proportion of storage used compared to your quota.

### Acceptance criteria:

* When we update a user's user tag for allowed storage, the amount should be visible in their account section.
* The amount of bytes stored in the DB should be displayed as amount in TiB

adamalton:
Fixes #1326.

Arguably this is more of a feature than a bug fix, but given that we already have the increasable storage limits deployed, I'm calling it a defect 🙂 .

* I've removed the storage tiers concept, which as far as I can tell was an idea that never actually got fully implemented.
* I've left the `"TiB"` string in `packages/website/content/pages/app/account.json` as it's a potentially translatable piece of text.
* The fact that the default storage limit value of 1TiB  is hard coded in several different places throughout the application makes me slightly uneasy, as one of them is bound to get missed if/when we change it, but getting around that is somewhat tricky given the application structure, so I'm calling it out of scope for this PR.


packages/website/components/account/storageManager/storageManager.js
```diff
@@ -34,39 +28,23 @@ const mailTo = `mailto:${emailContent.mail}?subject=${emailContent.subject}&body
...
-        [StorageTiers.TIER_3]: {
-          maxSpaceLabel: `${Math.floor(uploaded + pinned / (tebibyte * 10) + 1) + content.tiers[2].max_space_label}`,
-          // every increment of 10 changes the amount of space used
-          percentUploaded: ((uploaded % (tebibyte * 10)) / (tebibyte * 10)) * 100,
-          percentPinned: ((pinned % (tebibyte * 10)) / (tebibyte * 10)) * 100,
-        },
-      }[storageTier]),
-    [storageTier, uploaded, pinned, content.tiers]
+    () => ({
+      maxSpaceLabel: `${limit / tebibyte} ${content.max_space_tib_label}`,
```
francois-potato, line 42: Should we `Math.floor(limit / tebibyte)` in case we end up with a long floating point number?

adamalton: Good point. Done.

vasco-santos (approved):

packages/api/src/user.js
```diff
@@ -111,10 +111,13 @@ export async function userTokensPost (request, env) {
  * @param {import('./env').Env} env
  */
 export async function userAccountGet (request, env) {
-  const usedStorage = await env.db.getStorageUsed(Number(request.auth.user._id))
-
+  const [usedStorage, storageLimitBytes] = await Promise.all([
+    env.db.getStorageUsed(request.auth.user._id),
+    env.db.getUserTagValue(request.auth.user._id, 'StorageLimitBytes')
```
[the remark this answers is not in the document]
adamalton: Yep. It could also/instead possibly be changed to use the stuff that's now been done in #1322. I've made ticket #1383.

orvn:
> The fact that the default storage limit value of 1TiB is hard coded in several different places throughout the application makes me slightly uneasy, as one of them is bound to get missed if/when we change it, but getting around that is somewhat tricky given the application structure, so I'm calling it out of scope for this PR.

This was a (not so ideal) pattern from the old site, will consolidate it into one place and translate the value into a human readable format.

PR #1339, opened 2022-05-18
chore: improve cron job testing for storage quota check

francois-potato:
This PR update some tests that were calling external scripts and using stdout/stderr to assert actions. The tests are now using a mock, which also improves speed.


flea89:
LGTM

PR #1346, opened 2022-05-23
fix: typo in Logging constructor

alanshaw:
I assume this is a typo?

@adamalton can you also look into this logger running during testing in CI?

https://github.com/web3-storage/web3.storage/runs/6553500624?check_suite_focus=true


adamalton:
Ah yes, it is! Even it that was what I'd wanted to do, it wouldn't have occurred to me to use that syntax to achieve it!

adamalton (approved):

PR #1361, opened 2022-05-26
feat: add cloudflare img component

joannaong:


hugomrdias (changes requested):
LGTM, just one nit theres no real need for the illustrations components right ? Either they are used in the hero with the HeroIllustration comp (with hero classname and priority) or they can just be imported inline where used.

hugomrdias:
> Note that not all images are using Cloudflare images. Some of the backgrounds are kept as is since these requires specific positioning

They can you just need to dig a little deeper, check https://github.com/web3-storage/web3.storage/pull/1434

Even css urls can have the images.web3.storage prefixed!

orvn:
Looks good so far, just doing a cross browser test to make sure this is stable across different platforms!

orvn (changes requested):
It looks like some of the background images aren't stretched to their original size.

The one I'm seeing this most on right now is the uploads modal background. Video below starts with the current site, and then switches to this branch.

https://user-images.githubusercontent.com/3689720/173100279-e92e6ac6-d08a-440f-981a-ca0ed4299d60.mp4

joannaong:
> It looks like some of the background images aren't stretched to their original size.
> 
> The one I'm seeing this most on right now is the uploads modal background. Video below starts with the current site, and then switches to this branch.
> 
>  uploads-background-image-change.mp4

I've adjusted. Please let me know if I'm missing anything else

orvn (changes requested):
@joannaong I only see an issue on this card in the account page:
 

Second image is how it _should_ look.

Could you also fix the conflicts against `main`?

orvn (approved):
Tested throughout, looks good!

orvn:
@hugomrdias, merging is blocked for me until your review is cleared as well, but everything looks good from my end!

hugomrdias:
LGTM let ship it. Great work!!

Issue #1254, opened 2022-04-18
Show both stored and pinned tables in app UI


orvn:
Currently, the stored files are displayed in a table in the app UI, but pinned content is not. These should be displayed, and this feature should be implemented in lockstep with #989.

The tabular content needed in both cases is different enough to warrant two separate tables.

## Options

1. ❌ Separate routes, which open a page (or modal) showing the requested table
2. ❌ Two tables, underneath one another (user would scroll)
3. ✅ A tabbed UI, where the user can switch tables in the same container
(open to any other ideas)

The third option seems the best because: 

- It gives the user the ability to switch views with minimal layout shift, so they'll stay in the same mental context due to less cognitive load
- Table controls (sort, search, pagination) do not need to be duplicated or altered significantly

> _Edit: the third option was selected_

orvn:
(This issue was accidentally closed because it was linked to the related issue of separation of directly uploaded and pinned in the usage bar)

- Note that there's an approved copy change related to this, as to the naming convention of the two tables
  - _Uploaded Directly_
  - _Uploaded via Pinning API_ 
  
### Design

The following is being implemented in the app UI design. 
- Clicking one of the items in the legend near the storage bar also switches tables
- Two tabs allow switching between the direct uploads and pinned tables
- Switching tables add a GET param to the URL
  - Consequently, loading a page with one of these params opens the UI to that table

<img width="750" alt="app ui tabs pinned vs direct" src="https://user-images.githubusercontent.com/3689720/169576411-eaaf724c-7814-4c31-868d-b6ceb5705a5c.png">

orvn:
@JeffLowe, it's almost done in #1363

PR #1363, opened 2022-05-26
feat: Split file manager table into uploaded & pinned

svvimming:
This PR splits the file manager table into two tabs: Files uploaded directly and files uploaded through the pinning API.

completes #1254


orvn (changes requested):
(ignore, reviewed too early by accident, didn't see that this was a draft)

svvimming:
The files manager table now retrieves a list of files pinned using the pinning API. The token for this request has to be one of the generated API tokens associated with the user, not a magic token. If none are available the request is not made and the pinned files table will be disabled and its tab is greyed out.

orvn:
@dchoi27, before this is merged, I wanted to run the null case by you: if a user has nothing uploaded via the pinning API, should we hide that tab? 

Currently it's in a disabled state, but will always appear. On the one hand, it clutters up the UI (in a minor way), but on the other, it's useful to keep if you want to make people aware of the pining API in a more indirect way.

split-table-default

orvn (approved):
LGTM!

As a minor ux nit, in the future it might be good to alter the design such that it's super clear that the top-level items are tabs, not just headings.

PR #1369, opened 2022-05-27
chore: send list of storage quota violators to support@web3.storage

adamalton:
As specified in #990, these notification emails should go to support@web3.storage, rather than admin@web3.storage.

I considered changing all the associated mentions of "admin" (e.g. "adminUser") to "support" as well, but I don't think it's worth it for the amount of disruption it would cause, and "admin" is still a valid label, we're just changing the email address.


flea89 (approved):
LGTM

PR #1371, opened 2022-05-27
chore: change the storage cron job to run in prod, and every 6 hours

### LINKED ISSUE #1370: Run storage limit notifications cron on production every 6 hours
In #1273 we added a feature to send notification emails about storage quotas. Currently the GitHub workflow is only configured to run it on staging every 12 hours. We want to change this to run it on prod every 6 hours.

adamalton:
Fixes #1370.

### DIFF
--- a/.github/workflows/cron-storage-limits.yaml
+++ b/.github/workflows/cron-storage-limits.yaml
@@ -2,7 +2,8 @@ name: Storage Limit Email Notifications
 
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


adamalton:
For context, we've now deployed the change to display the storage quota in the UI (#1338) and bumped the quotas on prod for the list of people who need them. So we should now be good to start running this cron on prod to trigger notification emails about storage limits.

PR #1374, opened 2022-05-27
fix: missing deleted_at property on tags

flea89:


adamalton:
I've now written this up as a new PR with a more descriptive branch name in #1379. Closing this.

PR #1353, opened 2022-05-24
chore: update API and Website Readmes

joshJarr:
Updating the readme to address a few issues I found when setting up the project. 

- When running the API using the readme in `packages/api/README.md` I missed the DB  _"Load schema"_ step resulting in some confusion when trying to log in. Thought I'd add this context in the API readme.
- I got a little confused when running the website, I didn't realise running my env vars for "for local debugging API" meant I'd be using the mock API and not my local API, so updated some text here.

Hope that's okay!

Thanks

PR #1378, opened 2022-05-30
docs: make readmes more self contained

olizilla:
it's not super fun to have to read multiple readmes to get a project started.

Inspired by @joshJarr in #1353  this PR pulls the minimum steps to get a working set up into the main and api readme.

License: (Apache-2.0 AND MIT)


joshJarr:
Awesome stuff! 🙌  It makes a lot of sense to encourage everything to be run from root

Would it make sense to move the db init step to after the run step? Since running the API also starts the db container.
I've noticed `load-schema` doesn't start the db container, so being the first step would time out. 

Alternatively we could start the container in the `load-schema` step, what do you think?

PR #1379, opened 2022-05-30
chore: fix tags in api user info

adamalton:
#1302 introduced showing a message to the user if their account is restricted, but the way that it fetches the rows from the `user_tag` table doesn't fetch the `deleted_at` column, so it will return the wrong value when there's a deleted tag. This fixes that bug.

I've done this as a "chore" because the changes in #1302 aren't deployed to the website yet, so if this can be merged before they're deployed then this won't become a user-facing bug, hence chore rather than fix.

### DIFF
--- a/packages/db/index.js
+++ b/packages/db/index.js
@@ -42,7 +42,7 @@ const userQueryWithTags = `
   publicAddress:public_address,
   created:inserted_at,
   updated:updated_at,
-  tags:user_tag_user_id_fkey(user_id,id,tag,value)
+  tags:user_tag_user_id_fkey(user_id,id,tag,value, deleted_at)
 `
 
 const psaPinRequestTableName = 'psa_pin_request'
@@ -173,7 +173,7 @@ export class DBClient {
 
   /**
    * Create a user tag
-   * @param {number} userId
+   * @param {string} userId
    * @param {Object} [tag]
    * @param {string} [tag.tag]
    * @param {string} [tag.value]
--- a/packages/api/test/fixtures/init-data.sql
+++ b/packages/api/test/fixtures/init-data.sql
@@ -97,7 +97,9 @@ VALUES (
 );
 
 INSERT INTO user_tag (user_id, tag, value, reason, deleted_at)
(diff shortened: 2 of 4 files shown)


e-schneid (approved):

packages/db/index.js
```diff
@@ -42,7 +42,7 @@ const userQueryWithTags = `
   publicAddress:public_address,
   created:inserted_at,
   updated:updated_at,
-  tags:user_tag_user_id_fkey(user_id,id,tag,value)
+  tags:user_tag_user_id_fkey(user_id,id,tag,value, deleted_at)
```
e-schneid, line 45: thanks for fixing this.  looks like I lost this field when resolving the merge conflicts from all the template strings being extracted into consts.

adamalton: 👍 no worries.

jsdevel: @adamalton how does this work?  _why_ does adding `deleted_at` here change anything?  `hasTag` checks for `!tag.deleted_at` before confirming the value.

jsdevel: @adamalton so i think i get it.  because `public.user` _does not contain_ a `deleted_at` column, the matching algorithm will require the tag's `deleted_at` value to be null?

adamalton: @jsdevel without this, the value of the `deleted_at` column isn't returned by the query, and so when `hasTag` does its filtering `tag.deleted_at` will always evaluate as false-y because `tag.deleted_at` is `undefined`.

jsdevel: if that is the case, then this doesn't really "fix" anything because the code is currently looking for `!tag.deleted_at` anyway...

jsdevel: ahh, makes sens.  thank you!

vasco-santos, line 45:
```suggestion
  tags:user_tag_user_id_fkey(user_id,id,tag,value,deleted_at)
```

vasco-santos (approved):
LGTM
good catch, just a minor space to keep consistency

PR #1384, opened 2022-05-30
fix: logger should not send logs to Logtail from CI

adamalton:
I've reworked the PR description, please find the original [here](https://github.com/web3-storage/web3.storage/pull/1384#issuecomment-1233038565)

---

This is a follow-up to Alan's comment here: https://github.com/web3-storage/web3.storage/pull/1346#issue-1244957912

## Acceptance Criteria
- [ ] Logs are sent to logtail from Staging and Prod env
- [ ] Exceptions are sent to Sentry from Staging and Prod env
- [ ] Logs are **not** sent to logtail from dev and CI
- [ ] Exceptions are **not** sent to Sentry from dev and CI

## Notes
Annoyingly I think lint-staged modified files that have nothing to do with this PR 🤔


olizilla (changes requested):
I think the fix here is to avoid trying to send logs to logtail when in DEV/TEST envs.

adamalton:
@olizilla I'm not sure exactly what you mean by your comment. Are you saying that we should go for a different fix in order to prevent the Logtail logging in dev/test?

flea89:
I've updated the original title and description to what I think was the original intent of Alan's comment.

For reference here's the old ticket details:

Original ticket: 
Title chore: set DEBUG to false in the 'test' env to silence logging debug statements
Description:  
This is a follow up to Alan's comment here: https://github.com/web3-storage/web3.storage/pull/1346#issue-1244957912

I've gone with changing the `DEBUG` env var to `'false'` in wrangler.toml, as the only other place where we compare that var to `'true'` is in [packages/api/src/env.js](https://github.com/web3-storage/web3.storage/blob/51abb35b9c6f53ed1023a7dbb162ce2838fe09e9/packages/api/src/env.js#L94) in the Sentry setup. But given that `SENTRY_DSN` isn't set for the `env.test`/local environment, Sentry won't be getting initialised anyway.  So I don't think this affects any other debug configuration.

An alternative would be to use the pattern of:

```js
import debug from 'debug'
log = debug('logging')
```

But given that there are already several methods and variables called `log` and `debug` in the logging.js file, including in the scopes where we would want to call `log()`, that might cause more confusion than it's worth! Plus doing it the way I've done it keeps the code more similar to the implementation that's in nft.storage.

joshghent (approved):

flea89:
@olizilla, I stumbled upon this PR and I've reworked it to tackle what was the original purpose.
Can please confirm that's the case 🙏 before we merge it in?

olizilla:
I'm +1 on the fix here...  but could you close this PR and open fresh one without the linting noise (I had to do the same here on old PR https://github.com/web3-storage/web3.storage/pull/1593#issuecomment-1237056568)

mbommerez:
New PR to be linked.

PR #1387, opened 2022-05-31
chore: fix typedefs for user tag functions

adamalton:
This just tidies up the typedefs for some of the user tag related functions where were changed/added in #1322, and also improves the function signature of the `createUserTag` function.

### DIFF
--- a/packages/api/src/utils/tags.js
+++ b/packages/api/src/utils/tags.js
@@ -1,7 +1,7 @@
 /**
  * @param {import('@web3-storage/db/db-client-types').UserOutput} user
  * @param {string} tagName
- * @param {string|undefined} defaultValue
+ * @param {string} [defaultValue]
  * @returns {string|undefined}
  */
 export function getTagValue (user, tagName, defaultValue) {
--- a/packages/api/src/utils/tags.js
+++ b/packages/api/src/utils/tags.js
@@ -1,3 +1,9 @@
+/**
+ * @param {import('@web3-storage/db/db-client-types').UserOutput} user
+ * @param {string} tagName
+ * @param {string|undefined} defaultValue
+ * @returns {string|undefined}
+ */
 export function getTagValue (user, tagName, defaultValue) {
   return (
     user.tags?.find((tag) => tag.tag === tagName && !tag.deleted_at)?.value ||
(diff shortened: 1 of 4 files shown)


packages/api/src/utils/tags.js
```diff
@@ -1,10 +1,22 @@
+/**
+ * @param {import('@web3-storage/db/db-client-types').UserOutput} user
+ * @param {string} tagName
+ * @param {string|undefined} defaultValue
```
flea89, line 4:
```suggestion
 * @param {string} [defaultValue]
```
While setting the type to `string|undefined` works I think this is the jsdoc way

packages/db/index.d.ts
```diff
@@ -76,8 +77,8 @@ export class DBClient {
   createContent (content: ContentInput, opt?: {updatePinRequests?: boolean}) : Promise<string>
   deleteKey (id: number): Promise<void>
   query<T, V>(document: RequestDocument, variables: V): Promise<T>
-  createUserTag(userId: number, tag: UserTagInput): Promise<boolean>
-  getUserTags (userId: number): Promise<{ tag: string, value: string }[]>
+  createUserTag(userId: string, tag: UserTagInput): Promise<boolean>
+  getUserTags(userId: string, options: GetUserOptions): Promise<{ tag: string, value: string }[]>
```
flea89, line 81: Can we please update both functions in `db/index.js`?

packages/db/index.js
```diff
@@ -174,13 +175,13 @@ export class DBClient {
...
-   * @param {Object} [tag]
-   * @param {string} [tag.tag]
-   * @param {string} [tag.value]
-   * @param {string} [tag.reason]
+   * @param {import('./db-client-types').UserTagInput?} tag
    * @returns {Promise<boolean>}
    */
-  async createUserTag (userId, tag = {}) {
+  async createUserTag (userId, tag) {
+    if (!tag || !tag.tag) {
```
flea89, line 182:
```suggestion
    if (tag?.tag) {
```

flea89 (changes requested):
Some minor comments/suggestions

packages/db/index.js
adamalton: Ah yes, but with `!`. I'll make this change.

flea89: ups yeah :)

packages/db/index.d.ts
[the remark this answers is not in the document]
adamalton: Done (with `[]` on the end).

adamalton: Fixed these, as discussed on Hangout.

PR #1390, opened 2022-05-31
feat: Adding HasDeleteRestriction user_tag

jsdevel:
* Adding the type and failing HTTP DELETE operations if this tag is set.
* See nftstorage/admin.storage#66


vasco-santos (approved):
LGTM

Issue #1380, opened 2022-05-30
website: about page collapsed section is haunted


olizilla:
The collapsed sections on the about page are rapidly opening and closing of their own volition. Probably haunted.

<details>
<summary><b>Flicker warning! Click to see! Defo ghosts 👻</b></summary>

![w3-about-chaos](https://user-images.githubusercontent.com/58871/171011994-37106cc7-3d76-44db-9226-fad02c4c435d.gif)

</details>

orvn:
@JeffLowe it should be fixed in #1391 already (in review), but just needs some testing to make sure it works on different devices/browsers!

PR #1391, opened 2022-05-31
fix: refactor accordion content

joannaong:
refactor accordion to use css instead of js to fix flickering bug

completes #1380


orvn (approved):
Tested on desktop and mobile, issues resolved

PR #1396, opened 2022-06-01
fix: clone env so new each request

alanshaw:
Not sure if this will fix the error but it's worth a try and feels unlikely to break anything.

Reading through [this](https://github.com/web3-storage/web3.storage/pull/1331) it sounds as though env is **not** a new object for each request 😱


orvn:
It's looking like the fix worked!

adamalton:
Late to the party, but I approve of this change!

PR #1405, opened 2022-06-06
chore: optimise getUserByStorage query

flea89:
1. “sharding” it, go through users in batches as we do for pins
2. Using a direct connection Read-Only postgres connection so that:
    - query do not add load to the production postgrest dyno
    - send query to replica DB, reduce load on prod DB
    - No 30s timeout heroku dyno imposes


packages/cron/src/jobs/storage.js
```diff
@@ -44,40 +71,51 @@ const STORAGE_QUOTA_EMAILS = [
...
  * appropriate when approaching their storage quota limit.
  * @param {{
- *  db: import('@web3-storage/db').DBClient
+ *  roPg: import('pg').Client
  *  emailService: import('../lib/email/service').EmailService
+ *  userBatchSize?: number
  * }} config
  */
-export async function checkStorageUsed ({ db, emailService }) {
+export async function checkStorageUsed ({ roPg, emailService, userBatchSize = 100 }) {
```
flea89, line 79: This could possibly be bigger, I wonder if 500 or a 1000 would be better?

.github/workflows/cron-storage-limits.yaml
```diff
@@ -31,6 +31,10 @@ jobs:
         env:
           DEBUG: '*'
           ENV: ${{ matrix.env }}
+          STAGING_PG_CONNECTION: ${{ secrets.STAGING_PG_CONNECTION }}
+          STAGING_RO_PG_CONNECTION: ${{ secrets.STAGING_PG_CONNECTION }} # no replica for staging
+          PROD_PG_CONNECTION: ${{ secrets.PROD_PG_CONNECTION }}
+          PROD_RO_PG_CONNECTION: ${{ secrets.PROD_RO_PG_CONNECTION }}
```
vasco-santos, line 37: looks like we only need RO?

packages/cron/src/jobs/storage.js
vasco-santos: A note here: we rely on user id's that came from Fauna. The bad news is that they are not sequential, so there are a lot of gaps, which is really annoying for this kind of use cases...

With that in mind, there are huge gaps sometimes between ids. So, I think we should start with 1000 as you suggest

packages/cron/test/storage.spec.js
```diff
@@ -42,11 +47,16 @@ describe('cron - check user storage quotas', () => {
...
+      email: 'testUser90percent1@email.com',
       percentStorageUsed: 90
     })
 
+    testUser90percent2 = await createUserWithFiles(dbClient, {
+      email: 'testUser90percent2@email.com',
+      percentStorageUsed: 91
+    })
+
     test4user = await createUserWithFiles(dbClient, {
```
vasco-santos, line 60: Can we also rename this user variable? Liked the namings you started to use as they are way more clear

vasco-santos (approved):
Great work @flea89 

LGTM, left some minor things for you to consider

.github/workflows/cron-storage-limits.yaml
flea89: Good catch!

packages/cron/src/jobs/storage.js
flea89: Updated! 👍

PR #1406, opened 2022-06-06
fix: cloudflare loader in #1361

hugomrdias:
This PR is based on #1361 and should be merged to it directly not main! The changes in this PR should serve as a reference to be applied to the rest of the code, see notes below. 

- fix css urls for illustrations, there was several bad urls in css should probably check if the visuals are still good!!
- new cloudflare image loader that works in dev, preview urls and production.
- new HeroIllustration comp with good defaults and for hero images 

Notes:
- All images should be imported and not hardcoded as string paths
- HeroIllustration comp should be used instead of the per image component
- Local images should not have hardcoded width and height, these should come from the images import
- Per illustration component should be deleted in favor of HeroIllustration or similar.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24218,7 +24218,7 @@
     },
     "packages/api": {
       "name": "@web3-storage/api",
-      "version": "5.7.3",
+      "version": "6.0.0",
       "license": "(Apache-2.0 OR MIT)",
       "dependencies": {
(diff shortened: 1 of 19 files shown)


orvn:
Thanks! Merging and testing within our branch (plus applying this format to other images).

PR #1381, opened 2022-05-30
chore: Remove deal polling copy

olizilla:
Removes the deal polling copy until we make it do that. Shuffles paragraph around to avoid breaking page layout

Fixes: #1042

| before | after |
|--------|------|
|Screenshot 2022-05-30 at 15 47 31 | Screenshot 2022-05-30 at 15 45 38

License: (Apache-2.0 AND MIT)
Signed-off-by: Oli Evans <oli@tableflip.io>

PR #1407, opened 2022-06-06
fix: filecoin deals about text

dchoi27:


yusefnapora:
@dchoi27 I was able to remove that section, but I had to also remove the images at left to avoid a broken layout.

Here's what it looks like after my commit:
2022-06-06 - web3-storage-about-changes

Vs the current page, showing the file upload interface, etc:

2022-06-06 - web3-storage-about-current

There might be a way to keep the images without coming up with some replacement text, but someone more familiar with the code could probably figure it out faster than me 😄

olizilla:
@dchoi27 please close out https://github.com/web3-storage/web3.storage/pull/1381 if it's not the change you want.

Issue #1362, opened 2022-05-26
File load & pagination issues for user with many files


mbommerez:
## User observations
* Users with lots of uploads in web3.storage experience really slow loading times.
* Users with > 1000 uploads are not currently able to see them all in the web interface. It stops at 100 pages of 10 files (loads 1000 first ones only).

## Acceptance criteria
* As a user I’m able to navigate through all my uploads 
* Loading uploads should be reasonably quick

## Notes

There are 2 parts to this:
* Have proper paginations to the page ( not loading 1000 items and then paginate in memory)
* Look into query optimisation
* This was first raised and looked into for ticket [1147](https://github.com/web3-storage/web3.storage/issues/1147)

https://images.zenhubusercontent.com/378893240/704a835d-8695-4f22-853d-85ad199a1058/screen_recording_2022_03_21_at_10_38_42_720p.mp4

joshJarr:
Done some initial investigation on this, there's a few ways to approach this task and am keen to get some feedback before tackling it!

Look like right now 1000 results are requested by the frontend, Pagination, searching and sorting are completed on the Frontend.

If we want to keep all "sort by" and "search" functionality, we would have to replicate this in the API.

**Pagination**
- To show the numerical "pages" of the results, we may have to change the API response to also deliver some metadata on total uploads for the frontend to calculate the amount of pages. This change could be avoided if we were to simply show "next" and "previous" instead of numerical total pages.
- I'd also like to add an `offset` param to the endpoint that works with the existing `size` param to chunk results by "page".

**Sorting**
- Looks like the API already supports sorting by Date and Name, meaning we'd also have to add sorting by size to this. 
- We could use the existing `before` and `after` parameters to paginate through the API. I'm not a huge fan of this approach since it means we cannot sort by size or name since we'd be chunking the uploads by date.
Additionally, since date/time stamps are not unique, an edge case could occur in which two or more uploads with the same upload time stamp (an unlikely, event, unless a user is uploading files in parallel). Paginating by datetime with duplicated timestamps could result in missing uploads listed by the API. 

**Searching**
- I find searching a little odd, we currently allow searching by CID, but would a user ever search for results using a partial CID?
- Instead we could add a `CID` parameter to the `GET user/uploads` endpoint to only return a single response, or add a new endpoint of `/user/uploads/<CID>` to fetch a single result.

[...]

PR #1408, opened 2022-06-06
fix: get user/uploads endpoint pagination

joshJarr:
**Changes**
Adds pagination to the `GET user/uploads` endpoints on the API.
- Adds an `page` param to allow pagination when used with `size`.
- Adds tests for the `sortBy` param
- Adds tests for the `before` and `after` filtering params
- Adds missing documentation for the `after`, `sortBy` and `sortOrder` params

This PR should not introduce a breaking change for the API, however does change the outputs of the db `listUploads` method.
This PR does not regress any existing functionality or change any existing API functionality. The existing client and website packages should still work as intended.

**Related tickets:**
- master ticket https://github.com/web3-storage/web3.storage/issues/1362

### DIFF
--- a/packages/api/src/user.js
+++ b/packages/api/src/user.js
@@ -222,6 +222,7 @@ export async function userTokensDelete (request, env) {
 }
 
 const sortableValues = ['Name', 'Date']
+const sortableOrders = ['Asc', 'Desc']
 
 /**
  * Retrieve a page of user uploads.
@@ -274,8 +275,12 @@ export async function userUploadsGet (request, env) {
   const sortBy = searchParams.get('sortBy') || 'Date'
   const sortOrder = searchParams.get('sortOrder') || 'Desc'
 
+  if (!sortableOrders.includes(sortOrder)) {
+    throw Object.assign(new Error(`Sort ordering by '${sortOrder}' is not supported. Supported sort orders are: [${sortableOrders.toString()}]`), { status: 400 })
+  }
+
   if (!sortableValues.includes(sortBy)) {
-    throw Object.assign(new Error(`Sorting by '${sortBy}' is not supported.`), { status: 400 })
+    throw Object.assign(new Error(`Sorting by '${sortBy}' is not supported. Supported sort orders are: [${sortableValues.toString()}]`), { status: 400 })
   }
 
   const data = await env.db.listUploads(request.auth.user._id, {
@@ -290,12 +295,12 @@ export async function userUploadsGet (request, env) {
   let link = ''
   // If there's more results to show...
   if (page > 1) {
-    link += `<${requestUrl.pathname}?size=${size}&page=${encodeURIComponent(page - 1)}>; rel="previous"`
+    link += `<${requestUrl.pathname}?size=${size}&page=${page - 1}>; rel="previous"`
   }
 
   if (data.uploads.length + offset < data.count) {
     if (link !== '') link += ', '
-    link += `<${requestUrl.pathname}?size=${size}&page=${encodeURIComponent(page + 1)}>; rel="next"`
+    link += `<${requestUrl.pathname}?size=${size}&page=${page + 1}>; rel="next"`
   }
 
   const headers = {


packages/api/test/user.spec.js
```diff
@@ -229,11 +303,28 @@ describe('GET /user/uploads', () => {
...
+  it('does not paginate when all results are returned', async () => {
+    const token = await getTestJWT()
+    const size = 1000
+    const res = await fetch(new URL(`/user/uploads?size=${size}`, endpoint).toString(), {
+      method: 'GET',
+      headers: { Authorization: `Bearer ${token}` }
+    })
+    assert(res.ok)
+
+    assert.notEqual('link' in res.headers, 'does not have a Link header for the next page')
```
adamalton, line 324: I'm guessing this is a typo and should be `assert.not` (or similar)?

adamalton (commented):
This looks really good.

My only overall thought is about naming. To me it would make sense to have the parameters called `page` and `pageSize` rather than `offset` and `size. Intuitively for me, if I specified `offset=10`, I would expect to get the 10th or 11th _items_ and onwards, rather than the 10th/11th _page_. And "size" is confusable with the size of the upload, rather than the size of the page.

I realise that the `size` parameter already exists though, so would require an update to the FE, so maybe the answer for changing that is "yeah, but not right now".

packages/api/src/user.js
```diff
@@ -187,29 +187,49 @@ export async function userUploadsGet (request, env) {
     size = parsedSize
   }
 
-  let before = new Date()
+  let offset = 0
+  if (searchParams.has('offset')) {
+    const parsedOffset = parseInt(searchParams.get('offset'))
+    if (isNaN(parsedOffset) || parsedOffset <= 0 || parsedOffset > 1000) {
```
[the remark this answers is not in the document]
joshJarr: That's a really good point... I just copied the same validation from the `size`! I can totally kill this limit

packages/db/index.js
```diff
@@ -484,8 +499,13 @@ export class DBClient {
       query = query.gte('inserted_at', opts.after)
     }
 
-    /** @type {{ data: Array<import('./db-client-types').UploadItem>, error: Error }} */
-    const { data: uploads, error } = await query
+    // Apply pagination or limiting.
+    query = opts.offset
+      ? query.range(offset, limit - 1)
```
joshJarr, line 504: Looks like this is what I want, but curious as to if anyone knows a better way of doing this? 
Am unsure of the performance impact of using a range()...

packages/db/index.js
```diff
@@ -457,25 +457,40 @@ export class DBClient {
...
+    const sortBy = opts.sortBy === 'Name'
+      ? 'name'
+      : 'inserted_at'
+
+    const isAscendingSortOrder = opts.sortOrder === 'Asc'
+
     let query = this._client
       .from('upload')
-      .select(uploadQuery)
+      .select(uploadQuery, { count: 'exact' })
```
joshJarr, line 485: I _think_ I need an exact count here, but unsure how this would impact the query performance, open to suggestion as to how I could improve this!

packages/website/public/schema.yml
```diff
@@ -416,6 +425,32 @@ components:
...
+        type: string
+        enum: [Date, Name]
+    sortOrder:
+      description: Sorts results by a given parameter.
+      name: sortOrder
+      in: query
+      required: false
+      schema:
+        type: string
+        enum: [Asc, Desc]
```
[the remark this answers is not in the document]
joshJarr: Yes! Great spot, thanks

packages/api/src/user.js
```diff
@@ -187,30 +187,62 @@ export async function userUploadsGet (request, env) {
...
   const sortBy = searchParams.get('sortBy') || 'Date'
   const sortOrder = searchParams.get('sortOrder') || 'Desc'
 
-  const uploads = await env.db.listUploads(request.auth.user._id, {
+  const data = await env.db.listUploads(request.auth.user._id, {
     size,
-    before: before.toISOString(),
+    offset,
+    before,
+    after,
```
joshJarr, line 224: I added the `after` param for filtering. Also added a test for this.
Looks like this functionality has always existed but never exposed on the API which confused me a bit.

packages/api/src/user.js
```diff
@@ -187,30 +187,62 @@ export async function userUploadsGet (request, env) {
...
+  }
+
+  if (data.uploads.length + offset < data.count) {
+    const nextOffset = offset + size
+    headers.Next_link = `<${requestUrl.pathname}?size=${size}&offset=${encodeURIComponent(nextOffset)}>; rel="next"`
+  }
+
+  if (offset) {
+    const previousOffset = offset - size
+    headers.Prev_link = `<${requestUrl.pathname}?size=${size}&offset=${encodeURIComponent(previousOffset)}>; rel="next"`
```
joshJarr, line 242: Thoughts of including this? 
Since we have a next link it seemed to make sense to also include the previous link.
Though I can see an argument for not including links at all.

packages/api/src/user.js
```diff
@@ -187,30 +187,62 @@ export async function userUploadsGet (request, env) {
...
-  const oldest = uploads[uploads.length - 1]
-  const headers = uploads.length === size
-    ? { Link: `<${requestUrl.pathname}?size=${size}&before=${encodeURIComponent(oldest.created)}>; rel="next"` }
-    : undefined
-  return new JSONResponse(uploads, { headers })
+  const headers = {
+    Count: data.count,
+    Size: size,
+    Offset: offset
+  }
```
joshJarr, line 233: Keen to get some feedback on including pagination meta data in the header. 
I could include this info in the body, but since it's only used for paginating it _could_ make sense to be in the header.

packages/api/src/user.js
```diff
@@ -187,30 +187,62 @@ export async function userUploadsGet (request, env) {
...
+
+  if (data.uploads.length + offset < data.count) {
+    const nextOffset = offset + size
+    headers.Next_link = `<${requestUrl.pathname}?size=${size}&offset=${encodeURIComponent(nextOffset)}>; rel="next"`
```
joshJarr, line 237: What's the accepted syntax for this? Looks like capitalising the first letter is generally expected, but should this be snake_case?

packages/api/test/user.spec.js
```diff
@@ -209,6 +209,80 @@ describe('GET /user/uploads', () => {
...
+
+    const uploads = await res.json()
+
+    assert(uploads.length < userUploads.length, 'Ensure some results are filtered out.')
+    assert(uploads.length > 0, 'Ensure some results are returned.')
+
+    // Filter uploads fixture by the filter date.
+    const uploadsBeforeFilterDate = userUploads.filter((upload) => {
+      return upload.created <= beforeFilterDate
+    })
```
joshJarr, line 257: I generally don't like this kind of logic in tests, and perhaps could simply expect something like 3 results...
But seems like these fixtures are shared across tests, so didn't want this test to break when another result is added in the uploads fixture. 
Thoughts on this?

packages/website/public/schema.yml
```diff
@@ -416,6 +425,34 @@ components:
...
+        enum: [Date, Name]
+    sortOrder:
+      description: Sorts the order of the results.
+      name: sortOrder
+      in: query
+      required: false
+      default: Desc
+      schema:
+        type: string
+        enum: [Asc, Desc]
```
joshJarr, line 456: Decided to document this previously hidden functionality as it confused me when initially looking. Is this ok? Or are we hiding `sortBy`, `sortOrder` and `after` for a reason?

packages/api/test/user.spec.js
joshJarr: Looked at this, `.not` is not available when using node's `assert()`. I changed this to be `.notStrictEqual` as `.notEqual` [seems to be legacy](https://nodejs.org/api/assert.html#assertnotequalactual-expected-message). 

I could actually simply inverse the bool with `!` and just `assert(!'link' in res.headers)` it..
But I like how declarative `notEqual` is and how a failure is reported.. then again that's what that optional message field is for.🤔 
Am open to suggestion on how best to do this!

adamalton: I was just meaning that currently you're not supplying a message, you're comparing the boolean result of `'link' in result.headers` to the string `'does not have a Link...'`, so they will never be equal, regardless of whether `'link'` is in `res.headers` or not!

packages/api/test/user.spec.js
adamalton: I think that the way you've done this is probably the best way given that we're using fixtures that are defined elsewhere. Like you say, we don't want the test to break (or even worse, the code to break and the test to not catch it) if/when someone changes the fixtures.

I think that the real solution to this is for us to move away from using fixtures for tests, and to instead explicitly create the objects that we want within each test. We're starting to do this in places. This is probably a good example (creating a user tag via the DB client and then testing the API endpoint that fetches that info): https://github.com/web3-storage/web3.storage/blob/main/packages/api/test/user.spec.js#L52-L65

packages/db/index.js
adamalton: AFAIK the `.range()` method translates to the SQL `LIMIT a OFFSET b`. So I think it should be fine. The only other way I can think of doing it would be if we used the column that we're sorting by (e.g. `inserted_at`) and took the highest value from the previous page and then did `query.gt('inserted_at', prev_highest)`, which would presumably be very very sliiiightly faster because it would save the DB from having to count and skip the first `offset` rows that it finds. But given that we're only talking about the uploads for a single user here, I can't imagine that we're likely to get to the number of rows where this difference actually matters.

What probably _is_ worth doing is checking that we've got indexes for the columns that we're filtering and sorting on.

packages/website/public/schema.yml
adamalton: Seems good to me. And given that the whole project is open source, if we're trying to hide it then that isn't going to work for long 😂.

packages/api/test/user.spec.js
joshJarr: Ops, yep that's a great spot, my bad! Changed this to a `.ok` for now, thanks!

joshJarr:
Am now using `page` over `offset` in the `GET /user/uploads` endpoint.

packages/db/index.js
flea89: This is definitively correct.
This could probably be optimised using `estimate`, however depending on how pagination has been implemented that might complicate a bit the logic.
So I'd monitor performance for big lists and update this if required.
What do you think?

packages/db/index.js
```diff
@@ -506,24 +506,40 @@ export class DBClient {
     }
   }
 
+  /**
+   * @typedef {Object} listUploadReturn
```
flea89, line 510: Can we define this type in `db-client-types.ts`, and import it where needed?

We need to make sure not only the functions annotations are updated here, but also in `index.d.ts`.

I'd keep the same convention where the type is a plain (not promise one).
and the function returns `Promise<thetype>`

nit. should be capitalised, and maybe called `ListUploadOutput` for consistency.

packages/db/index.js
```diff
@@ -506,24 +506,40 @@ export class DBClient {
...
    * @param {number} userId
    * @param {import('./db-client-types').ListUploadsOptions} [opts]
-   * @returns {Promise<Array<import('./db-client-types').UploadItemOutput>>}
+   * @returns {listUploadReturn}
    */
   async listUploads (userId, opts = {}) {
+    const offset = opts.offset || 0
+    const limit = offset + (opts.size || 25)
+
+    const sortBy = opts.sortBy === 'Name'
```
flea89, line 526: I wonder if it'd be better to raise if we try to sort on an unsupported column instead?

packages/api/src/user.js
```diff
@@ -240,37 +240,62 @@ export async function userUploadsGet (request, env) {
...
+  let after
+  if (searchParams.has('after')) {
+    const parsedAfter = new Date(searchParams.get('after'))
+    if (isNaN(parsedAfter.getTime())) {
+      throw Object.assign(new Error('invalid after date'), { status: 400 })
+    }
+    after = parsedAfter.toISOString()
   }
 
   const sortBy = searchParams.get('sortBy') || 'Date'
```
flea89, line 272: I'd be quite nice to validate theses, and return the right errors for unsupported values.
We use declaritive validators for the pin endpoint and we should probably port that to the uploads.
I think this is outside of the scope of the PR though, what do you think?

flea89 (changes requested):
Nice work. Left some comments.

Some additional thoughts:
- Being a breaking change for the API, we need to make sure we mark the PR accordingly (see convention [here](https://www.conventionalcommits.org/en/v1.0.0/)) 
- I think I've seen a ticket for updating the website, do we have one for the client library?

packages/api/test/user.spec.js
joshJarr: Thanks! I totally agree, moving away from fixtures makes a huge amount of sense.
I think I'll leave this for now, so the upload tests can all use fixtures consistently, but it would make sense to update this at some point!

packages/db/index.js
joshJarr: Yep, this makes sense - would be good to keep an eye on this when testing on staging. Thanks!

packages/api/src/user.js
```diff
@@ -240,37 +240,62 @@ export async function userUploadsGet (request, env) {
...
-      : undefined
-  return new JSONResponse(uploads, { headers })
+  const headers = {
+    Count: data.count,
+    Size: size,
+    Page: page
+  }
+
+  if (data.uploads.length + offset < data.count) {
+    headers.Next_link = `<${requestUrl.pathname}?size=${size}&page=${encodeURIComponent(page + 1)}>; rel="next"`
```
flea89, line 291: Shouldn't this be speficied in `Link` header only?
Something like:
```
Link: <prevLink>; rel="prev",
  <theNexLink>; rel="next"
```?

File: packages/api/test/user.spec.js [hunk 16]
```diff
@@ -220,20 +294,53 @@ describe('GET /user/uploads', () => {
...
       method: 'GET',
       headers: { Authorization: `Bearer ${token}` }
     })
     assert(res.ok)
 
-    const expected = [userUploads[0]]
-    const link = res.headers.get('Link')
-    assert(link, 'has a Link header for the next page')
-    assert.strictEqual(link, `</user/uploads?size=${size}&before=${encodeURIComponent(expected[0].created)}>; rel="next"`)
+    assert.ok('link' in res.headers === false)
```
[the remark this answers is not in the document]
joshJarr: Great shout!
So I actually ensure the header metadata is tested in the `paginates by page` test just above, lines [308 => 323](https://github.com/web3-storage/web3.storage/pull/1408/files/7342533470f136a5c867f1c40535bb6c6ab20e75#diff-fd06e8f47f44010478243d794d7e859dc226dbda62aa8796511d5b86bc00d207R308). 

But here I'm expecting the `Link` to not exist, as in this test there is no more pages... but that's a problem, as I renamed `Link` to `Next_link`! So I'll update this.

Thanks!

packages/api/test/user.spec.js
```diff
@@ -209,6 +209,80 @@ describe('GET /user/uploads', () => {
...
 
+  it('lists uploads sorted by date', async () => {
+    const token = await getTestJWT()
+    const res = await fetch(new URL('/user/uploads?sortBy=Date', endpoint).toString(), {
+      method: 'GET',
+      headers: { Authorization: `Bearer ${token}` }
+    })
+    assert(res.ok)
+    const uploads = await res.json()
+    assert.deepStrictEqual(uploads, [...userUploads].sort((a, b) => b.created.localeCompare(a.created)))
```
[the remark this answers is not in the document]
flea89: yeah, just felt weird, but of course, you're right ISO 8601 sorts lexicographically. 
Let's ignore this then. 🙌

packages/db/index.js
```diff
@@ -506,24 +506,40 @@ export class DBClient {
...
    * List uploads of a given user.
    *
    * @param {number} userId
    * @param {import('./db-client-types').ListUploadsOptions} [opts]
-   * @returns {Promise<Array<import('./db-client-types').UploadItemOutput>>}
+   * @returns {listUploadReturn}
    */
   async listUploads (userId, opts = {}) {
+    const offset = opts.offset || 0
+    const limit = offset + (opts.size || 25)
```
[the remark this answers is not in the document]
joshJarr: Ooo great shout!
changing this to `offset => rangeFrom` and `limit => rangeTo` as I can effectively remove the `.limit` query as per your suggestion further down on this file.
Thanks!

packages/db/index.js
```diff
@@ -532,8 +548,13 @@ export class DBClient {
       query = query.gte('inserted_at', opts.after)
     }
 
-    /** @type {{ data: Array<import('./db-client-types').UploadItem>, error: Error }} */
-    const { data: uploads, error } = await query
+    // Apply pagination or limiting.
+    query = opts.offset
+      ? query.range(offset, limit - 1)
+      : query.limit(limit)
```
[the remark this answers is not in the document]
joshJarr: Yep! Totally right, I wasn't sure if there was any performance bennifit of using `.limit` over .`range`. But it essentially does the same so will just keep to the `.range`.
Thanks!

packages/api/src/user.js
joshJarr: Nice! Cool yeah agreed.
I added some validation below this however, now it'll validate on the API side if theres an unsupported `sortBy` param rather than borking the DB.

packages/api/src/user.js
joshJarr: Great point! I assumed `Link` was a custom response header, but it turns out it's legit! Learn something new every day!
Thanks

packages/api/test/user.spec.js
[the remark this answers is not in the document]
joshJarr: As pre another comment, changed `Next_link` back to `Link`

packages/db/index.js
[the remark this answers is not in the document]
joshJarr: I've got this validation on the API now, I could also add this validation here? But maybe it feels a bit duplicated. Happy to add it here too if you like?

packages/website/public/schema.yml
joshJarr: Great points! Addressed these issues now @flea89 
Thanks

joshJarr:
> * ~Being a breaking change for the API, we need to make sure we mark the PR accordingly (see convention [here](https://www.conventionalcommits.org/en/v1.0.0/))~  This is not entirely accurate, it is for the DB but not for the API.
>   Given the PR will trigger a release please for both packages I'm not sure what's the best approach here. Being the APIs the user facing one maybe this is ok?

Yeah this is a great point, what would be deployed first? 
I could look at coding defensively around this, testing the db response and changing how we handle this in the api. Do we feel this is worth it for no downtime? or does releasing them both in tandem prevent any issues?

> * I think I've seen a ticket for updating the website, do we have one for the client library?
I do not have one for the Client library, will make one now, thanks!

packages/db/postgres/tables.sql
```diff
@@ -271,6 +271,8 @@ CREATE INDEX IF NOT EXISTS upload_user_id_deleted_at_idx ON upload (user_id) WHE
 CREATE INDEX IF NOT EXISTS upload_content_cid_idx ON upload (content_cid);
 CREATE INDEX IF NOT EXISTS upload_updated_at_idx ON upload (updated_at);
 CREATE INDEX IF NOT EXISTS upload_user_id_idx ON upload (user_id);
+CREATE INDEX IF NOT EXISTS upload_name_idx ON upload (name);
+CREATE INDEX IF NOT EXISTS upload_inserted_at_idx ON upload (inserted_at);
```
adamalton, line 274: Are these only added for the purposes of sorting after already filtering by `user_id`? If so, I have a suspicion that Postgres might not actually use them, because by the time it's reduced the result set down to the rows that match the `user_id` it _might_ deem that it's then just as quick to sort the actual rows as it is to fetch the relevant index entries and cross reference them against the result rows.

I'm not sure without actually trying it out on some real data. It's just a thought.

joshJarr: 👍 yep! Would be great to test this with real data.
Is there anyway to export data to staging to test this before going live?

adamalton: I'm not sure. I suspect that for a few rows it might just sort the rows, but when it gets to 1000s of rows it might switch to using the index.

For optimal speed for each possible sorting (name asc, name, desc, date asc, date, desc) I think it would need composite indexes for each of:
* `user_id`, `name` ASC
* `user_id`, `name` DESC
* `user_id`, `inserted_at`, ASC
* `user_id`, `inserted_at`, DESC

But that's getting to be quite a few indexes.

Presumably, even if a user has several thousand uploads, sorting them in memory is probably fairly quick (compared to finding the uploads from the millions in the table). So I'm kind of leaning towards thinking that so long as we've got a `user_id` index, the other indexes are unlikely to make much difference! But I might be wrong.

joshJarr: This makes sense to me! Reverted the index changes for now. 
Thanks

flea89 (approved):

packages/api/src/cors.js
```diff
@@ -22,7 +22,7 @@ export function corsOptions (request) {
       // such as Authorization (Bearer) or X-Client-Name-Version
       'Access-Control-Allow-Headers':
         headers.get('Access-Control-Request-Headers') || '',
-      'Access-Control-Expose-Headers': 'Link'
+      'Access-Control-Expose-Headers': 'Link, Count, Page, Size'
```
alanshaw, line 25: These are not standard HTTP headers to my knowledge.

packages/api/src/user.js
```diff
@@ -240,37 +243,74 @@ export async function userUploadsGet (request, env) {
...
-          }?size=${size}&before=${encodeURIComponent(
-            oldest.created
-          )}>; rel="next"`
-        }
-      : undefined
-  return new JSONResponse(uploads, { headers })
+  let link = ''
+  // If there's more results to show...
+  if (page > 1) {
+    link += `<${requestUrl.pathname}?size=${size}&page=${page - 1}>; rel="previous"`
```
alanshaw, line 298: You need to include _all_ querystring parameters in this URL, otherwise clients will lose context when following the `Link` header for pagination.

e.g. setting `sortBy` for the first request, then using the URL in `Link` to get the next page.

packages/api/src/user.js
alanshaw: It needs to be included in the URL of the `Link` header. This is how the JS client paginates.

packages/api/src/cors.js
joshJarr: Nope, these are custom.

I think we could instead use [`range` ](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range). 

Alternatively we could include pagination meta in the response body, this would be a breaking change so deliberately avoided this in favour of using the headers for now.

packages/db/index.js
```diff
@@ -506,24 +506,40 @@ export class DBClient {
...
+
   /**
    * List uploads of a given user.
    *
    * @param {number} userId
    * @param {import('./db-client-types').ListUploadsOptions} [opts]
-   * @returns {Promise<Array<import('./db-client-types').UploadItemOutput>>}
+   * @returns {listUploadReturn}
    */
   async listUploads (userId, opts = {}) {
```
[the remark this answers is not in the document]
alanshaw: Yes, we _really_ needed that.

packages/api/src/user.js
joshJarr: Great catch. I have made a quick ticket for this: https://github.com/web3-storage/web3.storage/issues/1737

PR #1411, opened 2022-06-06
feat: font optimization

hugomrdias:
Reduce number of weights and font size using woff2 and latin lang subsets. 

Left the some fonts commented if we are really need those we can use https://transfonter.org/ to optimize.


olizilla:
@hugomrdias any idea what's going on with this https://github.com/web3-storage/web3.storage/pull/1411/files#file-packages-website-tests-homepage-e2e-spec-ts-L1

How did you make the subset fonts? Would be good to have a script or something we can capture so we know how to do the process again if we need to change them. If it's fully manual, then just describing the steps on this PR would be ok.

orvn (approved):
Did a cross-browser smoke test to ensure that everything loads up correctly and it seems to work as expected! (bold/italics looks good in different contexts).

orvn:
> _Any idea what's going on with this https://github.com/web3-storage/web3.storage/pull/1411/files#file-packages-website-tests-homepage-e2e-spec-ts-L1_

No idea what going on with that, but it's been happening for a week or two (I just thought new E2Es were being worked on).

hugomrdias:
> @hugomrdias any idea what's going on with this https://github.com/web3-storage/web3.storage/pull/1411/files#file-packages-website-tests-homepage-e2e-spec-ts-L1

No idea

> How did you make the subset fonts? Would be good to have a script or something we can capture so we know how to do the process again if we need to change them. If it's fully manual, then just describing the steps on this PR would be ok.

I used https://transfonter.org/ and selected woff2 and latin subset, all the other options just used the default values.

PR #1415, opened 2022-06-07
feat: blog

joannaong:
## New blog section of the website

A blog section of the website, including navigation, blog index, blog singular pages (single articles), search and filters, tags and categories, styling and interactions.

> _Note that each of the items above and below aren't simple tasks, they are epic-sized_ (e.g., 5, 8, or 13 story points)

- [x] Add blog index page (and all related features)
- [x] Add blog singular page (and all related features)
- [x] Add all markdown elements
- [x] Design review

## Blocked
- [ ] Hook up subscribe API (requires endpoint)
- [ ] Update content and images (requires content)
- [ ] Integrate CMS layer (possibly Forestry.io, but unconfirmed, currently blocked)


orvn (changes requested):
@joannaong I did a pretty thorough review and have a some changes:

## Blog index

- [x] 1. Spacing between tags (visible in the tag modal when tags are selected)
tags-margin

- [x] 2. Subscribe button contrast is low, let's try reversing the direction of the gradient? (cc @emilymariahh for any other ideas)
subscribe-contrast

- [ ] 3. Get param for tag array format: currently, when selecting tags, we get an array format including square brackets and quotes `blog/?tags=["dao","arweave"]`. Can we clean this up to look like `blog/?tags=dao,arweave` and also make sure multi-word tags and categories are slugified (space delimited/kebab case), e.g., `blog/?tags=the-internet,foo,web3-storage`

- [x] 4. Add a transition to the pagination hover and active states

- [ ] 5. Pagination change causes a flash with momentary layout shift when the route changes (see video)

https://user-images.githubusercontent.com/3689720/181076424-dfb88515-7794-4bca-b946-a6e02c854b25.mp4

## Singular blog post

- [x] 6. Append ` - Web3 Storage` to the end of singular post `<title>`s (like on other pages on the site)

- [ ] 7. Unable to select heading at the top of the singular blog post. It's possible some background elements are actually above it, in which case we should prevent them from being selected with pointer events.

- [x] 8. Each nested layer of lists (both `<ul>` and `<ol>`) should be indented such that the bullet point itself is also inset (not just the text). See the bullet points/list styles in [docs for reference](https://web3.storage/docs/#next-steps).
bullet-points

- [x] 9. Link hover effects are missing, the color change effect [in the docs](https://web3.storage/docs) is okay to use for this

- [ ] 10. If a link in the blog content is external (not `https?://web3.storage.*`), it should _always_ open in a new tab

- [ ] 11. Captions under images have too much line height. I notice we're inheriting an exact pixel value for the line height. Instead, can we add in a line-height for the singular blog page that applies to all content elements? Something like `line-height: 2.125` will keep what's there right now, but also help future proof from these issues.
caption-line-height

- [x] 12. Headings should be `<h2>`s, i.e., the page should only ever have one H1 heading (the name of the page itself), so all subsequent headings for major sections should be `<h2>` elements
heading-h2

- [ ] 13. Tables aren't rendering in the markdown (added a table in this page to test `Announcing: NFT.Storage Gateway SuperHot perma-cache`)

- [ ] 14. Linkedin share button leads to a malformed (404) Linkedin page (all other share options work though!)

- [ ] 15. Decrease single post grid width: for better readability, let's decrease the content area width for the singular posts. Below I have it set to `45rem` (only the content area requires this, it should not affect the header, nav or other pages)
single-width


## Common or global

- [ ] 16. Equalize cards: we want to make sure that each card is the same length (i.e., that all `read more`'s align vertically in the same row). We can do this by setting a min-height on the card headings or by setting a height on the cards and using flex.
equalize-read-more-cards

- [ ] 17. Contact footer icons are missing a hover effect

joannaong:
@orvn 

> 14. Linkedin share button leads to a malformed (404) Linkedin page (all other share options work though!)

 
It seems like linkedin does not like sharing sites that ends in `.dev`.

nft.storage dev sites has the same issue. See https://feat-add-rarible-logo.nft-storage-1at.pages.dev/blog/post/2022-07-06-elastic-ipfs/ but it works on the production site.

cmunns (commented):
Looks pretty good.  A couple minor things:

- I couldn't get the post nav or pagination to consistently scroll to the top after clicking it.
- for post images we could probably ditch the padding-top hack in favor of [aspect-ratio](https://caniuse.com/mdn-css_properties_aspect-ratio)
- similar minor css suggestion would be to swap flex for grid to avoid having to add breakpoint specific odd/even margin rules.
- Is there a reason that Pagination component has an uppercase class of ".Pagination"? Ideally we naming consistent but didn't know if there was a reason here.

cmunns:
Also found a weird flash/glitch (firefox) when you click on copy link.
https://user-images.githubusercontent.com/1189523/187780744-7c157e20-4d38-4953-a344-b9c21f877658.mov

packages/website/pages/blog/post/[slug].js
```diff
@@ -0,0 +1,302 @@
...
+  }, [posts, post]);
+
+  /*
+   * Related Posts
+   *
+   * Add in most recent posts of same category first
+   * If related posts < 4 still, add in posts with matching tags (sorted by most tags which match first)
+   * If related posts < 4 still, add in posts at random from the remaining pool
+   */
+  const relatedPosts = useMemo(() => {
```
Comment (cmunns) [inline packages/website/pages/blog/post/[slug].js:140]: This is a pretty big file, could we split out some of these into separate files to clean it up a bit?

JeffLowe:
Closing this out, since blog code now lives in the content repo.

PR #1418, opened 2022-06-08
chore: use NODE_TLS_REJECT_UNAUTHORIZED=0 env var for storage cron job

adamalton:
As with the other cron jobs, without this we get a connection error.
We might want to address this more properly at some point.


vasco-santos:
Note: this should be a fix to create a cron release

adamalton:
I've created #1451 to review the use of this flag.

PR #1419, opened 2022-06-08
fix: change batching strategy for user storage cron

adamalton:
Previously we were using ID ranges. But the IDs of rows in the `public.user` table are not consecutive/sequential. So when we tried this on staging it went a bit like this:

1. Lowest user ID is 5, highest user ID is 314286764840715234.
2. It splits that range into chunks of 1000, giving us over 300 trillion batches to work through, just to iterate over 79 users.
3. The internet melts.

This PR changes the batches strategy to the form `SELECT ... WHERE id > highest_id_of_last_batch LIMIT users_per_batch`., which will hopefully go better.

Given that the `015` migration hasn't yet been run on production, I've modified it, rather than making a new one. It's already been run on staging, but running the modified version should be fine.

### DIFF
--- a/packages/cron/src/jobs/storage.js
+++ b/packages/cron/src/jobs/storage.js
@@ -9,6 +9,10 @@ const USER_BY_USED_STORAGE_QUERY = `
   SELECT * FROM users_by_storage_used($1, $2, $3, $4)
 `
 
+const MAX_USER_ID_QUERY = `
+ SELECT max(id)::TEXT as max from public.user
+`
+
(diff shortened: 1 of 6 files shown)


adamalton:
Ok, after having to debug via the CI due to Docker not working locally, I hereby present the messiest PR ever!

flea89 (approved):
This looks good! 
🤞 it's enough to have the cron run smoothly.

I've created a follow up ticket #1479  to further optimise this.

PR #1430, opened 2022-06-08
Fix/user upload storage used

### LINKED ISSUE #1198: Users uploaded storage used should only count content that is pinned

flea89:
Resolves #1198

### DIFF
--- a/packages/db/postgres/functions.sql
+++ b/packages/db/postgres/functions.sql
@@ -256,11 +256,20 @@ DECLARE
 BEGIN
   uploaded :=
     (
-      SELECT COALESCE(SUM(c.dag_size), 0)
-      FROM upload u
-      JOIN content c ON c.cid = u.content_cid
-      WHERE u.user_id = query_user_id::BIGINT
-      AND u.deleted_at is null
+      SELECT COALESCE((
+        SELECT SUM(dag_size) 
+        FROM (
+          SELECT  c.cid,
+                  c.dag_size
+          FROM upload u
+          JOIN content c ON c.cid = u.content_cid
+          JOIN pin p ON p.content_cid = u.content_cid
+          WHERE u.user_id = query_user_id::BIGINT
+          AND u.deleted_at is null
+          AND p.status = 'Pinned'
+          GROUP BY c.cid,
+                  c.dag_size
+        ) AS uploaded_content), 0)
     );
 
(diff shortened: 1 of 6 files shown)


packages/db/postgres/functions.sql
```diff
@@ -256,11 +256,20 @@ DECLARE
...
+          SELECT  c.cid,
+                  c.dag_size
+          FROM upload u
+          JOIN content c ON c.cid = u.content_cid
+          JOIN pin p ON p.content_cid = u.content_cid
+          WHERE u.user_id = query_user_id::BIGINT
+          AND u.deleted_at is null
+          AND p.status = 'Pinned'
+          GROUP BY c.cid,
+                  c.dag_size
```
adamalton, line 271: Why do we group by `c.dag_size` here? Will that not mean that if two files have the same dag size that one of them will be ignored?

I tried removing this line, and the tests still pass. Should we remove it?

adamalton:
Other than my one comment above, I think this all looks good.

packages/db/postgres/migrations/016-user-used-storage-count-pinned-only.sql
```diff
@@ -0,0 +1,61 @@
+-- Due to the parameters differing from the old function definition, if we don't drop the old one then it's not replaced.
+DROP FUNCTION IF EXISTS users_by_storage_used;
```
adamalton, line 2:
```suggestion
DROP FUNCTION IF EXISTS user_used_storage;
```

packages/db/postgres/functions.sql
flea89: `c.dag_size` is used in the `SELECT`, so it should be there. 
I'm not sure why the test were not failing for you 🤔, you should get `column "c.dag_size" must appear in the GROUP BY clause or be used in an aggregate function`.

> Will that not mean that if two files have the same dag size that one of them will be ignored?
I don't think. The dag_size is a column on `content` and we're already grouping on the primary key `cid` so that adding the `dag_size` shouldn't really changed the returned rows but rather just make it available to the aggreate `SUM`.

PR #1431, opened 2022-06-08
fix: account view priority and table size

cmunns:
- tightened up spacing on both mobile and desktop for table
- removed ctas
- added alternating bg and normalized icon size and positions for tooltips
- bumped up checkbox border for better contrast
- aligned and normalized grid system to align text along same vertical axis.
Screen Shot 2022-06-08 at 1 47 04 PM
Screen Shot 2022-06-08 at 11 46 19 AM


cmunns:
I also imported some new icons to normalize the size and stroke. The existing ones were a little wonky in a few places
Screen Shot 2022-06-09 at 9 06 55 AM

Looks like this is causing the build to fail because package-lock is being ignored and not matching.

packages/website/components/account/filesManager/fileRowItem.js
```diff
@@ -1,12 +1,14 @@
...
+import { BsFillInfoCircleFill as InfoBIcon } from 'react-icons/bs';
 
-import Tooltip from 'modules/zero/components/tooltip/tooltip';
-import CheckIcon from 'assets/icons/check';
-import InfoBIcon from 'assets/icons/infoB';
-import CopyIcon from 'assets/icons/copy';
-import PencilIcon from 'assets/icons/pencil';
-import { addTextToClipboard, truncateString } from 'lib/utils';
+import CheckIcon from '../../../assets/icons/check';
+// import InfoBIcon from '../../../assets/icons/infoB';
```
jsdevel, line 7: can you remove this comment?

packages/website/components/account/filesManager/fileRowItem.scss
```diff
@@ -67,25 +69,33 @@ $file-size: auto;
           color: $pictonBlue;
         }
         .Tooltip {
```
jsdevel, line 71: can you delete this rule?

packages/website/components/account/filesManager/filesManager.scss
```diff
@@ -36,13 +36,13 @@
 
   // Title
   > :first-child {
```
jsdevel, line 38: can you delete this?

packages/website/components/account/filesManager/filesManager.scss
```diff
@@ -36,13 +36,13 @@
...
-    margin-right: 10rem;
+    // margin-right: 10rem;
   }
   .files-manager-title {
     display: flex;
     flex-direction: row;
     &:not(.has-upload-button) {
```
jsdevel, line 44: can you delete this?

packages/website/components/account/filesManager/filesManager.scss
```diff
@@ -73,7 +73,7 @@
     position: relative;
     top: 0.125rem;
     cursor: pointer;
-    margin-right: 3.125rem;
+    // margin-right: 3.125rem;
```
jsdevel, line 76: delete

packages/website/components/account/storageManager/storageManager.scss
```diff
@@ -4,17 +4,14 @@
...
-  }
+  padding: 2rem;
 }
 
 .storage-manager-info,
 .storage-manager-space {
   display: flex;
-  padding-left: 0.6875rem;
-  padding-right: 3.9375rem;
+  // padding-left: 0.4rem;
```
jsdevel, line 13: delete

packages/website/modules/zero/components/tooltip/tooltip.js
```diff
@@ -1,5 +1,8 @@
 import clsx from 'clsx';
-import InfoAIcon from 'assets/icons/infoA';
+import { BsFillQuestionCircleFill as InfoAIcon } from 'react-icons/bs';
+
+// import InfoAIcon from 'assets/icons/infoA';
```
jsdevel, line 4: delete

jsdevel (changes requested):

olizilla (changes requested):
Nice! This is going in the right direction

- [x] fix dem conflicts 
- [x] convert the `?` icon to use the `i`. Let's pick one and use it in both places.
- [x] Improve the date format. Avoid wrapping, and remove the comma. If we have to, truncate the date and add the full timestamp as a title attr. The full date portion should always be visible.
- [x] Make the name column the most important thing. Consider moving the date to the right hand side next to size.
- [ ] if not too filddly, rework the text styles in the tooltip to make them look more coherent with the table layout. e.g. the occasionally long list of Storage Providers ID addresses could have a smaller font. As they are ID addresses they could shown in monospace font similar to CIDs, but have a play and see what feels right there.

cmunns:
The latest from main included some tabs to separate out views between uploaded via web and pinned via api. I don't see a reason to show the tabs at all if the user is only ever interacting with the gui. I also added more distinct "selected" states to the whole row and card (pictured) and removed the scale onHover to the table buttons. I want to be sure that someone can test this that has API based pins so that the tabs show up when they need to before we merge.
Screen Shot 2022-06-13 at 12 52 40 PM
Screen Shot 2022-06-13 at 3 04 56 PM

JeffLowe:
Pinging @olizilla & @jsdevel for rereview.

olizilla:
If you can right align the text in the size column when on wider screens / table layout, then you have a deal.

> The latest from main included some tabs to separate out views between uploaded via web and pinned via api. I don't see a reason to show the tabs at all if the user is only ever interacting with the gui.

this sounds significant. I've not been following the "show pinned uploads seperately" thread. We should find out who was building that and avoid undoing there changes with this PR without some checking in first.

JeffLowe:
@cmunns looks like if you can address Oli's remaining concern, we are good to go on this one:
"...right align the text in the size column when on wider screens / table layout"

PR #1432, opened 2022-06-09
fix: upgrade wrangler

### LINKED ISSUE #1399: Upgrade to wrangler 2
Wrangler 2 is now available, it simplifies our toml configuration files and also eases publishing to Cloudflare Pages, resulting in way less tooling code for deploys. 

We should get it updated here in API and start using in website.

Also unblocks future use cases only supported in Wrangler 2, like service bindings

vasco-santos:
This PR upgrades wrangler to its v2. Simplifies API worker configuration ~~and allows us to deploy to cloudflare pages~~.

Some details on our wrangler migration also based on https://developers.cloudflare.com/workers/wrangler/migration/migrating-from-wrangler-1/

```
- 😶 Ignored: "type":
      Most common features now work out of the box with wrangler, including modules, jsx,
  typescript, etc. If you need anything more, use a custom build.
    - Deprecation: "build.upload.format":
      The format is inferred automatically from the code.
    - Deprecation: "build.upload.main":
      Delete the `build.upload.main` and `build.upload.dir` fields.
      Then add the top level `main` field to your configuration file:
      ```
      main = "dist/worker.mjs"
      ```
    - Deprecation: "build.upload.dir":
      Use the top level "main" field or a command-line argument to specify the entry-point for the
  Worker.
    - Unexpected fields found in top-level field: "watch_dir"
    - "env.staging" environment configuration
      - Deprecation: "zone_id":
        This is unnecessary since we can deduce this from routes directly.
```

~~Needed before merge:~~
- [ ] ~~Go to [CF Dashboard](https://dash.cloudflare.com/fffa4b4363a7e5250af8357087263b3a/pages/view/web3-storage/settings/builds-deployments) and change branch to main + disable automatic deploys~~

Closes #1399

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -139,10 +139,7 @@ jobs:
       # --- Website deploy steps ----------------------------------------------
       - name: Deploy to Cloudflare pages
         if: ${{ steps.tag-release.outputs.releases_created }}
-        # Reset the `website-prod` branch to trigger a production build & deploy on Cloudflare Pages.
-        run: |
-          git push origin --delete website-prod
-          git push origin main:website-prod
+        run: npx wrangler pages publish --project-name web3-storage --branch main ./packages/website/out
 
       - run: npm run build -w packages/client
         if: ${{ steps.tag-release.outputs.releases_created }}
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -139,7 +139,10 @@ jobs:
       # --- Website deploy steps ----------------------------------------------
       - name: Deploy to Cloudflare pages
         if: ${{ steps.tag-release.outputs.releases_created }}
-        run: npx wrangler pages publish --project-name web3-storage --branch main ./packages/website/out
+        # Reset the `website-prod` branch to trigger a production build & deploy on Cloudflare Pages.
+        run: |
+          git push origin --delete website-prod
+          git push origin main:website-prod
 
       - run: npm run build -w packages/client
         if: ${{ steps.tag-release.outputs.releases_created }}
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -42,8 +42,7 @@ jobs:
           node-version: 16
       - uses: bahmutov/npm-install@v1
       - name: Publish api worker
-        # workaround for https://github.com/cloudflare/wrangler-action/issues/59 to use node 16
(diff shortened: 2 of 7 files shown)


olizilla:
> Go to [CF Dashboard](https://dash.cloudflare.com/fffa4b4363a7e5250af8357087263b3a/pages/view/web3-storage/settings/builds-deployments) and change branch to main + disable automatic deploys

Does this mean that we won't have a preview pages deploy for the main branch? We currently get a pages build for main which I use when I'm looking for the latest staging deploy. 

We'd really like a way to tell CF to build main and deploy it as "staging.web3.storage", but this has been difficult to do externally as we'd have to wait until the CF pages build finishes and then update the dns record

olizilla:
My minor concern here is just that if we do the change to how the website gets deployed, then we would lose the cloudflare preview deployment for the main branch.

We could just leave the `website-prod` branch trick in place, as it seems to work well.

olizilla:
An alternative option for website deploying:
- Create a `staging.web3.production` site in CF Pages. Set that up to auto build all and any branches. and set main as the "production" branch, which updates staging.web3.production
- Update `web3.production` in CF pages to disabled auto-builds, and disconnect the repo. Only use the CLI to publish the website from CI.

Pro: gets us a staging deployment
Con: prod and staging deploys are significantly different.

.github/workflows/website.yml
```diff
@@ -139,10 +139,7 @@ jobs:
       # --- Website deploy steps ----------------------------------------------
       - name: Deploy to Cloudflare pages
         if: ${{ steps.tag-release.outputs.releases_created }}
-        # Reset the `website-prod` branch to trigger a production build & deploy on Cloudflare Pages.
-        run: |
-          git push origin --delete website-prod
-          git push origin main:website-prod
+        run: npx wrangler pages publish --project-name web3-storage --branch main ./packages/website/out
```
olizilla, line 142: Unless there is something I am missing here, I think we should revert this for now, and focus on upgrading wrangler in this PR. We (I) can bikeshed on the website deployment process as a seperate issue

olizilla (changes requested):
Rad!

vasco-santos:
@olizilla sounds good, I will revert website changes and we can talk about them in an issue for now

olizilla (approved):
👍

PR #1443, opened 2022-06-10
fix: user login without mode middleware

### LINKED ISSUE #1433: Disabled writes affect logging in
Noticed that when we set maintenance mode to read only, website log ins are also affected after https://github.com/web3-storage/web3.storage/pull/1271

Login route is `POST /user/login`, and POST requests are disabled on write disabled. We need a special case for log in

vasco-santos:
Per regression from middleware changes, users cannot login when RO maintenance. This PR changes that.

The login handler is now the first handler, which makes router not to add the withMode middleware for it. Within the route handler for this specific context, we handle the maintenance part. If RW, business as usual with `upsertUser`, however if RO we now do `getUser`. If no read and no write, we fail :)

Given we don't have any log in tests, I couldn't easily get a test for this, which is really unfortunate. We should really invest some time on having tests for login... 

Closes #1433


packages/api/src/user.js
```diff
@@ -42,7 +46,16 @@ async function loginOrRegister (request, env) {
     ? parseGitHub(data.data, metadata)
     : parseMagic(metadata)
 
-  const user = await env.db.upsertUser(parsed)
+  let user
+  // check if maintenance mode
+  if (env.mode === NO_READ_OR_WRITE) {
```
jsdevel, line 51: shouldn't this be `env.MODE`?

vasco-santos: good catch, thanks for fixing

PR #1444, opened 2022-06-10
feat: Docs polish

svvimming:
The following list of changes and polish has been made to the docs pages:

#### Desktop

- [x] added transitions between ToC highlighted items
  - highlighted title transitions in & chevron fades in
- [x] ordered & unordered lists bullet points indentation added
  - list item padding reduced
  - nested list styling added
- [x] hover effects added to tab component
  - gradients under tabs slide from left to right on hover
  - gradients transition on hover to active transition
  - tab content transitions from opacity 0 to 1
- [x] link hover effects added to match marketing site
- [x] clipboard copy buttons added to code blocks
- [x] additional styling added to inline code text/messages to visually indicate they are code
  - added padding, margin, border, border-radius
  - added inset box-shadow
- [x] reduced margins on `p` tags in all cases except when followed by another `p` tag
- [x] buttons vertically centred on ‘was this information helpful’ section
- [x] hover effects added to previous and next buttons at the end of articles, matches marketing site
- [x] transitions added to sidebar headings similar to table of contents
  - note: because these headings trigger a reload of the page, only transitions from non-active to active headings can be added, not transitions from active to non-active states
- [x] limited bash syntax highlighting added to code blocks with highlight js
- [x] native outline on accordions removed on focus and replaced with focus-visible psuedo class
- [x] height transitions added on accordions in HTTP API page
  - @orvn I'm not sure how you feel about this effect. Because these components are rendered from the swagger-ui library there's not much I can do to add height transitions, especially since I don't have access to the javascript and can only achieve css based solutions. Right now I'm using a keyframes animation to animate the max-height and at the end reset max-height to auto. There's no way to animate on close though because the swagger js actually removes the content from the DOM.
- [x] HTTP API page accordion chevron rotation transition
  - similar story for adding a transition to rotate the accordion heading chevrons; swagger actually replaces the arrow with an entirely different arrow facing the opposite way rather than transforming with css, so this transition isn't possible.
- [x] margins reduced on HTTP API page headings
- [x] indent added on accordion chevrons in the image gallery page

#### Mobile

- [x] Main container is not 100% viewport width (on iOS)
- [x] Navigation background image open state should be stretched to avoid contrast issues
- [x] Tables and toggles (on HTTP API) are compressed
- [x] Algolia no search results state suggestions look a little condensed (_"Try searching for"_ section)


orvn:
> **Height transitions added on accordions in HTTP API page** 
> @orvn I'm not sure how you feel about this effect. Because these components are rendered from the swagger-ui library there's not much I can do to add height transitions, especially since I don't have access to the javascript and can only achieve css based solutions. Right now I'm using a keyframes animation to animate the max-height and at the end reset max-height to auto. There's no way to animate on close though because the swagger js actually removes the content from the DOM.

I think this is okay @svvimming, unless @joannaong knows another approach from having worked with `swagger-ui`? The open effect with the `max-height` trick is fine, I actually use that as my default toggle effect personally. But it's too bad the element gets removed from the DOM before it's able to animate the height back.

> **HTTP API page accordion chevron rotation transition**
> _similar story for adding a transition to rotate the accordion heading chevrons; swagger actually replaces the arrow with an entirely different arrow facing the opposite way rather than transforming with css, so this transition isn't possible._

@joannaong, any ideas for this? (e.g. something that would keep the same arrow icon so that we can transform and transition it?)

joannaong:
> > **Height transitions added on accordions in HTTP API page**
> > @orvn I'm not sure how you feel about this effect. Because these components are rendered from the swagger-ui library there's not much I can do to add height transitions, especially since I don't have access to the javascript and can only achieve css based solutions. Right now I'm using a keyframes animation to animate the max-height and at the end reset max-height to auto. There's no way to animate on close though because the swagger js actually removes the content from the DOM.
> 
> I think this is okay @svvimming, unless @joannaong knows another approach from having worked with `swagger-ui`? The open effect with the `max-height` trick is fine, I actually use that as my default toggle effect personally. But it's too bad the element gets removed from the DOM before it's able to animate the height back.
> 

I don't think there's a quick fix for this in swagger

> > **HTTP API page accordion chevron rotation transition**
> > _similar story for adding a transition to rotate the accordion heading chevrons; swagger actually replaces the arrow with an entirely different arrow facing the opposite way rather than transforming with css, so this transition isn't possible._
> 
> @joannaong, any ideas for this? (e.g. something that would keep the same arrow icon so that we can transform and transition it?)

I added pseudo elements for accordion and removed existing one so we can transition. It only works with main accordions tho and not the `schemas` since the button element gets removed for that after clicking

558ab3db2ac7477e432c6527b05272fe

orvn:
@joannaong could you take a look at these style conflicts against `main`?

orvn (approved):
Added some more polish to a variety of elements:
- Link styles
- List styles
- Callout icon alignment on Firefox and Safari
- Codeblock copy button feedback
- Also fixed an error with prop not being passed to image component in `how-tos/list`

PR #1455, opened 2022-06-13
feat!: switch to Elastic IPFS

alanshaw:
This PR switches to uploading to the Elastic Provider S3 bucket by default and uploading to IPFS Cluster asynchronously after the response is sent to the client.

We pin the root CID to Cluster and if the upload to Cluster fails, we hope that Cluster can fetch it from elastic provider later.

A dockerised minio is added to dev and test which stands in for a real life S3 in these environments.

BREAKING CHANGE: uploaded files are no longer _instantly_ available via gateways or on the DHT. They may take a few seconds to become indexed by Elastic Provider.


packages/api/src/utils/logs.js
```diff
@@ -212,16 +212,19 @@ export class Logging {
       if (this.opts.sentry && message.status >= 500 && !skipForSentry.some((cls) => message instanceof cls)) {
         this.opts.sentry.captureException(message)
       }
+      if (this.opts?.debug) {
+        console[level](`[${dt}]`, message)
+      }
```
alanshaw, line 217: Logs out the full error object when in debug mode instead of just the message.

packages/api/test/car.spec.js
```diff
@@ -44,10 +55,14 @@ describe('POST /car', () => {
     assert(cid, 'Server response payload has `cid` property')
     assert.strictEqual(cid, expectedCid, 'Server responded with expected CID')
 
-    const statusRes = await fetch(new URL(`status/${cid}`, endpoint))
-    const status = await statusRes.json()
-    const pinInfo = status.pins.find(pin => PIN_OK_STATUS.includes(pin.status))
-    assert(pinInfo, `status is one of ${PIN_OK_STATUS}`)
+    const pinInfo = await retry(async () => {
```
alanshaw, line 58: This is now done in the `waitUntil` so we retry until we get the result we want.

packages/api/src/car.js
```diff
@@ -116,52 +116,56 @@ export async function handleCarUpload (request, env, ctx, car, uploadType = 'Car
...
 
   /** @type {(() => Promise<any>)[]} */
-  const tasks = []
+  const tasks = [async () => {
+    let pins = []
+    try {
+      pins = await addToCluster(car, env)
+    } catch (err) {
+      env.log.warn(new Error('failed to upload to cluster', { cause: err }))
+      pins = await pinToCluster(rootCid.toString(), env)
```
olizilla, line 147: 💡 we could pin first then add for maximum chance of actually getting the blocks on to cluster at some point... in the edge case where addToCluster runs out of function invokation time.

packages/api/package.json
```diff
@@ -9,22 +9,25 @@
...
+    "start": "npm-run-all -p minio:start \"minio:mkbucket -- dotstorage-dev-0\" cluster:start pg:start -s miniflare",
+    "start:preview": "npm-run-all -s minio:start \"minio:mkbucket -- dotstorage-dev-0\" -p cluster:start pg:start -s -p lt  wrangler-dev",
+    "stop": "npm-run-all -p minio:stop cluster:stop pg:stop",
     "wrangler-dev": "wrangler dev --env $(whoami)",
     "publish": "wrangler publish --env $(whoami)",
     "build": "node scripts/cli.js build",
-    "test": "npm-run-all -p -r mock:backup test:worker -s test:size",
-    "test:worker": "mocha --require test/hooks.js --experimental-vm-modules --timeout 5000",
+    "test": "run-s test:worker test:size",
+    "test:worker": "mocha --require test/hooks.js --experimental-vm-modules --timeout 5000 --bail",
```
olizilla, line 19: This looks like it might be left over from testing out tests.
```suggestion
    "test:worker": "mocha --require test/hooks.js --experimental-vm-modules --timeout 5000",
```

packages/api/package.json
```diff
@@ -42,9 +45,9 @@
     "mocha": "^9.2.0",
     "npm-run-all": "^4.1.5",
     "p-defer": "^4.0.0",
+    "p-retry": "^5.1.1",
     "process": "^0.11.10",
     "sade": "^1.8.1",
-    "smoke": "^3.1.1",
```
olizilla, line 47: aside: how do we test edge cases and error conditions without smoke or similar?

packages/api/src/car.js
```diff
@@ -116,52 +116,56 @@ export async function handleCarUpload (request, env, ctx, car, uploadType = 'Car
...
+      await waitAndUpdateOkPins(normalizedCid, env.cluster, env.db)
+    }
+  }]
 
   if (ctx.waitUntil) {
     tasks.forEach(t => ctx.waitUntil(t()))
   }
 
-  return new JSONResponse({ cid })
+  return new JSONResponse({ cid: rootCid.toString() })
```
olizilla, line 168: minor: We don't seem to use the CID object form in this method. We could just have `carStat(car)` return a property called `cid` as the string form, so we don't have to call `rootCid.toString()` a bunch o times.

olizilla (approved):
🚀

olizilla:
The minio command is a little to quick to give up on the first run... the the docker image is still being fetched from quay.io when the retry fails.

```console
$ npm start
...
[api] Creating network "minio_default" with the default driver
[api] Creating volume "minio_minio0" with default driver
[api] Pulling minio0 (quay.io/minio/minio:)...
[api] latest: Pulling from minio/minio
[web] info  - Using external babel configuration from /Users/oli/Code/web3-storage/web3.storage/packages/website/.babelrc
[api] file:///Users/oli/Code/web3-storage/web3.storage/packages/tools/scripts/minio.js:105
[api]       throw new Error(`Minio API not reachable on port: ${MINIO_API_PORT}`)
[api]             ^
[api] 
[api] Error: Minio API not reachable on port: 9000
[api]     at retry.retries (file:///Users/oli/Code/web3-storage/web3.storage/packages/tools/scripts/minio.js:105:13)
[api]     at processTicksAndRejections (node:internal/process/task_queues:96:5)
[api]     at async RetryOperation._fn (file:///Users/oli/Code/web3-storage/web3.storage/packages/tools/node_modules/p-retry/index.js:54:13) {
[api]   attemptNumber: 4,
[api]   retriesLeft: 0
[api] }
```

olizilla (changes requested):

packages/api/test/car.spec.js
vasco-santos: It would be great to consider changing to Ava here and rely on Miniflare lib to wait for this kind of things

packages/tools/scripts/minio.js
```diff
@@ -0,0 +1,141 @@
...
+    .describe('Remove a bucket, automatically removing all contents')
+    .action(bucketRemoveCmd)
+}
+
+/**
+ * @param {{ project?: string }} opts
+ */
+async function serverStartCmd ({ project = 'minio' }) {
+  if (await isPortReachable(MINIO_API_PORT)) {
+    return console.log('Skipped starting minio. Port 9000 is already in use, so assuming minio is already running.')
```
vasco-santos, line 52:
```suggestion
    return console.log(`Skipped starting minio. Port ${MINIO_API_PORT} is already in use, so assuming minio is already running.`)
```

packages/api/test/car.spec.js
alanshaw: I agree, but seperate from this PR!

packages/api/test/scripts/constants.js
```diff
@@ -1,3 +1,3 @@
 export const endpoint = 'http://localhost:8787'
-export const clusterApi = 'http://localhost:9094'
+export const clusterApi = 'http://127.0.0.1:9094'
```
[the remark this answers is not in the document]
alanshaw: Seems that it's because I'm using Node.js 18...which resolves localhost to `::1` on MacOS (no longer reorders for consistent resolution across platform).

packages/api/src/car.js
alanshaw: I looked and we use `rootCid.code` immediately after calling `carStat` so I've left `carStat` alone but just assigned the stringified version to a local var and used that everywhere 😉.

packages/tools/docker/minio/docker-compose.yml
```diff
@@ -0,0 +1,18 @@
+version: '3.7'
```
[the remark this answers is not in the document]
alanshaw: Yes, needs a bit more refactor as there is another one for the DB in `packages/db/postgres/docker`. I'll open an issue.

packages/api/src/car.js
olizilla: aye, that's better. Note that `uploadToBucket` just takes the cid object and calls toString on it again tho.

packages/api/src/car.js
```diff
@@ -209,18 +224,16 @@ async function addToCluster (car, env) {
...
  * @param {string} userId
- * @param {import('./env').Env} env
  * @param {('Unknown' | 'Partial' | 'Complete')} structure The known structural completeness of a given DAG.
  */
-async function backup (blob, rootCid, userId, env, structure = 'Unknown') {
-  if (!env.s3Client) {
-    return undefined
-  }
-
+async function uploadToBucket (s3, bucketName, blob, rootCid, userId, structure = 'Unknown') {
```
olizilla, line 236: minor: might be clearer here to change the function signature to take the string form of the `sourceCid`, as we're uploading a user generated CAR and we want to key it on the string form of the rootCid in the format provided in the CAR file.
```suggestion
async function uploadToBucket (s3, bucketName, blob, sourceCid, userId, structure = 'Unknown') {
```

olizilla (approved):
Minor note about using sourceCid everywhere, but not blocking, just an observation. Is GOOD!

PR #1456, opened 2022-06-13
chore: get back integrity and resolved keys for deps

flea89:
#1196  removed all the integrity hashes and resolved URLs from the package lock.

This PR is to get them back.

# Notes
I wasn't actually quite sure about the best way to do this.
Here's what I've done:
1. checked-out the `package-lock.json` from ~[dd9ce3b](https://github.com/web3-storage/web3.storage/commit/dd9ce3b89ccfb0b8b5a663e1673937c992720617). (`git checkout dd9ce3b~1 package-lock.json`)
2. ran `npm -i`

Unfortunately being the changes so big it's hard to spot any problems with this.

But hopefully #1453  might help with that?


alanshaw (approved):
I checked out your branch and ran `npm install` and got no changes to `package-lock.json` so I assume this is gtg.

PR #1459, opened 2022-06-13
test: Retrieve user pinned storage data

svvimming:
Retrieve user pinned storage data


svvimming:
This draft is no longer needed as the problem it was probing (storage bar not showing pinned file storage usage) has been [resolved here](https://github.com/web3-storage/web3.storage/pull/1333)

Issue #976, opened 2022-02-14
Improve error handler for Error children


alexandrastoica:
Follow up from https://github.com/web3-storage/web3.storage/pull/937

Currently `error-handler.js` uses `instanceof` to determine the type or error class (`HTTPError`). 

With the addition of `PinningServiceApiError` class (see `errors.js`), a bug where all `Error` children seem to be instances of the other `Error` children surfaced. As a temporary fix, we used duck typing to correctly identify the PSA error, however we should fix the general `instanceof` bug.

adamalton:
There's [a discussion on a change in  937](https://github.com/web3-storage/web3.storage/pull/937/files#r802774788) in which it's said that `instanceof PinningServiceApiError` returns true for instances of `Error`. But from a quick test of this in the JS console, that doesn't seem to be how things behave:

<img width="271" alt="Screenshot 2022-04-20 at 09 38 17" src="https://user-images.githubusercontent.com/270255/164188165-d5c276f5-fc46-4b78-aeb9-78f45fd4fbc2.png">

(The result is the same in the Node console.)

Am I misunderstanding the problem? Or is there a chance that there's something slightly different that was causing this problem?

flea89:
You understood it right, and indeed, that's what was happening in my tests. 
Both in Chrome console and Node console this is working as expected but not when the code is run in the Cloudflare Worker context.
I should have added more context to my comment, as you did 😛.

I did not spend a lot of time investigating the source of the issue, but something funky was happening.

Since then we changed how we define the worker (service worker API to es6 module and how we bundle it) so might be worth looking at this again.

mbommerez:
Discussed with @alanshaw . We need to get this fixed in next sprint.

adamalton:
Fixed by #1463.

PR #1462, opened 2022-06-13
chore: add API endpoint to test instanceof behaviour on Cloudflare

adamalton:
This is for testing whether the strange behaviour that's described in #976 is still occurring on Cloudflare.


francois-potato (approved):

alanshaw (approved):

adamalton:
And the result is…
Screenshot 2022-06-14 at 15 03 30

No bug!

PR #1463, opened 2022-06-13
chore: use `instanceof` for error type inference

adamalton:
This is a fix for #976, which, if #1462 shows that the bug no longer exists, we can merge.


francois-potato (approved):

vasco-santos (commented):
Oh nice interesting! Can we remove the temporary route?

adamalton:
@vasco-santos ah yes, good point about removing the temporary route. I'll do that.

francois-potato (approved):

PR #1465, opened 2022-06-13
fix: update getPg to automatically connect to the db

flea89:
I suspect the current jobs that use a direct connection to a RO/RW instances of the DB aren't currently working.

## Why?
If we take `Populate Missing Content DAG Sizes` as an example you can see that only the initial log (before any DB operations is logged). [Here](https://github.com/web3-storage/web3.storage/runs/6852404133?check_suite_focus=true)'s  an example.

As you can see only 
```2022-06-12T20:44:29.213Z dagcargo:updateDagSizes 🎯 Updating DAG sizes for content inserted after 2022-05-12T20:44:29.212Z```.
No other logs are present (ie `✅ Done`)

I suspect that's because we're not explicitly connecting something fails, not sure why that silently though, I'd expect `query` to raise ( I haven't actually dig deeper in the pg client code to check the actual problem).

## Fix
Refactor `getPg` to try to connect automatically to the DB.

Here's a [Draft PR](https://github.com/web3-storage/web3.storage/compare/chore/test-cron-connect-dag-size) where I showcase this is actually fixing the issue, see action [output](https://github.com/web3-storage/web3.storage/runs/6866149699?check_suite_focus=true).

### DIFF
--- a/packages/cron/src/bin/dagcargo-sizes.js
+++ b/packages/cron/src/bin/dagcargo-sizes.js
@@ -11,8 +11,8 @@ import { getPg } from '../lib/utils.js'
 const xDaysAgo = (days) => new Date().setDate(new Date().getDate() - days)
 
 async function main () {
-  const rwPg = getPg(process.env, 'rw')
-  const roPg = getPg(process.env, 'ro')
+  const rwPg = await getPg(process.env, 'rw')
+  const roPg = await getPg(process.env, 'ro')
 
   try {
     const after = new Date(process.env.AFTER || xDaysAgo(7))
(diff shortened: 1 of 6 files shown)


adamalton (approved):
LGTM.

I don't understand why the jobs fail with no output when there's no connection, but I saw exactly the same behaviour with the storage quota cron.

alanshaw:
In NFT.Storage we manage connection/disconnection in the bin script:

https://github.com/nftstorage/nft.storage/blob/e2eb81e2aec6bc3cbe915a195e9a4f5b63d25963/packages/cron/src/bin/dagcargo-sizes.js#L14-L20

I think at the time it just made sense to me to manage this in the same function.

Note that for pool you don't call connect, since it manages that for you (although you do have to call `end` anyway to close any connections it's holding open). Arguably we should just switch to using pool everywhere.

Issue #61, opened 2021-07-12
Listing user uploads could be faster


andrew:
The endpoint listing user uploads seems to be a little slow, could be because of the pagination options passed causing missed indexes perhaps?

Url: https://api-staging.web3.storage/user/uploads?before=2021-07-12T13%3A26%3A44.976Z&size=25
Number of files listed in response: 1
Avg time to load in firefox: ~1 second

Related to performance research in #94

vasco-santos:
We moved to Postgres and this should not be the case anymore. Closing this, but feel free to reopen if this is still a problem

andrew:
It's actually much worse for my account now:
```
$ time curl -H 'Accept: application/json' -H "Authorization: Bearer TOKEN" https://api.web3.storage/user/uploads
$ [jsonresultshere]
real	0m7.084s
user	0m0.022s
sys	0m0.010s
```

orvn:
@dchoi27 I think this might be mislabeled/an issue outside of `stack/website`? 

(On the frontend, #1341 should help with displaying the list more quickly, but that's not directly related to this issue)

drewdelano:
I think this a missing index issue

PR #1466, opened 2022-06-13
fix: naive take on adding an index

drewdelano:
user_id on uploads should probably be indexed...

completes #61

### DIFF
--- a/packages/db/postgres/migrations/016-add-uploads-user-id-index.sql
+++ b/packages/db/postgres/migrations/016-add-uploads-user-id-index.sql
@@ -1,0 +1,1 @@
+CREATE INDEX IF NOT EXISTS upload_user_id_deleted_at_idx ON upload (user_id) WHERE deleted_at IS NULL;
--- a/packages/db/postgres/migrations/016-add-uploads-user-id-index.sql
+++ b/packages/db/postgres/migrations/016-add-uploads-user-id-index.sql
@@ -1,1 +1,0 @@
-CREATE INDEX IF NOT EXISTS upload_user_id_deleted_at_idx ON upload (user_id) WHERE deleted_at IS NULL;
--- a/packages/db/postgres/migrations/018-add-uploads-user-id-index.sql
+++ b/packages/db/postgres/migrations/018-add-uploads-user-id-index.sql
@@ -1,0 +1,1 @@
+CREATE INDEX CONCURRENTLY IF NOT EXISTS upload_user_id_deleted_at_idx ON upload (user_id) WHERE deleted_at IS NULL;
--- a/packages/db/postgres/tables.sql
+++ b/packages/db/postgres/tables.sql
@@ -267,6 +267,7 @@ CREATE TABLE IF NOT EXISTS upload
 );
 
 CREATE INDEX IF NOT EXISTS upload_auth_key_id_idx ON upload (auth_key_id);
+CREATE INDEX IF NOT EXISTS upload_user_id_deleted_at_idx ON upload (user_id) WHERE deleted_at IS NULL;
 CREATE INDEX IF NOT EXISTS upload_content_cid_idx ON upload (content_cid);
 CREATE INDEX IF NOT EXISTS upload_updated_at_idx ON upload (updated_at);
 CREATE INDEX IF NOT EXISTS upload_user_id_idx ON upload (user_id);


vasco-santos (commented):
Hey @drewdelano 
Can we know more about the motivation for this index? We currently have `UNIQUE (user_id, source_cid)` for upload table, which already provides an index for `user_id` behind the scenes. Here you filter out non deleted uploads, but I would say we don't need it unless we are experiencing slower queries with the already existing index.

drewdelano:
@vasco-santos This is the issue I'm looking at:
https://github.com/web3-storage/web3.storage/issues/61
https://github.com/web3-storage/web3.storage/blob/112b1788b8f226ae8d984132dedea90019441ea5/packages/db/index.js#L437

We seem to be seeing a steady increase over time which in my mind points to indexing.

packages/db/postgres/migrations/016-add-uploads-user-id-index.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX IF NOT EXISTS upload_user_id_deleted_at_idx ON upload (user_id) WHERE deleted_at IS NULL;
```
vasco-santos, line 1:
```suggestion
CREATE INDEX CONCURRENTLY IF NOT EXISTS upload_user_id_deleted_at_idx ON upload (user_id) WHERE deleted_at IS NULL;
```

This needs to concurrently so that the operation does not lock the database

vasco-santos (changes requested):

vasco-santos:
> That said, it might be worth running that query against staging to see how long it takes to return. It's entirely possible I'm wrong here.

We can definitely try it. Maybe the deleted is the key difference for users with a large number of uploads

ribasushi:
@vasco-santos @drewdelano why not test all this ( EXPLAIN ANALYZE in hand ) on one of your replicas? The indices are all replicated and identical.

vasco-santos:
> @vasco-santos @drewdelano why not test all this ( EXPLAIN ANALYZE in hand ) on one of your replicas? The indices are all replicated and identical.

Yes, good call @ribasushi . I was thinking about trying that before we move forward here. @drewdelano does not have access so must be me to try it

vasco-santos:
It looks like this helps a lot! @drewdelano 

```
EXPLAIN ANALYZE 
SELECT * FROM upload
WHERE user_id = XXXXXXXXXXXXXXXXX
AND deleted_at IS NULL;
```

Before Index

```
 Bitmap Heap Scan on upload  (cost=6443.97..279801.88 rows=203188 width=279) (actual time=5260.205..68250.982 rows=187339 loops=1)
   Recheck Cond: (user_id = 'XXXXXXXXXXXXXXXXX'::bigint)
   Filter: (deleted_at IS NULL)
   Heap Blocks: exact=131036
   ->  Bitmap Index Scan on upload_user_id_source_cid_key  (cost=0.00..6433.81 rows=203799 width=0) (actual time=5228.591..5228.592 rows=187391 loops=1)
         Index Cond: (user_id = 'XXXXXXXXXXXXXXXXX'::bigint)
 Planning Time: 0.118 ms
 JIT:
   Functions: 4
   Options: Inlining false, Optimization false, Expressions true, Deforming true
   Timing: Generation 0.488 ms, Inlining 0.000 ms, Optimization 0.519 ms, Emission 3.815 ms, Total 4.822 ms
 Execution Time: 68301.216 ms
(12 rows)
```

After Index

```
 Index Scan using upload_user_id_deleted_at_idx on upload  (cost=0.09..216730.20 rows=202877 width=279) (actual time=0.071..525.612 rows=187339 loops=1)
   Index Cond: (user_id = 'XXXXXXXXXXXXXXXXX'::bigint)
 Planning Time: 3.819 ms
 JIT:
   Functions: 2
   Options: Inlining false, Optimization false, Expressions true, Deforming true
   Timing: Generation 0.194 ms, Inlining 0.000 ms, Optimization 0.000 ms, Emission 0.000 ms, Total 0.194 ms
 Execution Time: 564.808 ms
(8 rows)
```

packages/db/postgres/migrations/016-add-uploads-user-id-index.sql
vasco-santos, line 1: Also name number will need to be updated

vasco-santos (approved):
LGTM

PR #1484, opened 2022-06-15
fix: drop uploads user index

vasco-santos:
With index added in https://github.com/web3-storage/web3.storage/pull/1466 this should not be needed, as it includes it and also has optimization for not deleted


adamalton (approved):
Makes sense, and looks good to me.

vasco-santos:
Migration applied ✅

PR #1486, opened 2022-06-15
fix: cron test should run on db change

vasco-santos:
We noticed DB changes did not trigger cron tests and cron uses DB, so it should


flea89 (approved):
LGTM

PR #1487, opened 2022-06-15
fix: Navigation accordion styles

svvimming:
Newer accordion styles added to the accordion block component were overriding previous nav styles. This PR fixes that.


cmunns:
It is possible to get before and after screenshots? I don't see any difference locally with this checked out compared to prod?

PR #1494, opened 2022-06-16
fix: pins missing

drewdelano:
You'll need to have a token setup in the API tokens screen + have used the API to pin a file (status has transitioned to "Pinned") in order to see the changes (see https://web3.storage/docs/how-tos/pinning-services-api/)

Before:
image

After:
image
image

### DIFF
--- a/packages/website/components/account/filesManager/filesManager.js
+++ b/packages/website/components/account/filesManager/filesManager.js
@@ -62,7 +62,7 @@ const FilesManager = ({ className, content, onFileUpload }) => {
     storageData: { refetch },
     info,
   } = useUser();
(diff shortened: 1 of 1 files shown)


drewdelano:
This is a little weird because the `listPsaPinRequests` API is a little weird:
https://github.com/web3-storage/web3.storage/blob/8187bb5891ccf73c35289fd9f265ea110cbd5b9a/packages/db/index.js#L839

Might make sense to change the signature of the API to be user specific instead of key specific?  @jsdevel thoughts?

jsdevel (approved):

svvimming (approved):
looks good 👍

PR #1532, opened 2022-06-20
feat: abortable API method calls

### LINKED ISSUE #1516: Option to abort client.put request
Hi,
I am currently using the [web3.storage js package
](https://www.npmjs.com/package/web3.storage)

Is it possible to abort client.put request?

fetch API currently supports abort [signal](https://developer.mozilla.org/en-US/docs/Web/API/fetch#:~:text=Navigator.sendBeacon()%20API.-,signal,-An%20AbortSignal%20object)
Is it possible to expose the signal option in the package.

Use case:
Network request is expensive and sometimes the user wants to stop uploading the file.

Thanks

alanshaw:
resolves #1516 

TODO: add tests


fawazahmed0:
You might also need to update the [documentation](https://web3.storage/docs/reference/js-client-library/#store-files)

PR #1559, opened 2022-06-23
fix: no page reload when pointing to relative path in docs

joannaong:
Resolves an issue where, after docs links were tested and modified to use absolute paths, they all triggered a page reload. Because all links paths start with `/docs/foo`, they are treated as regular links that the client handles, rather than Next's built-in virtual routing.

### DIFF
--- a/packages/website/modules/docs-theme/index.js
+++ b/packages/website/modules/docs-theme/index.js
@@ -30,8 +30,9 @@ export default function Docs(props) {
     localLinks?.forEach(link => {
       link.addEventListener('click', e => {
         e.preventDefault();
+        e.stopPropagation();
         // @ts-ignore
-        router.push(e.target.href);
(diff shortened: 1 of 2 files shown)


orvn (changes requested):
@joannaong looks good overall, but I'm seeing a bug with one of the links that seems otherwise valid.

### Steps to reproduce

- In the build preview, see [this page](https://4530674b.web3-storage.pages.dev/docs/reference/js-client-library/#retrieve-files) (`docs/reference/js-client-library`)
- The 3rd link in the content `content identifier (CID)` is italicized
- Clicking this link throws a type error

Looks like the issue might have to do with nesting, since there's markdown within the link content (`_` rendering as `<em>`).

orvn (approved):
Issue resolved!

PR #1564, opened 2022-06-24
fix: build-cargo-connection-string

flea89:
We already have CARGO secrets in Github, let's build the connection string out of those instead of having a new one.

### DIFF
--- a/packages/cron/src/lib/utils.js
+++ b/packages/cron/src/lib/utils.js
@@ -93,7 +93,7 @@ function getPgConnString (env, mode = 'rw') {
 }
 
 /**
- * Builds a post connections string
+ * Builds a pg connections string
  *
  * @param {object} dbOptions
  * @param {string} dbOptions.host
--- a/packages/cron/src/lib/utils.js
+++ b/packages/cron/src/lib/utils.js
@@ -110,7 +110,7 @@ function buildPgConnectionString ({
   password,
   port = 5432
 }) {
-  return `postgres//${user}:${password}@${host}:${port}/${database}`
+  return `postgres://${user}:${password}@${host}:${port}/${database}`
 }
 
 /**
--- a/packages/cron/src/lib/utils.js
+++ b/packages/cron/src/lib/utils.js
@@ -143,8 +143,6 @@ export function getCargoPgPool (env) {
     host: env.DAG_CARGO_HOST
   })
 
-  console.log(`Using ${connectionString}`)
-
   return new pg.Pool({
     connectionString,
     max: MAX_CONCURRENT_QUERIES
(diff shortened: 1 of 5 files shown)


packages/cron/src/lib/utils.js
```diff
@@ -102,13 +123,34 @@ function getPgConnString (env, mode = 'rw') {
  * @param {Record<string, string|undefined>} env
  */
 export function getCargoPgPool (env) {
-  const connection = env.CARGO_PG_CONNECTION
-  if (!connection) {
+  let connectionString = env.CARGO_PG_CONNECTION
+  // if a connection string isn't specified try to build the connection string from other variables.
```
flea89, line 127: I've left this, just to allow for easier setup locally? Less variables to set :P
But please disagree, and we can remove this.

packages/cron/src/lib/utils.js
```diff
@@ -92,6 +92,27 @@ function getPgConnString (env, mode = 'rw') {
...
+ * @returns
+ */
+function buildPgConnectionString ({
+  host,
+  database,
+  user,
+  password,
+  port = 5432
+}) {
+  return `postgres//${user}:${password}@${host}:${port}/${database}`
```
olizilla, line 113:
```suggestion
  return `postgres://${user}:${password}@${host}:${port}/${database}`
```

packages/cron/src/lib/utils.js
olizilla: let's use the same vars in dev and prod!

packages/cron/src/lib/utils.js
```diff
@@ -102,13 +123,30 @@ function getPgConnString (env, mode = 'rw') {
...
+  })
+
+  const connectionString = buildPgConnectionString({
+    user: env.DAG_CARGO_USER,
+    database: env.DAG_CARGO_DATABASE,
+    password: env.DAG_CARGO_PASSWORD,
+    host: env.DAG_CARGO_HOST
+  })
+
+  console.log(`Using ${connectionString}`)
```
adamalton, line 146: Is there a danger of this logging the DB password somewhere? Or is `console.log` only ever hooked up to something in a dev/test environment?

packages/cron/test/cargo.spec.js
```diff
@@ -8,7 +8,10 @@ import { getCargoPgPool, getPg } from '../src/lib/utils.js'
 const env = {
   ...process.env,
   ENV: 'dev',
-  CARGO_PG_CONNECTION: 'postgres://postgres:postgres@127.0.0.1:5432/postgres?currentSchema=cargo'
+  DAG_CARGO_HOST: '127.0.0.1',
```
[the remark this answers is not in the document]
adamalton: I've added this explanation into a comment in the code.

adamalton (approved):
All looks good now.

PR #1573, opened 2022-06-27
fix: optimise pin content cid query

flea89:
Create an index on `pin(content_cid, status)` to speed up `user_used_storage` joining and filtering.

I'll create a separate issue to audit the use of `CREATE INDEX IF NOT EXISTS pin_status_idx ON pin (status);` and to verify if it can be dropped in favor of this new one.

[Storage Limit Email Notifications Storage Limit Email Notifications #43](https://github.com/web3-storage/web3.storage/runs/7074203706?check_suite_focus=true) ran on staging in 11 minutes. ( staging amount of data is now comparable to prod).

🛑  The index has already been created in STAGING to test the performance against a large amount of data.


vasco-santos:
Applying index in prod

PR #1576, opened 2022-06-27
fix: pg connection string append ssl-true

flea89:
Fix [issue](https://github.com/web3-storage/web3.storage/runs/7075750428?check_suite_focus=true) with connection to cargo from cron.


alanshaw (approved):

flea89:
@alanshaw updated to not use ssl if ENV=`dev`

PR #1583, opened 2022-06-28
fix: Fixed all broken links in docs/how-tos/store.md page

anomic30:


johnathonroach:
Good work! Fixed in another PR as well. Closing.

PR #1584, opened 2022-06-28
fix: only log user IDs, not names or email addresses

adamalton:
Although these logs are only accessible to project members, not the public, it still seems a bit risqué to be logging user PII.


flea89 (changes requested):
Shouldn't this be ~~user.id~~ `user._id` across the board?

adamalton:
@flea89 good point, it needs to be `user._id` in `email/service.js` (but still `user.id` in `jobs/storage.js`). Fixed.

joshghent (approved):

flea89 (approved):

PR #1593, opened 2022-06-29
fix: sentry sourcemaps on errors

olizilla:
the errors are now showing up in sentry with `abs_path: "orker.js"` which looks like we are trimming a w from `worker.js`.
that also suggests the file is now renamed worker.js by wrangler v2 on upload, so we rename ours in this PR.

- `worker.mjs` -> `worker.js` to match what we see in sentry error.
- stop stripping first char on paths in stack traces in sentry client config.
- set `rewriteFrames.root: '/'` to get toucan to prepend `/` to make it `/worker.js`

see: https://github.com/web3-storage/web3.storage/pull/1033

License: (Apache-2.0 AND MIT)

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "type": "module",
   "license": "(Apache-2.0 OR MIT)",
-  "main": "./dist/worker.mjs",
+  "main": "./dist/worker.js",
   "scripts": {
     "lt": "npm-run-all -p lt:*",
     "lt:cluster": "npx localtunnel --port 9094 --subdomain \"$(whoami)-cluster-api-web3-storage\"",
--- a/packages/api/scripts/cli.js
+++ b/packages/api/scripts/cli.js
@@ -29,7 +29,7 @@ prog
       entryPoints: [path.join(__dirname, '..', 'src', 'index.js')],
       bundle: true,
       format: 'esm',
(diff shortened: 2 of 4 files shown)


olizilla:
Fixed!
Screenshot 2022-06-29 at 16 08 57

I am at least getting faster at this.

vasco-santos (approved):

olizilla:
This isn't fully fixed yet. The sourcemaps are found on sentry, but the stacktrace still isn't useble... it doesn't point to where the error came from

hugomrdias:
workers are outputting wrong line numbers...

philipatkinson:
> This isn't fully fixed yet. The sourcemaps are found on sentry, but the stacktrace still isn't useble... it doesn't point to where the error came from

I seem to have the same issue. Did you ever identify a reason why the CF stack traces for errors seem to be incorrect?

hugomrdias:
Finally its fixed https://github.com/nftstorage/nft.storage/pull/2094 
add `no_bundle = true` to wrangler.toml

olizilla:
We need 
https://github.com/web3-storage/web3.storage/pull/1593/files#diff-af47dbf0e21ab2fbc5e41184392fcb87c93fb42a80ca6050c64db069fd447853L97

to fix our stack traces

olizilla:
This is painful to rebase. Have recreated in #1838

PR #1594, opened 2022-06-29
fix: make sure null claimed sizes are updated

flea89:
## Context
In SQL comparisons with null are a bit funky,
Given this query

```
SELECT d.cid_v1 as cid, d.size_actual
  FROM cargo.dag_sources ds
  JOIN cargo.dags d using ( cid_v1 )
  JOIN cargo.sources s USING ( srcid )
WHERE
  d.size_actual > 0 AND 
  d.size_actual != ds.size_claimed AND 
  ds.entry_last_updated > $1 AND
  s.project = 1
LIMIT $2
OFFSET $3
```

is wouldn't actually select rows with d.size_actual= 10 and  ds.size_claimed=null. While we should make sure those content rows do get updated.


joshJarr (approved):
👍 LGTM

PR #1598, opened 2022-06-29
fix: don't prompt free

dchoi27:
Would be good to have an engineer's eyes on this @JeffLowe to make sure this doesn't break anything


drewdelano (approved):
LGTM

PR #1600, opened 2022-06-30
fix: remove duplicate data call

drewdelano:


adamalton (commented):
These changes look generally good to me, although something that I've noticed is that in the 'website' package the `getInfo` function is used by the `useUser` function, which is used in quite a lot of places.

So although combining the `/user/account` and `/user/info` endpoints seems like it makes a lot of sense, are we actually shooting ourselves in the foot here by now causing additional fetching of the used storage and tag proposals in a bunch of places where we don't need to? I think it depends on whether all the various uses of `useUser` actually cause a re-fetching of data from the API or not.

I'm not very familiar with the FE website code, so maybe you can answer that question more quickly than I can..?

Other than that one possible issue, which maybe isn't an issue, these changes look really nice 👌

drewdelano:
It looks like `useUser()` calls `useQuery()` which does call the backend.  So we will be over returning data and running the new query every time we ask for updated user information:

```
    const { data, error } = await this._client
      .from('user_tag_proposal')
      .select()
      .eq('user_id', userId)
      .is('deleted_at', null)
```

I believe this should be pretty performant, but it is a tax we're paying to combine these two API calls.  I think the real question here is "is it worth it?" and I'm not sure.

However, looking at all of the places it's called I'm wondering if most/some of those `useUser()` calls should just be reading from the in-memory UserProvider which stores a cached version of those.  Some of them specifically refresh the user instance, but it seems like those would be better served by refreshing the provider instead...?

packages/api/test/user.spec.js
```diff
@@ -5,46 +5,18 @@ import { endpoint } from './scripts/constants.js'
 import { getTestJWT, getDBClient } from './scripts/helpers.js'
 import userUploads from './fixtures/pgrest/get-user-uploads.js'
 
-describe('GET /user/account', () => {
-  it('error if not authenticated with magic.link', async () => {
```
jsdevel, line 9: is there a reason you're removing these tests?

drewdelano: we're basically merging the `account` and `info` endpoints so these tests are now redundant with the `info` tests.  though i suppose i could copy the asserts over to prove that functionality still works.

drewdelano: fixed: https://github.com/web3-storage/web3.storage/pull/1600/commits/c23632f3d8f84cdceaf8b19108f7cadb9bce56d4#diff-fd06e8f47f44010478243d794d7e859dc226dbda62aa8796511d5b86bc00d207R38

jsdevel:
@JeffLowe sounds accurate to me @JeffLowe .  @drewdelano I feel for now we should just close this out and leave it functioning as is.

PR #1611, opened 2022-07-04
fix: update content dag_size only if it has changed

flea89:
## Context 
In this #1535 we moved to use a direct connection to cargo to avoid expensive FDW cross-joins.

As you can read in the [PR](https://github.com/web3-storage/web3.storage/pull/1535) description we also assumed cargo would pull the updated `claimed_dag_size`.
Unfortunately, that is not the case and that means we are updating the same content multiple times, despite it having been already fixed by the previous run.

## Solution
While more optimized solutions exist, given this runs in a cron job and we've been changing it so many times, I'm going with a naive solution here, where I just update dag size if it has changed. 
That means we still have a go through lots of content that should really be looked at. 

However, this shouldn't be a big problem, given this happens in a cron, and we mostly run this on new content anyway.

The same job is meant to be used to update historical data, I've added a `before` input so that we can confine the updates to specific ranges.

### DIFF
--- a/packages/cron/test/cargo.spec.js
+++ b/packages/cron/test/cargo.spec.js
@@ -1,6 +1,6 @@
 /* eslint-env mocha */
 import assert from 'assert'
-import { createCargoDag, createUpload, createUser, createUserAuthKey, dbEndpoint, getUpload, listUploads, randomCid, token } from '@web3-storage/db/test-utils'
+import { createCargoDag, createUpload, createUser, createUserAuthKey, dbEndpoint, getContents, getUpload, listUploads, randomCid, token } from '@web3-storage/db/test-utils'
 import { DBClient } from '@web3-storage/db'
 import { updateDagSizes } from '../src/jobs/dagcargo.js'
 import { getCargoPgPool, getPg } from '../src/lib/utils.js'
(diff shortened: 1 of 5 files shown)


packages/cron/test/cargo.spec.js
```diff
@@ -181,15 +181,28 @@ describe('Fix dag sizes migration', () => {
...
+
+    const beforeContents = await getContents(dbClient, toBeUpdated)
+    await updateDagSizesWrp({ user })
+    const afterFirstRoundContents = await getContents(dbClient, toBeUpdated)
+
+    const updatedCidsFirstRound = afterFirstRoundContents.filter((cAfter) => {
+      const beforeContent = beforeContents.find((cBefore) => cAfter.cid === cBefore.cid)
+      return beforeContent?.updated_at !== cAfter.updated_at
+    })
+    assert.strictEqual(updatedCidsFirstRound.length, toBeUpdated.length)
```
adamalton, line 197: Would it be better to compare the actual CIDs here, to ensure that the correct items were updated, rather than just checking that the correct number were updated?
```suggestion
    toBeUpdatedCids = toBeUpdated.map(c => ci.cid)
    assert.strictEqual(updatedCidsFirstRound.slice().sort(), toBeUpdatedCids.slice().sort())
```

flea89: Nice one!

PR #1615, opened 2022-07-04
feat: stop tracking remote pins and remote them from db

### LINKED ISSUE #1098: Don't keep track of Pins with remote status

flea89:
Resolve #1098 

Follow-up PR #1619 to update db type once this is merged, release, and cron have been run.


packages/db/postgres/migrations/020-delete-remote-pins.sql
```diff
@@ -0,0 +1,17 @@
...
+-- According to documentation https://www.postgresql.org/docs/7.2/locking-tables.html deleting rows should use a 
+-- `RowExclusiveLock`. 
+-- This could have a negative performance impact in case those rows are being updated by other queries.
+-- However in our specific case I don't think remote pin are being really ever touched so it should be safe to just run 
+-- the query.
+-- Not sure about what statement_timeout we have in prod? Could that be a problem?
+--
+-- Alternative approach would be to create a cron job and delete in batches.
+
+DELETE FROM pin WHERE status='Remote'
```
flea89, line 17: Still need to look into/think about cascading deletes

adamalton: In terms of performance, the `status` column has got an index, and like you say, these rows are unlikely to be being touched by other processes. But my concern is about how much strain that single `DELETE` statement puts on the DB instance/memory/CPU in one go. from the docs:

> PostgreSQL actually treats every SQL statement as being executed within a transaction. If you do not issue a BEGIN command, then each individual statement has an implicit BEGIN and (if successful) COMMIT wrapped around it.

I don't really know how postgres transactions work under the hood, but I imagine that a single operation on millions of rows _might_ be holding quite a lot of stuff in memory, and so it might just hog the resources of the machine, even if there isn't any conflict on the rows that it's touching.

So I'm kind of leaning towards the cron job idea.

adamalton (commented):
Aside from the individual comments I've put, my overall question is: how did these 'Remote' pins come into existence in the first place, and will more of them crop up in the future? Do we need a continuous way of pruning them? I can't see in the code where/how they come to be.

.github/workflows/delete-remote-pins.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+jobs:
+  delete-remote-pins:
+    name: Delete remote pins
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        env: [
+          'staging', 
+          # 'production' Enable once tested in staging
```
flea89, line 19: Disabled Prod for now.
Let's enable once we tested this on staging.

vasco-santos: Maybe we can use env as input to not need to change this for running prod?

.github/workflows/delete-remote-pins.yaml
```diff
@@ -0,0 +1,44 @@
...
+        with:
+          node-version: 16
+      - name: Install dependencies
+        uses: bahmutov/npm-install@v1
+      - name: Run job
+        env:
+          DEBUG: '*'
+          ENV: ${{ matrix.env }}
+          STAGING_RO_PG_CONNECTION: ${{ secrets.STAGING_PG_CONNECTION }} # no replica for staging
+          PROD_RO_PG_CONNECTION: ${{ secrets.PROD_RO_PG_CONNECTION }}
```
vasco-santos, line 41:
```suggestion
          STAGING_PG_CONNECTION: ${{ secrets.STAGING_PG_CONNECTION }} # no replica for staging
          PROD_PG_CONNECTION: ${{ secrets.PROD_RO_PG_CONNECTION }}
```

packages/cron/test/delete-remote-pins.spec.js
```diff
@@ -0,0 +1,111 @@
...
+
+    const user = await createUser(dbClient)
+    cid = await randomCid()
+    await createUpload(dbClient, user._id, undefined, cid, {
+      pins: [...remotePins, ...otherPins]
+    })
+    const pins = await dbClient.getPins(cid)
+    assert.strictEqual(pins.length, remotePins.length + otherPins.length)
+
+    await deleteRemotePins({ rwPgPool, batchSize: 5 })
```
vasco-santos, line 86:
```suggestion
```

is this it?

vasco-santos (changes requested):

packages/cron/test/delete-remote-pins.spec.js
flea89: Good catch, moved this to beforeEach but havenn't deleted here.

vasco-santos (approved):
LGTM

.github/workflows/delete-remote-pins.yaml
flea89: @vasco-santos actually, shouldn't this be?
```
PROD_PG_CONNECTION: ${{ secrets.PROD_PG_CONNECTION }}
```

vasco-santos: oh yes sorry

.github/workflows/delete-remote-pins.yaml
```diff
@@ -0,0 +1,47 @@
...
+        with:
+          node-version: 16
+      - name: Install dependencies
+        uses: bahmutov/npm-install@v1
+      - name: Run job
+        env:
+          DEBUG: '*'
+          ENV: ${{ github.event.inputs.env }}
+          STAGING_PG_CONNECTION: ${{ secrets.STAGING_PG_CONNECTION }} # no replica for staging
+          PROD_PG_CONNECTION: ${{ secrets.PROD_RO_PG_CONNECTION }}
```
vasco-santos, line 44:
```suggestion
          STAGING_PG_CONNECTION: ${{ secrets.STAGING_PG_CONNECTION }} # no replica for staging
          PROD_PG_CONNECTION: ${{ secrets.PROD_PG_CONNECTION }}
```

vasco-santos (changes requested):

vasco-santos (approved):

Issue #1597, opened 2022-06-29
Implement blog subscription api endpoint


JeffLowe:
Blog developers need a subscribe api endpoint for the "subscribe to blog" functionality. 
The desire is something similar in function to nft implementation https://github.com/nftstorage/nft.storage/blob/main/packages/website/lib/subscribe.js

Blog developer will do the frontend integration ( via this branch https://github.com/web3-storage/web3.storage/tree/feat/blog ) we are to implement just the api endpoint for them to use. 

- [x] implement stub immediately to unblock blog development
- [ ] once email service determined, complete buildout of endpoint.

JeffLowe:
With the stub in place unblocking blog development, we now await decision on mail service so we can build out & connect the endpoint.

joannaong:
hi @JeffLowe do you have ETA for the final API? cc @orvn

PR #1617, opened 2022-07-05
fix: add blog subscription stub

drewdelano:
Partially completes #1597

### DIFF
--- a/packages/api/src/blog.js
+++ b/packages/api/src/blog.js
@@ -1,0 +1,16 @@
+import { JSONResponse } from './utils/json-response.js'
+
+/**
+ * @param {import('./user').AuthenticatedRequest} request
+ * @param {import('./env').Env} env
+ * @param {import('./index').Ctx} ctx
+ * @return {Promise<JSONResponse>}
+ */
+export async function blogSubscriptionCreate (request, env, ctx) {
+  const requestData = await request.json()
+  const { email } = requestData
+
+  // placeholder for actual subscribe logic
+
+  return new JSONResponse({ email }, { status: 202 })
+}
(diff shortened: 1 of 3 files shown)


packages/api/src/blog.js
```diff
@@ -0,0 +1,16 @@
+import { JSONResponse } from './utils/json-response.js'
+
+/**
+ * @param {import('./user').AuthenticatedRequest} request
+ * @param {import('./env').Env} env
+ * @param {import('./index').Ctx} ctx
+ * @return {Promise<JSONResponse>}
+ */
+export async function blogSubscriptionCreate (request, env, ctx) {
```
jsdevel, line 9: can this be removed and `lib/api.js` used instead?

drewdelano: I might be misunderstanding this so bear with me:

I think `lib/api.js` is a client side wrapper around the API call that goes into the `blogSubscriptionCreate` API

so they're the same thing, one i just a client side wrapper around a server side endpoint

jsdevel: ok.  so it looks like this is the only `blogSubscriptionCreate` that's being used.  i don't see the one you added in the `website` package imported from anywhere.

drewdelano: It isn't being used anywhere yet.  The frontend is being done by a different group.  I think it's Potato.  @JeffLowe do you recall which team it is?

drewdelano: Talked with @jsdevel offline.  It sounds like he'd prefer I remove the `lib/api.js` bit so it's just a backend stub.  Which seems fine.

jsdevel (approved):

PR #1627, opened 2022-07-07
fix: add source_cid index on upload table

ericssontadashi:
It was using a composite index before.

image


packages/db/postgres/tables.sql
```diff
@@ -269,6 +269,7 @@ CREATE INDEX IF NOT EXISTS upload_auth_key_id_idx ON upload (auth_key_id);
 CREATE INDEX IF NOT EXISTS upload_user_id_deleted_at_idx ON upload (user_id) WHERE deleted_at IS NULL;
 CREATE INDEX IF NOT EXISTS upload_content_cid_idx ON upload (content_cid);
 CREATE INDEX IF NOT EXISTS upload_updated_at_idx ON upload (updated_at);
+CREATE INDEX IF NOT EXISTS source_cid_idx ON upload (source_cid);
```
vasco-santos, line 272:
```suggestion
CREATE INDEX IF NOT EXISTS upload_source_cid_idx ON upload (source_cid);
```

packages/db/postgres/migrations/020-add-source-cid-index.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX CONCURRENTLY IF NOT EXISTS source_cid_idx ON upload (source_cid);
```
vasco-santos, line 1:
```suggestion
CREATE INDEX CONCURRENTLY IF NOT EXISTS upload_source_cid_idx ON upload (source_cid);
```

vasco-santos (changes requested):
Let's keep names consistent between indexes for each table :)

vasco-santos (approved):
LGTM thanks

PR #1628, opened 2022-07-07
refactor: remove cron to delete remote pin rows

vasco-santos:
We ran https://github.com/web3-storage/web3.storage/pull/1615 to delete all the not needed remote pin status rows.

So, we can drop all this code :)


flea89 (approved):
Thanks, LGTM!

PR #1630, opened 2022-07-08
feat: copy change to pricing

dchoi27:


JeffLowe (approved):
Reviewed and confirmed.  copy only change to pricing page.

PR #1634, opened 2022-07-11
feat: dag size job improvements

flea89:
A few quick wins to improve dag size job:
- run in prod only (both staging and prod run against the same cargo replica, let's avoid adding unnecessary load)
  - still able to run against staging manually 
- run on a shorter time frame by default
- longer timeout for manual distpatch to be able to run in against longer timeranges


adamalton (approved):
LGTM

PR #1637, opened 2022-07-12
feat: contact

joannaong:
new contact page


orvn (approved):
@dchoi27 I'm merging this into `main`, but one thing to note is that the navigation was too crowded to add a _Contact_ link. It's in the footer of every page, but in the future we might want to add it to the main navigation.

PR #1641, opened 2022-07-14
Feat/admin email improv

### LINKED ISSUE #1640: Adding extra info on admin storage notification emails
## Overarching user needs / context:
Administrators need more information on user exceeding their used storage, to be able to assess the right action to take, i.e. block or not, need to contact yes/no, etc.

## Acceptance Criteria
- Include the absolute storage amount used by each user who is using more than 100% (on top of the percentage)
- Include the date of last upload for each user as well

flea89:
Resolve #1640

### DIFF
--- a/packages/cron/src/jobs/storage.js
+++ b/packages/cron/src/jobs/storage.js
@@ -26,6 +26,20 @@ const USER_BY_EMAIL_QUERY = `
   SELECT id AS _id , email, name FROM public.user WHERE email=$1
 `
 
+const GET_DATE_LATEST_UPLOAD_OR_PSA_REQUEST = `
+  SELECT max(updated_at) as last_updated
+  FROM (
+    SELECT updated_at
+    FROM upload up
+    WHERE user_id=$1
+    UNION ALL
+    SELECT psa.updated_at
+    FROM psa_pin_request psa
+    JOIN auth_key a ON a.id = psa.auth_key_id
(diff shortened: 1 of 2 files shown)


packages/cron/src/jobs/storage.js
```diff
@@ -179,11 +193,23 @@ export async function checkStorageUsed ({ roPg, emailService, userBatchSize = 20
       throw new Error(`${supportEmail} does not exists`)
     }
 
+    const usersOverQuotaAugmented = await Promise.all(
+      usersOverQuota.map(async u => {
+        const { rows } = await roPg.query(GET_DATE_LATEST_UPLOAD_OR_PSA_REQUEST, [
```
francois-potato, line 198: This might become a bit slow.  Could bulk fetch `GET_DATE_LATEST_UPLOAD_OR_PSA_REQUEST` with a list of user IDs instead of a single ID and then merge.

francois-potato (approved):

packages/cron/src/jobs/storage.js
flea89: Good point.
However this list **should** never grow, the idea is that users are disabled once they over quota and they wouldn't appear in that list anymore. 
So not really sure is worth "complicating" this? What do you think?

francois-potato: Agreed it is fine as it is 👍

PR #1650, opened 2022-07-18
feat: update email docs

### LINKED ISSUE #1506: Tidy up quota exceeded email
Some suggestions to tidy up the quota exceeded email:

Screenshot 2022-06-17 at 15 03 01

flea89:
Resolve #1506 

As part of this I update the templates in Mailchimp, please review.

At the moment updating email styles and templates is quite painful, I've timeboxed this to the quick wins:

- casing, :white_check_mark: :white_check_mark:
- logo :white_check_mark: :white_check_mark:
- round-corners :white_check_mark: :white_check_mark:
- Padding adjusted, it’s not the same but closer…  I think it now has a better balance
- Title, I think left aligned looks better
- Haven't changed the email to show the quota, given it requires changing the jobs and it was deemed as not worth it by David

Once reviewed, I will export to mandril and to the required changes.

### DIFF
--- a/packages/cron/src/lib/email/README.md
+++ b/packages/cron/src/lib/email/README.md
@@ -34,8 +34,8 @@ but the templates that you create in Mailchimp can't be used directly for transa
 (1) they need to be exported to Mandrill first, and
 (2) they need some manual adaptations and fixes in order to fully work.
 
-Be aware that once you send a template to Mandril that templates is automatically published.
-If the template requires some code changes, makes sure those are merged and deployed first.
+Once you send a template to Mandril that template is automatically published.
+If the work requires some code changes, those needs to be merged and deployed first.
 Alternatively duplicate the template, update it and make sure you update the ID to the new one in your PR.
 Even if no code changes are required, please allow for some buffer between exporting the template to Mandrill and the mail is supposed to go out. You need some time go through the list below.
 
--- a/packages/cron/src/lib/email/README.md
+++ b/packages/cron/src/lib/email/README.md
@@ -45,30 +45,27 @@ To add or edit a template:
 * Export the template from Mailchimp to Mandrill using the "Send to Mandrill" option.
 * Go to the [templates page in Mandrill](https://mandrillapp.com/templates)
 * Click on the template to edit the HTML, and make the following changes/fixes:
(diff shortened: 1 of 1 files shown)


packages/cron/src/lib/email/README.md
```diff
@@ -34,25 +34,37 @@ but the templates that you create in Mailchimp can't be used directly for transa
...
   background: linear-gradient(to top left, #d169db, rgba(199, 166, 251, 0), #5da9e7), linear-gradient(to top right, #5a69da, rgba(199, 166, 251, 0), #d4a8e7) rgba(199, 166, 251, 1);
 }
+
+#templateHeader {
+    border-radius: 6px 6px 0 0;
+}
+
+#templateBody {
+    border-radius:  0 0 6px 6px;
+}
```
adamalton, line 67: Do you know that this doesn't get retained when you export the template to Mandrill? The rounded corners look like they're working fine in Mailchimp, so my guess is that that CSS will still be working anyway when the template is exported.

adamalton (approved):
From looking at the templates in Mailchimp, they look good.

Have you looked at how to fix the repeating background? I assume that it need to be moved to a more outer element?

packages/cron/src/lib/email/README.md
flea89: I wasn't sure actually, lots of guesswork involved in this. So I thought to just add it to be sure.
You probably know more than me on this, so happy to follow your lead.

flea89:
I couldn't actually reproduce the background issue, are you @adamalton?

packages/cron/src/lib/email/README.md
adamalton: I tested this out, and even though the style seems to exist in Mailchimp, it gets put into a `<style>` tag in the body rather than in the head, and so doesn't seem to work. So this addition to the instructions is needed.

adamalton:
I've now tested this, updated the instructions, and ported/converted all the template updates over to Mandrill.

This process is incredibly painful, so if we want to make more changes we should consider switching email provider.

adamalton:
And re. the background issue, I think it was probably because in the instructions we had:

```css
body,#bodyTable {
    background: linear-gradient ... 
}```

which should have just been:

```css
#bodyTable {
    background: linear-gradient ...
}```

I've now fixed that in the instructions.

PR #1655, opened 2022-07-19
refactor: remove pinata

### LINKED ISSUE #1015: Pin Requests to pinata only when IPFS Cluster has content pinned
We have currently two issues on request pins to Pinata:
- We do chunk uploads and ask both Pinata and our cluster to Pin a root CID right after first chunk arrives
  - Uploads that don’t fully make it will have their root CID pending there forever
- Pinning API Usage - We ask to Pin CIDv1 but the Pin could be created with a CIDv0: https://github.com/web3-storage/web3.storage/issues/918

Potential solutions:
1. Add rows to PinRequest table only when Pin Status is Pinned on our Cluster, instead of right away in the createUpload SQL Func. That is, insert to PinRequest table on the pins cron job when status is Pinned. With this, we can also downgrade CID to v0 and if the v0 is pinned we can ask v0 instead (caveat that currently pin_request.content_cid is FK of content (cid)
2. in the cron job that sends the pins to pinata, could we join with the pins/locations tables and only send to pinata when pinned? otherwise leave it in the queue...

2 would not fix the Pinning API usage problem. We can just insert the `source_cid`  and drop the FK to content, that will make things complex then as we won’t be able to join with content + pin to see the status

### LINKED ISSUE #1620: Drop pin sync request table and remove insertions from upload flow
We don't use pin sync requests anymore (these were to track Pinata ongoing pins).

We are accumulating unecessary data in our DB and having an extra DB insertion per upload chunking, so uploads might even get a bit faster

alanshaw:
This PR removes the `pinpin` package, removes the `pin_requests` DB table, updates the `create_content` DB function to not insert into the `pin_requets` table, adds a DB migration to do those two things and removes the `*PinRequest` related functions from the DB client.

We will need to separately stop and remove the running pinpin job on Heroku.

This effectively stops Web3.Storage pinning to Pinata.

closes #1015
resolves #1620


packages/db/index.d.ts
```diff
@@ -17,19 +17,15 @@ import type {
...
   PinsUpsertInput,
   BackupOutput,
   PsaPinRequestItem,
   PsaPinRequestUpsertOutput,
   PsaPinRequestUpsertInput,
-  ContentInput,
   ListPsaPinRequestOptions,
   ListPsaPinRequestResults,
-  UserStorageUsedInput,
-  UserStorageUsedOutput,
```
alanshaw, line 32: These are unused in this file.

packages/db/index.d.ts
```diff
@@ -74,7 +68,6 @@ export class DBClient {
   getPsaPinRequest (authKey: string, pinRequestId: string) : Promise<PsaPinRequestUpsertOutput>
   listPsaPinRequests (authKey: string, opts?: ListPsaPinRequestOptions ) : Promise<ListPsaPinRequestResults>
   deletePsaPinRequest (pinRequestId: string, authKey: string) : Promise<PsaPinRequestItem>
-  createContent (content: ContentInput, opt?: {updatePinRequests?: boolean}) : Promise<string>
```
alanshaw, line 77: This function does not exist in the DB client but references pin requests in it's options. I've removed it from here to avoid any confusion.

vasco-santos (approved):
LGTM 💅🏼

PR #1660, opened 2022-07-20
fix: cron pins skips remote status

vasco-santos:
Our cron job is writing remote status pins in the table...


flea89 (approved):
Good catch!

PR #1661, opened 2022-07-20
chore: bring back cron to remove remote status pins

vasco-santos:
Brings back job from https://github.com/web3-storage/web3.storage/pull/1615 because of regression introduced with cron job https://github.com/web3-storage/web3.storage/pull/1660


flea89 (approved):
LGTM

PR #1664, opened 2022-07-20
fix: update user req modals to match site

cmunns:
- update textarea height to be more aligned with expected input

### DIFF
--- a/packages/website/styles/layouts/userRequestsModals.scss
+++ b/packages/website/styles/layouts/userRequestsModals.scss
@@ -15,21 +15,24 @@
   }
 
   .input-container {
-    margin-top: 30px;
+    margin-top: 50px;
   }
 
   label {
     display: inline-block;
     line-height: 1.4;
+    font-size: 0.9rem;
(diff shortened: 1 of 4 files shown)


packages/website/styles/layouts/userRequestsModals.scss
```diff
@@ -15,21 +15,24 @@
...
 
   textarea {
     display: block;
     width: 100%;
-    height: 150px;
     line-height: 1.5;
-    padding: .5rem;
-    resize: none;
+    padding: 0.5rem;
+    // resize: none;
```
e-schneid, line 32: do we want this commented or removed?

e-schneid (approved):

packages/website/styles/layouts/userRequestsModals.scss
johnathonroach: cmunns is OOO - I'll go ahead and remove.

johnathonroach (approved):

PR #1669, opened 2022-07-22
fix: show only a single selected nav item at a time

### LINKED ISSUE #1558: Docs nav shows many sections as active on scroll
The right hand side nav in the docs shows all sections with the same name as active **at the same time** when scrolling. Only the current section in view should be highlighted.

w3-active-index-chaos

e-schneid:
Closes #1558


e-schneid:
Revise this so that the first id doesn't have a suffix (i.e., remove `-1`).  As it is, it won't work properly with footer links.

jsdevel (approved):

PR #1674, opened 2022-07-25
fix: increase cron pins timeout to 15 mins

vasco-santos:
This cron job timeouts once in a while https://github.com/web3-storage/web3.storage/actions/workflows/cron-pins.yml. Let's use 15min instead


flea89 (approved):
LGTM

PR #1675, opened 2022-07-25
fix: cron remote status

vasco-santos:
It was still possible for the cron to write Remote status given status variable could be changed for it. Ran today the delete job and found a few more that came from there https://github.com/web3-storage/web3.storage/runs/7496383579?check_suite_focus=true#step:6:43


flea89 (approved):
Good catch, LGTM

PR #1683, opened 2022-07-27
feat: embed w3name package

adamalton:
This removes the inbuilt `packages/client/src/name.js` library (now moved into w3name) and replaces it with an import of the new `w3name` package.

Note that we've still got other w3name-related code in `packages/api/name.js`, which this PR doesn't touch.


packages/w3/package.json
```diff
@@ -31,6 +31,7 @@
     "ora": "^5.4.1",
     "sade": "^1.7.4",
     "uint8arrays": "^3.0.0",
+    "w3name": "^1.0.0",
```
francois-potato, line 34: Should we point to the latest 1.0.3 version?

packages/client/package.json
```diff
@@ -45,7 +45,8 @@
     "p-retry": "^4.5.0",
     "streaming-iterables": "^6.2.0",
     "throttled-queue": "^2.1.2",
-    "uint8arrays": "^3.0.0"
+    "uint8arrays": "^3.0.0",
+    "w3name": "^1.0.0"
```
francois-potato, line 49: Should we point to the latest 1.0.3 version?

packages/w3/package.json
adamalton: Good point. Done.

francois-potato (approved):

PR #1685, opened 2022-07-27
chore: ipns record migration to w3name

francois-potato:
This script will run as a CRON job to push name records to w3name.

If the record was already migrated or superseded you will see the following error message `400 Bad Request {"message":"invalid record: the record is outdated."}` otherwise `202 Accepted {"id":"k51qzi5uqu5dkf4iy7zpxjasuk2b6x4wqr9vp7cd8m3x2kdlav02ppw9d9ulwf"}`.

### DIFF
--- a/.github/workflows/cron-names.yml
+++ b/.github/workflows/cron-names.yml
@@ -1,0 +1,44 @@
+name: Cron migrate IPNS records to w3name
+
+on:
+  schedule:
+    - cron: '* */24 * * *'
+  workflow_dispatch:
+
+jobs:
+  update:
+    name: Export IPNS records to w3name
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        env: ['staging', 'production']
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v2
+        with: 
+          fetch-depth: 0
+      - name: Checkout latest cron release tag
+        run: |
+          LATEST_TAG=$(git describe --tags --abbrev=0 --match='cron-*')
+          git checkout $LATEST_TAG
+      - name: Setup node
+        uses: actions/setup-node@v1
+        with:
+          node-version: 16
+
+      - name: Install dependencies
+        uses: bahmutov/npm-install@v1
+
+      - name: Run job
+        env:
+          DEBUG: '*'
+          ENV: ${{ matrix.env }}
+          PROD_PG_REST_JWT: ${{ secrets.PROD_PG_REST_JWT }}
+          STAGING_PG_REST_JWT: ${{ secrets.STAGING_PG_REST_JWT }}
+          PROD_PG_REST_URL: ${{ secrets.PROD_PG_REST_URL }}
+          STAGING_PG_REST_URL: ${{ secrets.STAGING_PG_REST_URL }}
+          CLUSTER_API_URL: ${{ secrets.CLUSTER_API_URL }}
+          CLUSTER_BASIC_AUTH_TOKEN: ${{ secrets.CLUSTER_BASIC_AUTH_TOKEN }}
+          CLUSTER_IPFS_PROXY_API_URL: ${{ secrets.CLUSTER_IPFS_PROXY_API_URL }}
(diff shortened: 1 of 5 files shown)


packages/cron/src/bin/names.js
```diff
@@ -0,0 +1,20 @@
...
+async function main () {
+  const env = process.env.ENV || 'dev'
+  const db = getDBClient(process.env)
+
+  await postNamesToW3name({ env, db })
+}
+
+envConfig()
+
+main()
```
adamalton, line 20: What you've got here is the same as the other cron jobs we've got, but should we be doing `await main()` here instead?

packages/cron/src/jobs/names.js
```diff
@@ -0,0 +1,50 @@
+import debug from 'debug'
+import retry from 'p-retry'
+
+const MAX_REQUEST_PAGE = 10
+const W3NAME_URL = 'http://localhost:8989'
```
adamalton, line 5: Should this be switchable via an environment variable for production?

packages/cron/src/bin/names.js
francois-potato: Good point, but if I add `await`, it complains about having an async call inside a non-async context. Other cron jobs don't use it too.

adamalton: Ah yeah, sorry, my bad. I'd forgotten that it doesn't apply on the outermost function 🤦 .

.github/workflows/cron-names.yml
```diff
@@ -0,0 +1,52 @@
...
+        with:
+          node-version: 16
+
+      - name: Install dependencies
+        uses: bahmutov/npm-install@v1
+
+      - name: Run job
+        env:
+          DEBUG: '*'
+          ENV: ${{ matrix.env }}
```
adamalton, line 43: From looking at this is done in other configs, I think that maybe `matrix.env` will only get the value from the `matrix` setting, not from the `inputs` section in the `workflow_dispatch`.

```suggestion
          ENV: ${{ matrix.env || github.event.inputs.env }}
```

This is only a guess though. [Stack Over doesn't seem to know!](https://stackoverflow.com/questions/69781005/combine-dynamic-github-workflow-matrix-with-input-values-and-predefined-values)

packages/cron/src/jobs/names.js
```diff
@@ -0,0 +1,73 @@
+import debug from 'debug'
+import retry from 'p-retry'
+
+const MAX_REQUEST_PAGE = 10
+const W3NAME_API_URL_STAGING = 'https://w3name-staging.protocol-labs.workers.dev/'
+const W3NAME_API_URL_PRODUCTION = 'https://w3name-production.protocol-labs.workers.dev/'
```
adamalton, line 6:
```suggestion
const W3NAME_API_URL_STAGING = 'https://name.web3.storage/'
const W3NAME_API_URL_PRODUCTION = 'https://name-staging.web3.storage/'
```

This might be slightly preemptive, as [w3name/#23](https://github.com/web3-storage/w3name/pull/23) isn't merged yet, even if it's not set up this way yet, it will be. So I think it's worth changing our URLs here.

adamalton (commented):
Other than my two suggestions above, I think this looks good 👍 .

adamalton (approved):

packages/cron/src/jobs/names.js
```diff
@@ -0,0 +1,73 @@
+import debug from 'debug'
+import retry from 'p-retry'
+
+const MAX_REQUEST_PAGE = 10
+const W3NAME_API_URL_STAGING = 'https://name.web3.storage/'
+const W3NAME_API_URL_PRODUCTION = 'https://name-staging.web3.storage/'
```
joshJarr, line 6: @adamalton @francois-potato Is this reversed for a reason? Should staging point to staging instead?

adamalton: Ooh, no! Good spot!

PR #1686, opened 2022-07-27
chore: remove /name API endpoints and storage

adamalton:
Completes #1703.

The `w3name` client library (and the proxy for it in `web3.storage/name`) should be released and working (so that all IPNS-related requests are going to the new `w3name` Cloudflare Worker, not to this one, before we release this.

* Removes `/name/*` API endpoints.
* Removes `NameRoom` Durable Object class
    - and thereby the associated websocket "rooms".
* Removes `ipns` dependency from package.json.

This functionality has moved to the separate w3name package.


joshghent (approved):
nice work, lgtm

PR #1687, opened 2022-07-27
feat: blog subscribe

### LINKED ISSUE #1597: Implement blog subscription api endpoint

johnathonroach:
Fixes #1597 
This PR adds a subscribe endpoint using mailchimp.


packages/api/src/blog.js
```diff
@@ -7,10 +8,19 @@ import { JSONResponse } from './utils/json-response.js'
...
+  const headers = {
+    'Content-Type': 'application/json; charset=UTF-8',
+    Authorization: `Basic: ${apiKey}`
+  }
 
-  // placeholder for actual subscribe logic
-
-  return new JSONResponse({ email }, { status: 202 })
+  const body = await request.json();
+  (await isChimpUser(body.email, prefix, listId, headers))
```
dashcraft, line 20: suggestion (non-blocking):
Personaly preference but there's a bit of complexity trying to grok this mentally, it might be better to just create meaningful variable names and well formed code paths instead of the 3 awaited functions + ternary

packages/api/src/utils/mailchimp.js
```diff
@@ -0,0 +1,58 @@
+/**
+ *  @param {string} email
+ *  @param {string} prefix
+ *  @param {string} listId
+ *  @param {object} headers
+ */
+export const isChimpUser = async (email, prefix, listId, headers) => {
```
dashcraft, line 7: comment:
return value

dashcraft (approved):
Some minor comments and suggestions, I'm approving, but if you can take a look at the comments and fix the missing return values, and double check the syntax/structure of the awaited functions, that'd be great!

johnathonroach:
@JeffLowe Before deployment, we need to generate and update the Web3 environment with three new env vars for the Mailchimp subscription functionality.

```
MAILCHIMP_API_KEY
MAILCHIMP_BLOG_LIST_ID
MAILCHIMP_API_PREFIX
```

packages/api/src/blog.js
johnathonroach: @dashcraft agreed, explicit simplicity applied.

Issue #1689, opened 2022-07-28
Create documentation for w3name


mbommerez:
We need to document:
* w3name functionalities
* the JS client usage
* the w3name API

Some current docs in readme
* https://github.com/web3-storage/w3name
* https://github.com/web3-storage/w3name/blob/main/packages/client/README.md

mbommerez:
Subticket
https://github.com/web3-storage/w3name/issues/27

mbommerez:
This is live in Production

![Image](https://user-images.githubusercontent.com/93770197/183436865-eae938da-95a7-4101-9308-6e111175f016.png)

PR #1694, opened 2022-08-01
feat(website/docs): add draft of w3name how-to

yusefnapora:
This is an early draft of the how-to article for using w3name.

@francois-potato does the client support the `/name/:key/watch` endpoint yet? I didn't see it in the code, but maybe it's on a branch or something? If not, I'll pull that section from the doc.

As a side note, it would be good to publish a couple of example records, so we can have valid `/ipns/` links in the docs that will always resolve.

Things left to do:

- [ ] test & validate all code examples
- [ ] add example that uses web3.storage client to get CIDs to publish
- [ ] Show how to make gateway URLs for published names
- [ ] revise for grammar, etc
- [ ] link to API docs (once published)
- [ ] `/:key/watch` endpoint - either add to docs or remove section

Bonus:

- [ ] `ipfs get /ipns/record-we-just-published` example, to show off interop
- [ ] curl example of resolving latest value (no signature verification - binary record ignored for simplicity)


joshJarr:
Amazing stuff @yusefnapora! 

I have two quick questions around the HTTP documentation.

1. I've noticed some web3.storage docs show cURL examples, should we do the same with W3Name? 
2. Where would be best to host the W3Name HTTP API docs?

    So we currently don't have any FE for the w3name API and rely on the product page ([web3.storage/products/w3name ](https://9fab523e.web3-storage.pages.dev/products/w3name/)) and these how-to docs. 

    So we could either: 

    - Have the w3name HTTP Docs on the product page (this could feel a little misplaced imo)
    - Have the w3name HTTP API Docs in these web3.storage docs (similarly to the [web3 http API](https://0b538102.web3-storage-staging.pages.dev/docs/reference/http-api/))
    - Have the w3name HTTP Docs on a separate URL, this would just be the standard swagger UI, and link out to it from the how-to.

Keen to get your opinion on this, either way we can expose the schema via _name.web3.storage/shema.yml_ so wherever we host the Swagger UI it should be able to fetch the latest documentation. 

Thanks!

adamalton:
This is really nice. Really well laid out and and clear.

There's just one bit which I think could do with being a bit clearer, which is in the **What's in a name?** section.

> The "names" in IPNS and w3name are unique identifiers for cryptographic verification keys. In fact, under the hood, each name is an IPFS CID (Content Identifier). The content it identifies is a libp2p key, which can use one of several signing algorithms (RSA, Ed25519, etc).

We say that a "name" is an identifier for a key, but then we say that a name is a content identifier. It feels a bit confused. Bear in mind that I might have misunderstood the details of how this works slightly myself, but here's my attempt at a slight rewording:

> The "names" in IPNS and w3name are unique identifiers, and each one is a string representation of the public half of a cryptographic key-pair. It can use one of several signing algorithms (RSA, Ed25519, etc).

> Each "name" maps to a "record", which contains:
> * The CID of the IPFS content that the name is pointing to.
> * A sequence number and validity date.
> * A verification signature, created using the private key, providing proof that this record was generated by the key holder.

I feel like maybe there's a nuance to this that (despite having worked with it) I haven't fully grasped. Maybe that's further evidence that a really clear explanation here would be useful! Are you able to incorporate this extra info about the structure of a record and exact relations of the CID(s) and what's contained within what?

mbommerez:
I linked this PR to ticket #1689.

yusefnapora:
@adamalton that's a great revision, thanks - the bit about the names technically being CIDs is not really relevant to how they're used and is confusing on a second read.

@joshJarr good question about where to put the HTTP docs. Right now I'm leaning toward putting them in the web3.storage docs site, next to the web3.storage HTTP api docs in the nav, only because it's a bit easier than setting up a new site, and like you said it doesn't quite fit on a "product page".

I think I'm going to mark this as ready to review, despite the unchecked TODOs in the initial comment. This seems like a good start, and we can add more examples and links to the HTTP docs, etc in another PR.

francois-potato (approved):

PR #1701, opened 2022-08-03
fix: limit production job should not fail if staging one fails

flea89:
## Context
The `Storage Limit Email Notifications` workflow use stategy matrix (as other worflows in the project, ie metrics, cron-pins etc) to run against different environments.

That's cool, however, by default strategy has `fail-fast: true` and that means all jobs are cancelled if one in the workflow fails.
This practically means that if staging fails prod is skipped/cancelled, creating a dependency/coupling between the 2 environments we don't want.


joshghent (approved):
lgtm

PR #1708, opened 2022-08-03
fix: throttled names cron job requests

francois-potato:
* Throttled requests to w3name to avoid rate limits


joshJarr:
LGTM! 
I've seen the throttle queue implementation done a little differently, and I think the idea is to avoid adding the request to the queue, See: https://github.com/web3-storage/web3.storage/blob/f80833ceb8c86086f535a20408a6febdff23140c/packages/client/src/lib.js

But I think this implementation in this PR is sound - just wanted to share for context, as it took me a while to understand the other implementation

joshJarr (approved):

PR #1710, opened 2022-08-04
fix: Make PSA Errors compliant

### LINKED ISSUE #1320: Show a useful error message to blocked users trying to use the pinning service
as per #1319 when adding a pin as a blocked user today we get:

```console
$ ipfs pin remote add --service=web3stage --name="ayy3.jpg" bafybeidluj5ub7okodgg5v6l4x3nytpivvcouuxgzuioa6vodg3xt2uqle
Error: reason: "", details: "": 403 Forbidden
```

this could explain why, and what we want the user to do about it if anything.

joshghent:
Resolves #1320

In investigating the issue above, we discovered that the pinning service API was not compliant with the IPFS Spec when throwing errors. Hence, why for the CLI, there were no errors shown.

To make the pinning service compliant with the spec (found [here](https://ipfs.github.io/pinning-services-api-spec/#operation/getPins)). I have nested the error and updated the properties to `reason` and `details`. I have updated the tests to match the specification.

The logic to decide whether we should use the IPFS style errors is in the error handler and based on the request path.
This seemed like the cleanest way to do it, but please let me know if you have other ideas.

@flea89 and I came up with other solutions:
1. Pass a "type" to the middleware that checks for PSA Authorization and then throws either HTTPErrors or PSAErrors.
2. Create new middleware that clones the other but uses PSAErrors

### DIFF
--- a/packages/api/src/error-handler.js
+++ b/packages/api/src/error-handler.js
@@ -17,11 +17,11 @@ export function errorHandler (err, { log }, request) {
   // As the pinning service requires a certain specification
   // Always return the nested error object for any pinning service endpoints
   const requestUrl = new URL(request.url)
-  if (err instanceof PinningServiceApiError || requestUrl.pathname.test(/^\/pins$/)) {
+  if (err instanceof PinningServiceApiError || (requestUrl && /^\/pins/.test(requestUrl.pathname))) {
     const error = {
       error: {
-        reason: err.reason || err.message,
-        details: err.details || err.code
+        reason: err.reason || err.code,
+        details: err.details || err.message
       }
     }
     return new JSONResponse(error, { status })
--- a/packages/api/src/error-handler.js
+++ b/packages/api/src/error-handler.js
@@ -18,7 +18,7 @@ export function errorHandler (err, { log }, request) {
   // As the pinning service requires a certain specification
   // Always return the nested error object for any pinning service endpoints
   const requestUrl = new URL(request.url)
-  if (err instanceof PinningServiceApiError || (requestUrl && /^\/pins/.test(requestUrl.pathname))) {
+  if (err instanceof PinningServiceApiError || /^\/pins/.test(requestUrl.pathname)) {
     const error = {
       error: {
         reason: err.reason || err.code,
--- a/packages/api/src/error-handler.js
+++ b/packages/api/src/error-handler.js
@@ -5,6 +5,7 @@ import { HTTPError, PinningServiceApiError } from './errors.js'
 /**
  * @param {Error & {status?: number;code?: string;reason?: string; details?: string; }} err
  * @param {import('./env').Env} env
+ * @param {Request} request
  */
 export function errorHandler (err, { log }, request) {
   console.error(err.stack)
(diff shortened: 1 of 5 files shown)


packages/api/src/error-handler.js
```diff
@@ -6,18 +6,23 @@ import { HTTPError, PinningServiceApiError } from './errors.js'
  * @param {Error & {status?: number;code?: string;reason?: string; details?: string; }} err
  * @param {import('./env').Env} env
  */
-export function errorHandler (err, { log }) {
+export function errorHandler (err, { log }, request) {
```
flea89, line 9: Could we please update the function annotation with request?

packages/api/src/error-handler.js
```diff
@@ -6,18 +6,23 @@ import { HTTPError, PinningServiceApiError } from './errors.js'
...
   let status = err.status || 500
   if (status >= 500) {
     log.error(err)
   }
 
-  if (err instanceof PinningServiceApiError) {
+  // As the pinning service requires a certain specification
+  // Always return the nested error object for any pinning service endpoints
+  const requestUrl = new URL(request.url)
+  if (err instanceof PinningServiceApiError || (requestUrl && /^\/pins/.test(requestUrl.pathname))) {
```
flea89, line 20: I see you're checking `requestUrl &&` 
Can requestUrl be `undefined` here? Wouldn't `new URL` throw for invalid URLs?

packages/api/src/error-handler.js
```diff
@@ -6,18 +6,23 @@ import { HTTPError, PinningServiceApiError } from './errors.js'
...
     const error = {
-      reason: err.reason,
-      details: err.details
+      error: {
```
flea89, line 22: nit. Please ignore this if you don't think is a good idea.

I was wondering if we coud add a `static` method on `PinningServiceApiError` to have an API like this:
````
const pErr = PinningServiceApiError.fromHTTPError(err)
```

I think it'd be neat to also have the error classes be responsible to encode themselves in Json errors, but that's defo outside of this ticket.

packages/api/src/errors.js
```diff
@@ -31,11 +46,11 @@ export class UserNotFoundError extends HTTPError {
 }
 UserNotFoundError.CODE = 'ERROR_USER_NOT_FOUND'
 
-export class PinningUnauthorizedError extends HTTPError {
+export class PinningUnauthorizedError extends PinningServiceApiError {
   constructor (msg = 'Pinning not authorized for this user, email support@web3.storage to request authorization.') {
     super(msg, 403)
     this.name = 'PinningUnauthorizedError'
-    this.code = PinningUnauthorizedError.CODE
+    this.reason = PinningUnauthorizedError.CODE
```
flea89, line 53: Shouldn't we refactor this to have `status` and `detail`?

flea89 (changes requested):
Nice work! I've tested this and now the right error messages are displayed by the client.

Left a few minor comments!

packages/api/src/error-handler.js
joshghent: Good point. Removed this check.

packages/api/src/errors.js
joshghent: Resolved. But only status because detail is set in the constructor.

packages/api/src/error-handler.js
joshghent: Yes that might be a neat idea actually if it ever comes up again. @mbommerez is it worth making a note of this somewhere?

adamalton (approved):
This looks great. I haven't re-tested it after Paolo's testing, but all the changes look good.

Issue #1507, opened 2022-06-17
Add pagination to the /pins endpoint 


joshJarr:
As discussed on the [previous pagination ticket](https://github.com/web3-storage/web3.storage/issues/1362#issuecomment-1140105729), it may make sense to add pagination on the pins endpoint. This will help when loading a large set of pins and keep the endpoint consistent with the `GET /user/uploads` endpoint.

JeffLowe:
Let us know once this lands @joshJarr so we can update the frontend to use it.

mbommerez:
@dchoi27 Are you happy for us to add this into the psa follow up initiative as well? This would make things a lot easier for the pins table in the front end, as it will mean they use the same structure as the uploads table.

Let us know if you need more info and we'll be happy to explain further why that is the case.
CC @joshJarr

joshghent:
This should have been done as part of https://github.com/web3-storage/web3.storage/pull/1718

PR #1718, opened 2022-08-09
feat: New Users Pins Endpoint

### LINKED ISSUE #1616: Display all pinned files in account when using multiple tokens
### Overarching user needs / context:

Currently users who use multiple tokens only find CIDs pinned using the first token displayed in their web3.storage account page.

The Pinning Services API is currently scoped by token, meaning it only writes or reads pins by token, not by user (web3storage account).

Additionally, some users will use the CLI to list pinned CIDs for 1 specific token. Other users will want to periodically change their token and might expect to see old CIDs (pinned using an old token) still displayed in their web3storage account.

### Scope of this ticket:

* Find a way to display all pinned CIDs a user owns, no matter which token was used.
* Avoid adding user management to the pinning services API.
* We do not need to identify which token was used for each file (CID).

### Acceptance criteria:

* When a user logs in to the account section of web3storage, they find pinned CIDs/files that were pinned through every token they use, not just the first token.
* The list of files needs to be sorted just like uploads, there is no distinction between tokens or sorting/identifying by token,

joshghent:
Resolves #1616 and #1507
This covers the backend portion of those tickets.
Work will still need to be done to switch the frontend to use this new endpoint. Happy to include that change in this PR but felt it might bloat this PR.

The current Pinning API is scoped by token because we do not want to pollute the entity with the concept of "users" from web3.storage.
Therefore, this PR adds a new `/user/pins` endpoint. It returns the same data format as the PSA `/pins` but returns all records for a user regardless of their authentication key used to pin the file.

Begins to standardise the logic for pagination of records in the API.
Query params can be passed - `sortBy`, `sortOrder`, `page`, `size`, `before`, `after` and `offset`.


packages/api/src/user.js
```diff
@@ -347,6 +300,72 @@ export async function userUploadsRename (request, env) {
...
+  if (psaParams.error) {
+    throw psaParams.error
+  }
+
+  const tokens = (await env.db.listKeys(request.auth.user._id)).map((key) => key._id)
+
+  let pinRequests
+
+  try {
+    pinRequests = await env.db.listPsaPinRequests(tokens, {
```
adamalton, line 329: Would it make more sense here to update the `listPsaPinRequests` method to optionally take a user ID instead of a list of tokens, and have it do the JOIN in the DB, rather than doing two separate queries in sequence?

adamalton (approved):
This looks good to me. I don't feel strongly about my comment re. avoiding two separate queries, so will leave that decision to you.

packages/api/src/user.js
joshghent: Yeah there is definitely an argument for that here.
The reason I chose against that approach is because
A) we can reuse exactly the same query (and therefore method) as the Pinning Service API by just passing tokens
B) we don't want to give the PSA awareness of user management

But you are absolutely right, this is more computationally expensive than doing it in the DB. It just seemed the cleanest approach at the time. I'll take your lead on what you reckon.

joshJarr (approved):
LGTM, nice work!

Issue #1684, opened 2022-06-08
Migrate IPNS records from existing database to durable objects


mbommerez:
### Context:

In migrating to w3name, current IPNS requests stored in Web3.storage must be relocated, so users don't need to do anything.

### Scope of this ticket:

* Migrate IPNS records in the Web3.storage database to w3name Durable Objects

### Implementation details:

Existing records: 326

Some ideas:

- Export records as a JSON file
- Create a worker that connects to Web3.storage Postgrest endpoint and fetch records

### Acceptance criteria:

- [ ] As a Web3.storage, I don't have to do anything to migrate to w3name
- [ ] All the IPNS records in Web3.storage are transferred to w3name

PR #1719, opened 2022-08-09
feat: drop name table & IPNS migration cron

### LINKED ISSUE #1704: Drop name db table and remove the data migration CRON job
Tasks
- [ ] Drop `name` db table
- [ ] Remove the running CRON job (ticket #1695) that migrates data to Durable Objects

Note
* The first task is already done - code is in #1686 - it needs splitting out into a new PR. And #1686 needs to be binned or reduced to not include the table drop.

adamalton:
Fixes #1704.

🚨 Requires a DB migration 🚨

⚠️ We must make sure that the cron job from #1684 is successfully running (and [this fix](https://github.com/web3-storage/web3.storage/pull/1759) deployed) before we release this. (Because this PR deletes the name records from the DB, so they must be migrated first).

I would actually be in favour of just leaving this for a week or so so that we're 100% sure this data isn't used or needed before we irreversibly delete it. Alternatively we could split this again and have the migration completely separate, just doing the code changes for now.

* Drops `name` Postgres table and associated functions.
* Removes the cron job that migrates IPNS records from the Postgres DB to the new `w3name` service.


francois-potato (approved):

francois-potato:
The IPNS records were migrated to Cloudflare (see "Cron migrate IPNS records to w3name" job) ✅

PR #1734, opened 2022-08-12
fix: run tests in jobs on package-lock change

vasco-santos:
We already had regressions in the past because package lock was updated and it did not trigger tests of all the packages. This also brings other issues as workflows are not triggered like https://github.com/web3-storage/web3.storage/commit/000e4dcd9e1119361c29c0dad851a5ea6d6db1b5

### DIFF
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -7,11 +7,13 @@ on:
       - 'packages/api/**'
       - 'packages/db/**'
       - '.github/workflows/api.yml'
+      - 'package-lock.json'
   pull_request:
     paths:
       - 'packages/api/**'
       - 'packages/db/**'
       - '.github/workflows/api.yml'
+      - 'package-lock.json'
 jobs:
   test:
     runs-on: ubuntu-latest
--- a/.github/workflows/client.yml
+++ b/.github/workflows/client.yml
@@ -6,12 +6,14 @@ on:
     paths:
       - 'packages/client/**'
       - '.github/workflows/client.yml'
+      - 'package-lock.json'
   pull_request:
     branches:
       - main
     paths:
       - 'packages/client/**'
       - '.github/workflows/client.yml'
+      - 'package-lock.json'
 jobs:
   test:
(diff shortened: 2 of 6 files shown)


olizilla:
hmm, in an ideal world we'd only trigger the build on a change to the relevant package.json and we just wouldn't let folks merge PRs with changed a package-lock.json that didn't have a change to the package.json. that way we could limit the CI builds to just the project that changed... this change means any package.json that changes means **everything** gets re-built

vasco-santos:
>  this change means any package.json that changes means everything gets re-built

And that is helpful in my opinion. Two points:

1. There are global packages in main `package.json`
2. Removing/upgrading a dependency from a package.json within one of the packages will change `package-lock` and might have side effects on the other packages. 

We already had 2nd point happening in nft.storage a few times https://github.com/nftstorage/nft.storage/pull/1839 and breaking CI for packages blindly. Then we need to look for a needle in a haystack as we don't know what changed. More than a mess, this takes engineering time that is more useful in others places.

olizilla:
in https://github.com/web3-storage/web3.storage/pull/1733 it seems that `wrangler` should be removed from the  devDependencies of the root package.json and added to the package.json of the api module.

I'm ok with trigger a re-build on all projects if the root `package.json` changes, but I don't want to re-build everything if _any_ workspace module changes a dep.

vasco-santos:
> in https://github.com/web3-storage/web3.storage/pull/1733 it seems that wrangler should be removed from the devDependencies of the root package.json and added to the package.json of the api module.

it is also used in the website package

olizilla:
ok, sure, I hear you. I can live with this if it's gonna save us from pain.

> it is also used in the website package
**pedantic mode** then it should be a devDep of both the website module and the api module. There is no root npm script or code that needs `wrangler`. A separate issue, but if we had put that dep in the right modules, this PR wouldn't have been needed.

olizilla (approved):
I concede to @vasco-santos's wisdom of _better to run the builds than be surprised._

PR #1735, opened 2022-08-12
chore(docs): add curl CAR upload example to store howto

yusefnapora:
@ribasushi reminded me that we could really use an example of POST-ing a CAR directly to the API.

This adds a new tab to the "Storing IPFS Content Archives" section of the Store how-to with a few examples, and a callout to use `--data-binary` instead of `-d`.

### DIFF
--- a/packages/website/pages/docs/how-tos/store.md
+++ b/packages/website/pages/docs/how-tos/store.md
@@ -269,6 +269,38 @@ cid, err := client.PutCar(context.Background(), carfile)
 ```
 
 </TabItem>
+<TabItem value="curl" label="HTTP API (curl)">
+
+If you're using the [HTTP API][reference-http] directly, you can upload a CAR file by sending a `POST` request to the `/car` endpoint.
+
+Your request must include an `Authorization` header with a value of `Bearer $API_TOKEN`, where `$API_TOKEN` is a valid JWT token for the Web3.Storage service. 
+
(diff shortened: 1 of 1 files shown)


packages/website/pages/docs/how-tos/store.md
```diff
@@ -269,6 +269,38 @@ cid, err := client.PutCar(context.Background(), carfile)
 ```
 
 </TabItem>
+<TabItem value="curl" label="HTTP API (curl)">
+
+If you're using the [HTTP API][reference-http] directly, you can upload a CAR file by sending a `POST` request to the `/car` endpoint.
+
+Your request must include an `Authorization` header with a value of `Bearer $API_TOKEN`, where `$API_TOKEN` is a valid JWT token for the Web3.Storage service. 
+
+The body of the request should be the binary CAR data, with no additional padding or encoding. Please note that the maximum size for a single CAR upload is 100 MiB. See [Working with CAR files][howto-car-files] for information about splitting CARs into smaller chunks for upload.
```
ribasushi, line 278: Love this!

Is it actually 100MiB exactly ( `100 << 20` ) or is it a bit less? Since this is for developers, probably worth validating the value so folks do not get surprises...?

yusefnapora: That's a good call out - I know that we were having some intermittent issues with uploads that were near the limit (like 90 MiB), but that was before the new upload pipeline went live. I'll throw some test uploads at it and revise this if I hit a "soft limit".

orvn (approved):
lgtm! Tested the docs (`/docs/store`) page for functionality on desktop and mobile.

@yusefnapora, I'll leave to you to merge :)

ribasushi:
@orvn @yusefnapora can haz merge...?

PR #1700, opened 2022-08-03
feat: reusable and accessible table component

flea89:
Resolves #1711 
Relates to bigger ticket #1362 
In this PR we created a Table component that is  accessible and reusable

PR #1699, opened 2022-08-03
feat: upload server side pagination

flea89:
Resolves #1362

PR #1739, opened 2022-08-15
fix: upload list pagination headers

### LINKED ISSUE #1362: File load & pagination issues for user with many files

### LINKED ISSUE #1501: Update frontend to use pagination from the apis
BLOCKED until api is completed via #1507 and #1362

## Tasks
- [ ] adjust accounts page in website to use the updated uploads api and pinning api to retrieve needed data for page. 
- [ ] implement pagination for the accounts page list of user uploads
- [x] eliminate the "sort by size" from the list of UI options to sort files (since that will not be supported by updated pagnated api. 
- [x] Disable the file search bar feature, since no way currently to search with new pagination ( this ticket was removed from scope #1509). This feature will return with new uploads v2.  
- [x] Highlight that search is deactivated, but being upgraded.  Possibly fade out the bar visually to indicate deactivation. 
 also need to put some message there about "Upgrade to search underway" perhaps in the placeholder copy, maybe also with construction icons. 
### Rough mock showing the bar with updated placeholder copy
Screen Shot 2022-06-24 at 9 39 53 AM

## Original Spec (some elements may be inaccurate, like size sorting)
With this PR: https://github.com/web3-storage/web3.storage/pull/1408 pagination is added to the `GET /user/uploads` endpoint. Using additional `size` and `page` params, uploads can be fetched in batches with the pagination metadata being returned by the response header.

The frontend should use these new parameters to batch user uploads into pages to improve the efficiency of calls for users with a large number of files.

### LINKED ISSUE #1610: Getting that error when "sudo w3 list"
Error: 400 Bad Request - invalid before date
    at Function.list (file:///usr/local/lib/node_modules/@web3-storage/w3/node_modules/web3.storage/src/lib.js:286:15)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async list (file:///usr/local/lib/node_modules/@web3-storage/w3/index.js:81:20)
And I can't load many previously uploaded files/directories...well I do admit I use A LOT the web interface to upload MANY individual files so my account has 10k++++ CIDs ;P What can I do? Should I request your pinning service? Thank you!

### LINKED ISSUE #1737: Include query string parameters in the pagination Link URLs
We should include _all_ querystring parameters in the pagination response header `Link` URLs for the list Upload and Pin endpoints otherwise clients will lose context when following the `Link` header for pagination.

e.g. setting `sortBy` for the first request, then using the URL in `Link` to get the next page.

_Originally posted by @alanshaw in https://github.com/web3-storage/web3.storage/pull/1408#discussion_r945588182_

alanshaw:
1. Fixes `Link` headers to include required pagination variables.
1. Adds "first" and "last" URLs to link header so pagination can operate properly.
1. Catches DB range error and re-throws it with the appropriate HTTP status 416.
1. Adds indexes on the uploads table for list items.
1. Fixes website to allow pagination above 1,000. Only loads the current page instead of everything.
1. Removes the search box - it only filters the current page so is almost useless. Needs proper integration with server side API.

Visual preview:

Screenshot 2022-08-17 at 17 32 36

Note new "Request ID" column for pinned items:

Screenshot 2022-08-17 at 17 32 29

resolves #1610
resolves #1501
resolves #1362
resolves #1737


joshJarr:
Nice! Code looks good and the BE changes are ace, I found a few issues with the FE code: 

1. I can not unselect uploads if I navigate to a new page, and upon navigating back the upload does not appear selected, but If I click delete, it's removed.
**Steps to replicate:**
    - On the uploads table the delete button is disabled until something is selected
    - Select a (or many) uploads (checkbox), the delete button becomes enabled
    - Navigate to the next page, the delete button remains enabled
    - Navigate back to the previous page, the delete button is enabled but the upload is not visually selected.
    
    **Expected:** 
    - Either selection is wiped upon navigating to a new page, or the selection is persisted when navigating back
    - If selection is wiped, the delete button should become disabled.
    
    **Actual**
    - Selection is not visually persisted on the table
    - Delete button is active, and clicking will delete the previously selected item.
    
    
    Aug-16-2022 10-50-40 (can't upload MP4, give me a shout if you want a higher res or a quick demo!)
    
We could just wipe the selection upon navigating away.
  
  ----
    
2. Total amounts of uploads/pins are incorrect on the tabs, only the selected result length is shown.
A quick win here could be as simple as removing the total number?
     
Screenshot 2022-08-16 at 11 17 41

----

3. Pins and Uploads are broken and appear on the same table with table columns also breaking.
    - Hopping between the two tables I find they sometimes stack together resulting in the screenshot below
Screenshot 2022-08-16 at 11 19 47

There's a few more minor issues here:
    - Sorting is also persisted through the upload & pin tabs.
    - When I hit the http://localhost:4000/account/?table=pinned URL I am directed to the uploads table no pin, despite having  pings

I think the table and pagination components are tightly coupled and need a bit of refactoring in order to separate pins and uploads, but reuse the same components. @flea89 has worked on decoupling them from the table here: https://github.com/web3-storage/web3.storage/pull/1700 and in https://github.com/web3-storage/web3.storage/pull/1699/files

I think the FE code here may require a bit more work, but the BE aspect is perfect! Would it make sense to split this into two PRs so that the Link UR, error handling changes & indexes aren't blocked by the UI?

 Thanks!

packages/website/components/account/filesManager/filesManager.js
```diff
@@ -103,12 +104,10 @@ const FilesManager = ({ className, content, onFileUpload }) => {
...
   }, [query, currentTab, pinned, replace]);
 
-  // Initial fetch on component load
+  // Re-fetch uploads when props change
   useEffect(() => {
-    if (!fetchDate && !isFetchingUploads) {
-      getUploads();
-    }
-  }, [fetchDate, getUploads, isFetchingUploads]);
+    getUploads({ size: itemsPerPage, page, sortBy, sortOrder });
```
joshJarr, line 109: Might be good to reset the `selectedFiles` here too, so that hidden selections aren't stored in that state.

packages/db/index.js
```diff
@@ -503,41 +504,49 @@ export class DBClient {
...
-    // Apply filtering
-    if (opts.before) {
-      query = query.lt('inserted_at', opts.before)
-    }
-
-    if (opts.after) {
-      query = query.gte('inserted_at', opts.after)
+  async listUploads (userId, pageRequest) {
+    let query
+    if (pageRequest.type === DATE_TIME_PAGE_REQUEST) {
```
joshJarr, line 512: I find the Date time pagination a little tricky,
Say I had two or more records with the exact same date time, wouldn't this break pagination & it would miss some results? 
Super edge-casey, but am keen to understand why filtering by datetime is necessary, is it purely for the existing UI?

alanshaw: It's required by the pinning service API https://ipfs.github.io/pinning-services-api-spec/#section/Pagination-and-filtering and that's what was originally implemented.

packages/website/components/account/fileUploader/fileUploader.js
```diff
@@ -109,7 +109,6 @@ const FileUploader = ({ className = '', content, uploadModalState, background })
             }}
             icon={<FolderIcon />}
             dragAreaText={content.drop_prompt}
-            maxFiles={3}
```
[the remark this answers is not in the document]
joshJarr: Yep, that's a great point!

packages/db/index.js
joshJarr: Ahh ok, in which case should this also be replicated in `listPsaPinRequests`?

alanshaw: `listPsaPinRequests` serves the PSA (`GET /pins`) primarily. It's now date ordered explicitly (and you can't change the sort field or direction from the UI). So, I've left it mostly as is and dealt with the "page number pagination" at the API handler level instead. If we want to enable that UX for the PSA table then yes we can probably refactor to use this pattern also.

packages/api/src/user.js
```diff
@@ -263,38 +264,86 @@ export async function userUploadsGet (request, env) {
...
-    before,
-    after,
-    sortBy,
-    sortOrder
-  })
+  let data
+  try {
+    data = await env.db.listUploads(request.auth.user._id, pageRequest)
+  } catch (err) {
+    if (err.code === 'RANGE_NOT_SATISFIABLE_ERROR_DB') {
```
francois-potato, line 273: Maybe reuse RangeNotSatisfiableDBError.CODE here

packages/api/src/user.js
```diff
@@ -357,36 +406,39 @@ export async function userPinsGet (request, env) {
...
-      sortBy,
-      sortOrder
+      limit: pageRequest.size,
+      offset: pageRequest.size * (pageRequest.page - 1)
     })
-  } catch (e) {
-    console.error(e)
-    throw new HTTPError('No pinning resources found for user', 404)
+  } catch (err) {
+    if (err.code === 'RANGE_NOT_SATISFIABLE_ERROR_DB') {
```
francois-potato, line 413: Maybe reuse RangeNotSatisfiableDBError.CODE here

packages/website/lib/api.js
```diff
@@ -267,3 +274,26 @@ export async function listPins(status, token) {
...
 }
+
+/**
+ * Deletes a pin request.
+ *
+ * @param {string} requestid
+ */
+export async function deletePinRequest(requestid) {
+  const tokens = await getTokens();
+  if (!tokens[0]) {
```
francois-potato, line 285: A bit confusing; could we use `tokens.length > 0`?

joshJarr (approved):

joshJarr:
LGTM now conflicts are resolved, lots of changes here so would be great to get a second approval, @orvn @olizilla @vasco-santos are either you guys available to take a look?
@e-schneid would be good if you could take a quick look too, just to ensure I've not caused havoc merging the upload table changes.
Thanks!

packages/api/src/user.js
joshJarr: Great point! I could make a constant in the API package so that there's less dependancy on the `db` package, or import it over from the db package. am struggling to weight up which would be best!

packages/website/lib/api.js
joshJarr: Possibly, but with this we're checking if `tokens[0]` is truthy, as it could be null. 
Tokens could be `[null, null]` and .length would be 2

joshJarr:
Merging this since it resolves a fair few issues that are currently live, but if anyone else has any feedback feel free to leave it and will address in a separate PR.
Thanks!

alanshaw:
@joshJarr thanks for finishing this!

Did the indexes get applied in prod? I just realised I should have written a migration for applying them.

ref https://github.com/web3-storage/web3.storage/pull/1739/files#diff-c387c1cfe9b20b289a2e4b668a020d576f0c3acda4c9379bef0a7c04718ecc11

PR #1754, opened 2022-08-16
test: e2e test website accountPayment loggedIn, includes magic.link testMode support in test-e2e job only

gobengo:
Motivation
* e2e test the logged in flow for /account/payment ui
* we hadn't previously found a way to e2e test auth
* playwright.dev's html reporter is awesome and lets you see console, network, screenshots of the whole e2e test! [preview](https://bafybeigc3p3rdlwq7c7leibcfjio7flmksp63mn5zioj5cmor7hrifv3dy.ipfs.nftstorage.link/trace/index.html?trace=https://bafybeigc3p3rdlwq7c7leibcfjio7flmksp63mn5zioj5cmor7hrifv3dy.ipfs.nftstorage.link/data/099727deb29243bd96b85a12fe331c1078f6e2dc.zip)
  * [more about this playwright.dev feature on debbie.codes](https://debbie.codes/blog/reports-tracing-tests-in-playwright/) 
  * this can really speed up the amount of time understanding e2e test failures!

What
* add [`playwright-report` (html)](https://playwright.dev/docs/test-reporters#html-reporter) to ci artifacts. this is useful to debug failed e2e test runs (with screenshots, console, network recorded!)
* add ability to opt-in to using [magic.link testMode](https://magic.link/docs/auth/introduction/test-mode) (not on by default, on in ci test-e2e)
* test-e2e run has `api` running and `db` with loaded schema
* test logged-in case of accountPayment


.github/workflows/website.yml
```diff
@@ -120,6 +145,7 @@ jobs:
         id: finder
       - name: Update Preview URLs PR Comment
         uses: marocchino/sticky-pull-request-comment@v2.2.0
+        continue-on-error: true
```
gobengo, line 148: I was hitting rate limits failures on this

gobengo: ... and a bunch of other actions. should i check this in?

gobengo:
The more reviews the merrier, since this touches authz checks. Seems low risk to me. This adds support for [magic.link testMode](https://magic.link/docs/auth/introduction/test-mode) but only when `env.NEXT_PUBLIC_MAGIC_TESTMODE_ENABLED`

.github/workflows/website.yml
olizilla: you were seeing workflows fail due to hitting a rate limit during the `sticky-pull-request-comment` step?

.github/workflows/website.yml
```diff
@@ -48,30 +68,77 @@ jobs:
       - uses: bahmutov/npm-install@v1
         env:
           PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: 1
-      - run: npm run build -w packages/website
+      - run: touch .env
```
olizilla, line 71: I think we can drop this line now we're creating the .env on the next line. The file doesn't need to exist to be able tot direct output to it.
```suggestion
```

.github/workflows/website.yml
```diff
@@ -48,30 +68,77 @@ jobs:
...
+      - run: |
+          (
+            npm -w '@web3-storage/website' run ci:envfile --silent \
+            | sed "s/^NEXT_PUBLIC_MAGIC=$/NEXT_PUBLIC_MAGIC=$NEXT_PUBLIC_MAGIC/" \
+            | sed "s/^MAGIC_SECRET_KEY=$/MAGIC_SECRET_KEY=$MAGIC_SECRET_KEY/" \
+            ;
+            echo 'NEXT_PUBLIC_MAGIC_TESTMODE_ENABLED="1"';
+          ) > .env
```
olizilla, line 79: Can you give me some more context on what's going on here? In general, we should be able to just set the ENV in CI, rather than being forced to write things to a .env file... those things are supposed to be a local dev convenience, rather than a sed inducing beast.

.github/workflows/website.yml
```diff
@@ -48,30 +68,77 @@ jobs:
...
         env:
-          NEXT_PUBLIC_API: http://localhost:8787
-          NEXT_PUBLIC_ENV: test-e2e
           NEXT_PUBLIC_MAGIC: ${{ secrets.STAGING_MAGIC_PUBLIC_KEY }}
+          MAGIC_SECRET_KEY: ${{ secrets.STAGING_MAGIC_SECRET_KEY }}
+      - name: load db schema
+        run: |
+          npm -w '@web3-storage/db' run start
+          npm -w '@web3-storage/db' run load-schema
+          npm -w '@web3-storage/db' run stop
```
olizilla, line 87: nit: elsewhere in our workflows we're using the workspace path like:
```suggestion
          npm run start -w packages/db
          npm run load-schema -w packages/db
          npm run stop -w packages/db
```

.github/workflows/website.yml
```diff
@@ -48,30 +68,77 @@ jobs:
...
+      - run: node -r dotenv/config $(which npm) run build -w packages/website
```
olizilla, line 88: What happens if we specify the `env` on this step? That should work. If it doesn't we should fix that.
```suggestion
      - run: npm run build -w packages/website
        env:
          NEXT_PUBLIC_MAGIC: ...
```

.github/workflows/website.yml
```diff
@@ -48,30 +68,77 @@ jobs:
...
+      - id: website_test_e2e
+        run: xvfb-run --auto-servernum --server-args="-screen 0 1280x960x24" -- node -r dotenv/config $(which npm) run -w @web3-storage/website test:e2e -- --project=${{ matrix.browser }} --reporter=html
         env:
-          NEXT_PUBLIC_API: http://localhost:8787
-          NEXT_PUBLIC_ENV: dev
-          NEXT_PUBLIC_MAGIC: ${{ secrets.STAGING_MAGIC_PUBLIC_KEY }}
-      - name: Check for website test results
+          PLAYWRIGHT_TIMEOUT: 120000
+
+      - if: always()
```
olizilla, line 96: I had to look this up! Worth adding a comment here.

.github/workflows/website.yml
```diff
@@ -48,30 +68,77 @@ jobs:
...
+      - name: add playwright-report to web3.storage
+        if: inputs.add_playwright_report_to_web3_storage && steps.check_website_playwright_report.outputs.files_exists == 'true'
+        uses: web3-storage/add-to-web3@v2
+        id: ipfs
+        with:
+          path_to_add: packages/website/playwright-report
+          web3_token: ${{ secrets.WEB3_TOKEN }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: log playwright-report on web3.storage cid, url
```
olizilla, line 112: nit: _sometimes blank line preceding step, sometimes not_

.github/workflows/website.yml
```diff
@@ -48,30 +68,77 @@ jobs:
...
+        if: inputs.add_playwright_report_to_web3_storage && steps.check_website_playwright_report.outputs.files_exists == 'true'
+        run: |
+          echo cid: ${{ steps.ipfs.outputs.cid }}
+          echo url: https://nftstorage.link/ipfs/${{ steps.ipfs.outputs.cid }}
```
olizilla, line 116: even newer! even shinier!
```suggestion
          echo url: https://w3s.link/ipfs/${{ steps.ipfs.outputs.cid }}
```

.github/workflows/website.yml
```diff
@@ -48,30 +68,77 @@ jobs:
...
+      - id: create_output_playwright_reporter_url
```
olizilla, line 117: are these outputs used anywhere? If not let's delete them. If they are, then let's also reuse them above where we echo out the url and cid.

.github/workflows/website.yml
```diff
@@ -48,30 +68,77 @@ jobs:
...
+        if: inputs.add_playwright_report_to_web3_storage && steps.check_website_playwright_report.outputs.files_exists == 'true'
+        run: echo "::set-output name=url::https://nftstorage.link/ipfs/${{ steps.ipfs.outputs.cid }}"
+      - id: create_output_playwright_reporter_cid
+        if: inputs.add_playwright_report_to_web3_storage && steps.check_website_playwright_report.outputs.files_exists == 'true'
+        run: echo "::set-output name=cid::${{ steps.ipfs.outputs.cid }}"
+
+      - name: upload playwright-report artifact
```
olizilla, line 124: I think you could drop a whole bunch of `if` logic from the steps above by having the e2e test run job only do the uploading of the playwright-report (where it exists)... you can a dependent job to publish the report and have that fetch the artefact from GitHub, and skip the whole job if there is no report to publish

olizilla (commented):
I'm mid review but sharing thoughts on the workflow part early while I go through the auth part

packages/api/src/utils/env.js
```diff
@@ -0,0 +1,3 @@
+export function magicTestModeFromEnv (env) {
+  return Boolean(env?.NEXT_PUBLIC_MAGIC_TESTMODE_ENABLED)
```
olizilla, line 2: **danger danger**
```node
> Boolean("false")
true
```

packages/website/lib/constants.js
```diff
@@ -1,16 +1,18 @@
-const API = /** @type {string} **/ (process.env.NEXT_PUBLIC_API)
-const MAGIC_TOKEN = /** @type {string} **/ (process.env.NEXT_PUBLIC_MAGIC)
+const API = /** @type {string} **/ (process.env.NEXT_PUBLIC_API);
```
olizilla, line 1: no-semis plz. always double check PR diffs for meaningless changes, to spare your reviewers eyeballz.

packages/website/lib/magic.js
```diff
@@ -1,20 +1,21 @@
-import { Magic } from 'magic-sdk'
-import { OAuthExtension } from '@magic-ext/oauth'
-import constants from './constants'
+import { Magic } from 'magic-sdk';
```
olizilla, line 1: so many semis

packages/api/src/auth.js
```diff
@@ -145,14 +146,34 @@ export function withPinningAuthorized (handler) {
...
+    switch (error.code) {
+      case 'ERROR_INCORRECT_SIGNER_ADDR':
+      case 'ERROR_MALFORMED_TOKEN':
+        break
+      default:
+        console.warn('unexpected error validating magic token: ', error.name, error.message)
+    }
+  }
+  const isAllowedTest = (env, token) => {
+    return Boolean(env.DANGEROUSLY_BYPASS_MAGIC_AUTH && token === 'test-magic')
```
olizilla, line 165: historical note: `DANGEROUSLY_BYPASS_MAGIC_AUTH` was [introduced as a workaround](https://github.com/web3-storage/web3.storage/pull/237#issuecomment-889772445) until magic-sdk landed support for test-mode. We can get rid of it now test-mode is a thing.

olizilla: I don't think we need both `NEXT_PUBLIC_MAGIC_TESTMODE_ENABLED` and `DANGEROUSLY_BYPASS_MAGIC_AUTH`. One way to bypass auth is one too many but a necessary horror.

packages/api/src/user.js
```diff
@@ -39,9 +40,33 @@ async function loginOrRegister (request, env) {
...
   const token = env.magic.utils.parseAuthorizationHeader(auth)
-  env.magic.token.validate(token)
-
-  const metadata = await env.magic.users.getMetadataByToken(token)
+  const isMagicTestMode = magicTestModeFromEnv(env)
+  let isAllowedTestmodeAuthn = false
+  try {
+    env.magic.token.validate(token)
+  } catch (error) {
+    if (error.code === 'ERROR_INCORRECT_SIGNER_ADDR' && isMagicTestMode) {
```
olizilla, line 48: Would it be possible to avoid leaking awareness of magic test mode here? I think we are just supposed to set it as an option on when creating the sdk and then control if it errors or not by what email we use in test code per https://magic.link/docs/auth/introduction/test-mode

.github/workflows/website.yml
gobengo: @olizilla Yeah, it was a several different steps, but that was one of them. I think it rate limiting was coming from github apis

.github/workflows/website.yml
gobengo: > Can you give me some more context on what's going on here?

def

> In general, we should be able to just set the ENV in CI, rather than being forced to write things to a .env file

oh yeah. I totally I tried that... like a lot. I wanted to avoid writing the MAGIC_SECRET_KEY to the fs and just have it in the env...

This is the only thing I could get to work. It has to do with `dotenv`... which I rarely use so perhaps I'm missing something. 

The other thing is how api uses miniflare here. https://github.com/web3-storage/web3.storage/blob/main/packages/api/package.json#L30
* See how miniflare uses `--env ../../.env`. In my testing, it will *only* use that file, and will not inherit other env vars. The only way I could get the API to have these env vars set at runtime was to write them into the `.env` file, and it has to do with how the api package starts with `miniflare` cli

With all that in mind... is there another way you'd want me to do it? The only other thing I could think of involved meddling with the `miniflare` npm script in api package, e.g. using the 'binding' flag. but this seemed less gross

Whether or not we make a bigger change in how to provide these values dynamically like this, I will add a comment to this part of the workflow file

gobengo: btw I could also get rid of the use of `sed` by putting that find/replace logic in `npm -w '@web3-storage/website' run ci:envfile` and having that script read from envs or its own flags, which seems like a nice way of encapsulating this in a node script with no `sed`

.github/workflows/website.yml
gobengo: ok I find it less readable because 1) it helps me to first know what package the thing will run in and 2) the '-w' line up better in the text editor if they go first. however, I will use the suggested format for the sake of consistency

.github/workflows/website.yml
gobengo: it wont work! lol 🤣 

So this is another quirk of how our npm scripts sorta rely on dotenv.

`npm run build -w packages/website` will run the build script in that package, which will use dotenv. BUT. dotenv will read from `.env` within the package website, not within the monorepo root.

This workflow only writes it to the monorepo root. And then we want those envs to be used by more than just the website package (e.g. also api, db)

.github/workflows/website.yml
gobengo: done!

.github/workflows/website.yml
gobengo: :) I did it on purpose as a way of separating groups of related steps.

Would you prefer I remove them?

.github/workflows/website.yml
gobengo: @olizilla Good call. Honestly it was an experiment. It felt useful to 'output' them. However, I looked around and it doesn't seem like outputs get shown in the github ui or anything :/

packages/api/src/utils/env.js
gobengo: @olizilla nice. I actually thought of this too when I authored it :) Usually what I do to accomodate for this is `JSON.parse` the string, which will parse this case to `false`. I'll do that and add a test case

.github/workflows/website.yml
gobengo: it takes so long to get to this part of the build so its time-expensive to iterate on refactors like that :/

gobengo: lookin forward to future of tooling like this https://dagger.io/

packages/website/lib/constants.js
gobengo: https://github.com/web3-storage/web3.storage/issues/1762

packages/website/lib/magic.js
gobengo: im removin them and will later take on https://github.com/web3-storage/web3.storage/issues/1762

packages/api/src/auth.js
gobengo: @olizilla I noticed that `DANGEROUSLY_BYPASS_MAGIC_AUTH` was there and mentioned that testmode wasn't out yet when it was authored.

However, I think actually these are worth being separate things... now that testMode exists and we can learn its limitations.

Specifically, the testMode stuff seems like it was released to be mostly about the *frontend* state managed by the magic sdk. e.g. it has some localstorage and stuff for its session. So when we're using it in the e2e tests, we need to construct `Magic` with `options.testMode === true` and that is what `NEXT_PUBLIC_MAGIC_TESTMODE_ENABLED` is for. Notably, that all works regardless of the value of `DANGEROUSLY_BYPASS_MAGIC_AUTH`. And when test-e2e uses those code paths, the magic tokens are syntactically correct magic tokens (iirc they are base64-encoded JSON strings).

But there is still some use for `DANGEROUSLY_BYPASS_MAGIC_AUTH`. When this is enabled, and only when this is enabled, the 'tokens' for magic auth are allowed to be 'test-magic'. Quite a few old unit tests use this token value and the tests only pass because `DANGEROUSLY_BYPASS_MAGIC_AUTH` is truthy.

So... I think both are useful and slightly different. There's not an easy way to remove `DANGEROUSLY_BYPASS_MAGIC_AUTH` without refactoring quite a few old tests. We can do that... but plz lets consider that as a separate issue.

.github/workflows/website.yml
gobengo: > If they are, then let's also reuse them above where we echo out the url and cid.

I updated to do this. Now the only place the url is 'formatted' is when setting the output. This feels better

.github/workflows/website.yml
gobengo: I cleaned up the uses of `if`

I also evaluated deferring the playwright-report storage to another job. Because test-e2e uses a matrix to test across more than one browser, so would any other job, which I think would make the code a bit more complex (keeping 'matrix' in sync across several jobs).

packages/api/src/user.js
gobengo: I reverted just this part to verify that test-e2e can't succeed without this little change. i.e. the tokens that come from using testMode do *not* go through this `magic.token.validate()` fn without throwing.
* test run https://github.com/web3-storage/web3.storage/runs/7887390621?check_suite_focus=true#step:9:19
* playwright-report of fail: https://ipfs.io/ipfs/bafybeif3vlpvh3yqxpwttnlknnc544xluvhvivnumpzs6fqll2lf5qu2iy/trace/index.html?trace=https://ipfs.io/ipfs/bafybeif3vlpvh3yqxpwttnlknnc544xluvhvivnumpzs6fqll2lf5qu2iy/data/dc71c7740419de3f7316bba58bc3d1c6854cf39f.zip

This is in line with my intuition that they mostly shipped testMode for the *frontend* magic sdk instance and in-browser session management stuff.

gobengo: > Would it be possible to avoid leaking awareness of magic test mode here?

Not that I can tell. The test mode stuff really is a special case. *only* when its enabled should we allow an error from the call to `validate`.

I made this commit that encapsulates this logic so this requestHandler is smaller. https://github.com/web3-storage/web3.storage/commit/3505eeb4e44d58159d28ee862279fe626fbcd79f#diff-88d8b13b7d7c1bb4b032ea967283797522a6984ca7424d7ea49bc311b4c82796R84

.github/workflows/website.yml
gobengo: done

packages/api/src/utils/env.js
gobengo: I made it so providing `"false"` results in testMode being disabled

.github/workflows/website.yml
yusefnapora: @gobengo that's interesting... in https://github.com/nftstorage/nft.storage/pull/2130 I'm overriding some env variables in the current nodejs process and running the `miniflare` cli using `execa`, and miniflare does seem to use the overrides instead of what's in the .env file.

But I've also seen what you're describing with the current setup - running `yarn dev` only grabs the vars from the file... Now I'm curious what the difference is...

gobengo:
status: today I'll resolve these conflicts and early next week will finish review with oli

gobengo:
status: I resolved the conflicts with main. I think the checks would be green if not for the release-please rate limit errors.

packages/api/src/auth.js
gobengo: Since last comment, I did land this which provides a way for tests to not have to know the details of the 'test-magic' token (which only works when env.DANGEROUSLY_BYPASS_MAGIC_AUTH)
https://github.com/web3-storage/web3.storage/pull/1769

gobengo: Perhaps something more reasonable than a `DANGEROUSLY_BYPASS_MAGIC_AUTH` flag would be a new env var `WEB3_STORAGE_AUTHENTICATORS`

Most of the time
```
WEB3_STORAGE_AUTHENTICATORS=JSON.stringify({ oneOf: ['magic.link'] })
```

While testing:
```
WEB3_STORAGE_AUTHENTICATORS=JSON.stringify({
  oneOf: [
    'magic.link',
    {
      bearerTokens: ['test-magic'],
      capabilities: ['*'],
    },
  ]
})
```

At least this way it's just explicit extra configuration for 'oneOf the ways to authenticate a requestor', and not something 'dangerous' and explicitly having to do with disabling magic.link.

olizilla: The naming `DANGEROUSLY_BYPASS_MAGIC_AUTH` was deliberate! It should feel awkward and make us double check why and where we are setting it.

This proposal makes it look all sleek and like something that would be totally chill to set even in production!

.github/workflows/website.yml
olizilla: > With all that in mind... is there another way you'd want me to do it?

Let's figure out how we can set env vars for miniflare in a way that doesn't force us to not use actual ENV vars. I'm hopeful we can fix that without changing miniflare itself.

.github/workflows/website.yml
olizilla: If we can, lets fix the bug that is _(how we set env vars for building the website build)_

.github/workflows/website.yml
olizilla: Let's grab that error and track down why it's happening. I don't want the website preview url step to introduce problems.

packages/api/src/utils/env.js
olizilla: The only time test mode should be set is if the user provides the string value `"true"` so let's say it loud and clear here:
```suggestion
  return env.NEXT_PUBLIC_MAGIC_TESTMODE_ENABLED === "true"
```
the optional chaining would guard against folks using this function but passing no arguments, which feels like it would be a programming mistake that should throw rather than be dodged.

olizilla: obvs "string value true" is gross, but these are env vars. we are deep in the string mines.

packages/api/src/auth.js
```diff
@@ -146,14 +147,34 @@ export function withPinningAuthorized (handler) {
  */
 async function tryMagicToken (token, env, bypass = magicLinkBypass) {
```
gobengo, line 148: maybe `bypass` should still be `dangerouslyBypass`

gobengo: `magicLinkBypass` should be defined in this file so it's front and center

packages/api/src/utils/env.js
gobengo: I think you commented on out outdated part. I updated `magicTestModeFromEnv` to handle values like `"false"` as falsy (and values like `"1"` as truthy) via https://github.com/web3-storage/web3.storage/commit/754e9eb0b60aad56279e9d628267e823cf28bbef

packages/api/src/auth.js
gobengo: after this commit, the (now two) bypasses are all prefixed with `dangerous` https://github.com/web3-storage/web3.storage/pull/1754/commits/ded465b5f949dfad96b89890681d46bef79f613a#diff-d8c9d6f9ae3c9c310454275870e4b2b397b36411752f664a0540d301027bbc1dR157

gobengo:
I (think I) cleaned up the api/src/auth code a bit. One bit that came from walkthrough with @olizilla and @hugomrdias this morning is realizing that our magic.link token check actually has two opportunities to be bypassed, which are now made explicit as `dangerousAuthBypassForUnitTesting` and `dangerousAuthBypassForE2eTesting`. The former encapsulates allowing `token='test-magic'` for our api unit tests (which needs to lead to a certain corresponding `issuer`). The latter encapsulates allowing magic testMode tokens, which are more syntactically correct than the unit test tokens in that they include their own `issuer` value, but still need to be blocked in prod and allowed only when testMode is enabled. Also included in this new commit is explicit detection of the 'test mode' tokens by parsing them and asserting at their `aud` and `iss` claims. Being able to to check for that explicitly and dispatch to `dangerousAuthBypassForE2eTesting` logic helped make the rest of the validation simpler than it had been.
https://github.com/web3-storage/web3.storage/commit/ded465b5f949dfad96b89890681d46bef79f613a

.github/workflows/website.yml
gobengo: @olizilla I found a way: https://github.com/web3-storage/web3.storage/pull/1781

.github/workflows/website.yml
gobengo: I reverted this change. I don't think this PR introduced a root cause of why that happened. AFAICT it happens because we have enough workflow items that use `secrets.GITHUB_TOKEN` that run enough times per hour that we're getting rate limited by github's 1000req/hr default.

packages/api/src/auth.js
gobengo: @olizilla In my last proposal defining `WEB3_STORAGE_AUTHENTICATORS`, there would no longer be a `DANGEROUSLY_BYPASS_MAGIC_AUTH`, so there would be no temptation to set (that env var) in production.

.github/workflows/website.yml
gobengo: Turns out we dont need to do a (prod) build of the website here because later when playwright runs (monorepo) `npm start` , it'll run `start` for api, which will run miniflare in such a way that it triggers a build based on the env of that step. https://github.com/web3-storage/web3.storage/pull/1782#issuecomment-1223363343

packages/website/components/loading/loading.js
```diff
@@ -16,7 +16,7 @@ export const SpinnerColor = {
 };
 
 /**
- * @param {Object} [props]
```
gobengo, line 19: This was required for `next build` to build locally for me with tsc, and makes sense to clarify that the argument destructuring wont work if `props === undefined`

packages/website/lib/constants.js
```diff
@@ -1,5 +1,6 @@
 const API = /** @type {string} **/ (process.env.NEXT_PUBLIC_API);
 const MAGIC_TOKEN = /** @type {string} **/ (process.env.NEXT_PUBLIC_MAGIC);
+const MAGIC_TESTMODE_ENABLED = Boolean(process.env.NEXT_PUBLIC_MAGIC_TESTMODE_ENABLED);
```
gobengo, line 3: I think that we're relying on miniflare/wrangler to replace this `NEXT_PUBLIC_MAGIC_TESTMODE_ENABLED` is one reason you have to explicitly tell `miniflare` about your env vars either through the env file or through the `--binding KEY=VALUE` flags I added when the website ci workflow invokes `npm -w '@web3-storage/api' start`

gobengo:
the main new thing I'd point out, which I think made the code clearer, was to reify the fact that that there are two distinct kinds of of 'magic.link bypasses'
1. [one we use only in unit tests](https://github.com/web3-storage/web3.storage/pull/1754/files#diff-74d3bc0861a7bbeb228eaa72564c93b96ab019c6d5e9b58cb33be88e8bd09d2bR29), which allows a special allowlisted bearer token to work
2. [one we use only in e2e tests,](https://github.com/web3-storage/web3.storage/pull/1754/files#diff-74d3bc0861a7bbeb228eaa72564c93b96ab019c6d5e9b58cb33be88e8bd09d2bR57) which enables detection of any incoming Bearer tokens that are JWTs that look like a magic.link test mode token (it looks in jwt sub and aud), and doesnt bother to validate those tokens + ensures those tokens can be resolved to appropriate metadata for the rest of our 'user login' stuff to work
    * this bypass, when enabled, is also used in a controller used by the loginOrRegister route that handles magic.link tokens (because now those tokens can be testMode tokens but only when enabled via env var aka only in test-e2e ci)

packages/api/src/magic.link.js
```diff
@@ -1,3 +1,5 @@
+import { magicTestModeIsEnabledFromEnv, maybeJsonParseable } from './utils/env.js'
+
 export const createMagicTestTokenBypass = (
```
yusefnapora, line 3: If we can remove the `export` here and just export the concrete value `magicLinkBypassForUnitTestingWithTestToken`, that seems a bit cleaner. But if you need to export this for testing, that's okay by me...

packages/website/scripts/envfile.ts
```diff
@@ -0,0 +1,18 @@
+import * as fs from 'fs';
+import * as path from 'path';
+
+/**
+ * print a good .env file to stdout that can be used for web3.storage
```
yusefnapora, line 5:
```suggestion
 * print a good .env file to stdout that can be used for web3.storage
 * The `ENVFILE_WITHOUT` env variable may contain a comma-separated list of variable names to exclude from the output.
```

yusefnapora: Maybe also add a comment that explains why we need this for CI.

yusefnapora (approved):
I don't see anything that looks dangerous here (at least, nothing that's not lampshaded with "dangerouslyDoTheThing" naming 👍 ), and I like the explicit naming for the different types of bypass.

I'm not super crazy about the env file rewriter script... but I'm guessing that we need it in order to allow miniflare to use the bindings set in CI? That seems like a reasonable justification, IMO.

It would also be rad to add a note to the dev docs / README about the bypass modes - @gobengo maybe you could grab that as a follow up when you're back from dweb camp? I don't think we need to block on it, but it seems worth a mention. I left a couple other non-blocking suggestions if you have time to grab.

packages/api/src/magic.link.js
gobengo: https://github.com/web3-storage/web3.storage/pull/1754/commits/88a488d3affb2f90a77918b324656798932c081b

gobengo:
status: I believe checks are green except for GITHUB_TOKEN rate limits

packages/api/test/env.spec.js
```diff
@@ -0,0 +1,31 @@
...
+  }), false)
+  assert.equal(fn({
+    NEXT_PUBLIC_MAGIC_TESTMODE_ENABLED: 'true'
+  }), true)
+  assert.equal(fn({
+    NEXT_PUBLIC_MAGIC_TESTMODE_ENABLED: 'false'
+  }), false)
+  assert.equal(fn({
+    NEXT_PUBLIC_MAGIC_TESTMODE_ENABLED: 'not-json-parseable'
+  }), true)
```
olizilla, line 30: @gobengo having anything truth that isn't 'false' or '0' mean true seems like a foot gun. Please can we have exactly 1 valid way to enable test mode:
```
NEXT_PUBLIC_MAGIC_TESTMODE_ENABLED === 'true'
```

Issue #1741, opened 2022-08-15
ui: PaymentSettings component to go inside /account/payment page


gobengo:
Motivation
* create ui components we can use in payment settings page

Goal:  overall goal is to ship something that can accept an authenticated user's payment method (only CC for now).  Form can submit, but doesnt need backend connection.

Idea
* iterate in storybook until #1698 ?

gobengo:
work happening here: https://github.com/web3-storage/web3.storage/pull/1767

PR #1769, opened 2022-08-18
feat: add to api GET /user/payment, add AuthorizationTestContext to encapsulate magic.link test bypass

gobengo:
Motivation
* @cmunns is gonna use this in https://github.com/web3-storage/web3.storage/issues/1741

What
* add the route
* it always returns the same thing
  * I'll add a way of mutating it in followup PR
* test that this route requires auth
* hide 'test-magic' details from the test via new AuthorizationTestContext
  * mocha hooks installs this context
  * tests can use the context to `createUserToken()`
  * provides abstraction on how our tests get an authz token
    * followup pr can replace other repeated instances of 'test-magic' with `createUserToken()` https://github.com/web3-storage/web3.storage/search?q=test-magic

### DIFF
--- a/packages/api/src/magic.link.js
+++ b/packages/api/src/magic.link.js
@@ -1,0 +1,25 @@
+export const createMagicTestTokenBypass = (
+  requiredVariableName,
+  requiredTokenValue
+) => {
+  return {
+    requiredVariableName,
+    requiredTokenValue,
+    defaults: {
+      issuer: 'test-magic-issuer'
+    },
+    summary: [
+      'This is a bypass for testing our APIs even though most of them require a valid magic token.',
+      'When testing, we\'ll use a special-use token.',
+      'And our token-validating middleware will allow that token,',
+      `but *only* when env.${requiredVariableName} is truthy.`
+    ].join(' ')
+  }
+}
+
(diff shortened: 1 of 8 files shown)


packages/api/src/magic.link.js
```diff
@@ -0,0 +1,25 @@
+export const createMagicTestTokenBypass = (
```
yusefnapora, line 1: Just curious: what's the motivation for exporting the "create bypass" function? Will we ever need to make one elsewhere?

yusefnapora (approved):
I really like putting an abstraction around the auth bypass thing - hardcoding magic test stuff is never great. Having `bypass.defaults` is neat also :+1:

packages/api/src/magic.link.js
gobengo: @yusefnapora Mostly just abstraction. Also doing it as a function might make it easier to unit test if things get complex.

yusefnapora: ah, I missed the bit where you can override the default bypass in the AuthorizationTestContext constructor. Seems like a good idea, since it sounds like there's going to be an "official" magic link bypass & we might have other auth providers with their own bypass details

Issue #1762, opened 2022-08-17
website package lint script doesn't run against `./lib` dir because of .eslintignore


gobengo:
Motivation
* debugging https://github.com/web3-storage/web3.storage/pull/1754#discussion_r948061296

What
* the files in packages/website/lib aren't being linted right now by `npm -w '@web3-storage/website' lint` nor ci

PR #1770, opened 2022-08-18
fix: website lint script now lints its 'lib' directory after it was previously eslintignored

gobengo:
Motivation:
* https://github.com/web3-storage/web3.storage/issues/1762
* discovered this while devving on #1754


yusefnapora (approved):
Looks solid to me :+1:

PR #1774, opened 2022-08-19
test: replace constant 'test-magic' with use of AuthorizationTestContext#createUserToken()

gobengo:
Motivation
* encapsulate the details of how our tests make requests as a fake user
  * prev: lots of test just somehow know to use a 'test-magic' token
  * next: tests that want a valid user token can do `AuthorizationTestContext.use(this).createUserToken()`


yusefnapora (approved):
nice :)

PR #1782, opened 2022-08-23
ci: build website doesnt use dotenv/config

gobengo:
Note:
* non-main pr target

Motivation:
* experimenting with solution for: https://github.com/web3-storage/web3.storage/pull/1754#discussion_r951279621


gobengo:
turns out that we dont need to build it here, because the api server running via miniflare will build as part of boot

Issue #1748, opened 2022-08-15
api: create api that can save payment settings


gobengo:
## Progress

* [x] first draft of openapi spec
* [x] implement api interface in stubbed way
* [x] implement api interface with stripe.com api integration
* [x] may require @web3-storage/db schema changes to store mapping between our user and stripe records
* [x] test api edge cases
* [x] on merge to main and there is a release-please PR, add reminder comment to run migration [packages/db/postgres/migrations/023-add-user-customer-table.sql](https://github.com/web3-storage/web3.storage/pull/1818/files#diff-913f37e5c36e19554c50a0c2ef1ccf134680fbb58bc3b6583a65f5acacd64c2c)
* [x] on release to staging, run migration [packages/db/postgres/migrations/023-add-user-customer-table.sql](https://github.com/web3-storage/web3.storage/pull/1818/files#diff-913f37e5c36e19554c50a0c2ef1ccf134680fbb58bc3b6583a65f5acacd64c2c)
* [x] on release to production, run migration [packages/db/postgres/migrations/023-add-user-customer-table.sql](https://github.com/web3-storage/web3.storage/pull/1818/files#diff-913f37e5c36e19554c50a0c2ef1ccf134680fbb58bc3b6583a65f5acacd64c2c)

## Spec

Visualize with: https://editor-next.swagger.io/

```yaml
openapi: 3.0.3
info:
  title: Payment API
  version: "0.0.1"
servers:
  - url: https://staging.web3.storage/api/user
tags:
  - name: payment
    description: Pay for things
paths:
  /payment:
    get:
      tags:
        - payment
      summary: get payment settings
      operationId: getPayment
      responses:
        '200':
          description: Successful operation
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/PaymentSettings'

    put:
      summary: Update payment settings
      tags:
        - payment
      operationId: setPayment
      requestBody:
        $ref: '#/components/requestBodies/PaymentSettingsUpdate'
      responses:
        '202':
          description: Successful operation

components:
  examples:
    stripePaymentMethodId:
      value: pm_1LYC6JIfErzTm2rEeZ44St7k

  schemas:
    PaymentSettings:
      type: object
      properties:
        method:
          type: object
          properties:
            id:
              type: string
              example:
                $ref: '#/components/examples/stripePaymentMethodId/value'

  requestBodies:
    PaymentSettingsUpdate:
      content:
        application/json:
          schema:
[...]

PR #1767, opened 2022-08-18
feat: account payments UI start

cmunns:
Screen Shot 2022-08-18 at 9 26 26 AM

PR #1783, opened 2022-08-23
feat: initial PUT /user/payment api to ui dev against

gobengo:
Motivation:
* unblocking ui dev against our https api contract https://github.com/web3-storage/web3.storage/issues/1748

Note:
* this stubs the http api behavior so @cmunns can dev ui against it, but doesn't actually save to stripe.com or our db. I'll do that ASAP after that to unstub

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -4,6 +4,8 @@ on:
     branches:
       - main
     paths:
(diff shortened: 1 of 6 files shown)


gobengo:
This has been incorporated into https://github.com/web3-storage/web3.storage/pull/1767

Issue #1784, opened 2022-08-23
Only write to S3 in upload flow


olizilla:
Currently we write user uploaded CARs to elastic-ipfs then ipfs-cluster, while pinning service requests go only to ipfs-cluster. 

We need to be able to disable writing uploads to ipfs-cluster from an env var, and only write uploads to elastic-ipfs, so we can handle spikes in uploads without reducing the perf of the pinning service api.

olizilla:
We currently use the ipfs-cluster pin status to determine what we report as the status of an `Upload`. As part of this work we can include a `Pin` entry in our db for the copy of the DAG that is stored in elastic-ipfs. Our notion of a `Pin` in the db is "an IPFS node that has a copy of the data", so it seem like a good fit. By adding that record, all the existing APIs and pin status logic can remain unchanged.

PR #1785, opened 2022-08-23
feat: put write to cluster behind a flag

olizilla:
- Set `ENABLE_ADD_TO_CLUSTER` to send upload to ipfs-cluster after sending response to user. The current behavior is to always try to write user uploads to cluster. With this PR we disable that but leave a feature flag to restore it if we need to.
- Adds LinkIndexer to decode and record links from each block as we iterate through them. It allows us to set the DAG structure as 'Complete' where we know the upload contains all the blocks for a complete DAG with no dangling links.
- Adds task to call to `LINKDEX_API` after the response is sent where the DAG is `Partial`, to check if it is complete across other CARs under the same upload.
- Where we have a complete CAR we set the status as "Pinned" early as it's already in elastic-ipfs. This is debatable, as there is a lag until it's discoverable, but I think it's reasonable as I want to push us towards telling users when we safely have their stuff early.
- We write uploads to more than one place, by adding an R2Bucket binding to our worker env called `carpark` where we write cars keyed by their car CID.
- R2 only supports an [md5 integrity check on put](https://developers.cloudflare.com/r2/runtime-apis/#r2putoptions), so this PR re-uses the md5 lib that the aws-s3-sdk uses to generate that and also stores it in the r2 metadata so we see what hash was used to verify it later.
- CAR CIDs are generated as [w3up-cli does it](https://github.com/web3-storage/w3up-cli/blob/b313c1221a0fb93a7d7cc3ced5fd1d4bce537a7a/src/utils.js#L46-L50) with the [`0x202` code](https://github.com/multiformats/multicodec/blob/955e61fe6221454c682ca9ba4901558bfc1bac96/table.csv#L130), and added to the /car response body.
- miniflare is now available as a global in tests so we can make assertions about [r2 bucket contents in tests](https://miniflare.dev/storage/r2#manipulating-outside-workers), which is nice.
- we make use of [miniflare's new fetch mocking support](https://miniflare.dev/core/standards#mocking-outbound-fetch-requests) to test calls to the linkdex-api happen when we expect them to.

_NOTE: We've been using ipfs-cluster as a DAG completeness check-ing service, and we're using DAG completeness as a proxy for "we got all the CARs and the DAG is stored". We need to do some reframing as it's a feature of DAGs and CARs and IPFS that you could upload some `dab-json` that links to the complete wikipedia archive. We want to inform you when your blocks are safely stored, but we don't want to necessarily also pull in all of wikipedia, nor do we want to avoid telling you your blocks are pinned where they link to some other dataset._

Fixes: https://github.com/web3-storage/web3.storage/issues/1784

## TODO
- [x] create staging and production r2 buckets on the dag.house Cloudflare account.
- [ ] post deploy: function.sql needs to re-run on the db so we stop creating rows in `pin_sync_requests`.
- [ ] post deploy: when we've handled the existing pin_sync_requests in the pins cron, disable that job.
- [ ] post deploy: PR to remove pin sync cron job

License: MIT


olizilla:
The issue with this as it stands is we have been using ipfs-cluster as the source of truth to tell us when an entire DAG is reachable, and only set a uploads status to `pinned` if at least one cluster node reports that... which also tells us it was able to reach the entire DAG.

olizilla:
To fix this we need something to be able to tell us when we have a complete DAG stored and available. Right now cluster will do so eventually, and we have a cron job to ask cluster again later when things get too slow, and we don't get the answer within the time available in a CF worker.

olizilla:
I'm fixing up the conflicts and adding a test, but if you have a moment to review this @alanshaw / @vasco-santos that'd be rad. I'll be working on updating the pin status cron job next, based off of this branch.

packages/api/src/car.js
```diff
@@ -3,23 +3,23 @@
...
 import { CarBlockIterator } from '@ipld/car'
 import { toString, equals } from 'uint8arrays'
-import { Block } from 'multiformats/block'
 import { sha256 } from 'multiformats/hashes/sha2'
+import { LinkIndexer } from 'linkdex'
+import { maybeDecode } from 'linkdex/decode'
 import * as raw from 'multiformats/codecs/raw'
-import * as cbor from '@ipld/dag-cbor'
 import * as pb from '@ipld/dag-pb'
+import pRetry from 'p-retry'
```
alanshaw, line 11: Move from devDependencies to dependencies.

packages/api/src/car.js
```diff
@@ -3,23 +3,23 @@
...
 import { InvalidCarError } from './errors.js'
 import { MAX_BLOCK_SIZE } from './constants.js'
 import { JSONResponse } from './utils/json-response.js'
 import { getPins, PIN_OK_STATUS, waitAndUpdateOkPins } from './utils/pin.js'
 import { normalizeCid } from './utils/cid.js'
+import assert from 'assert'
```
alanshaw, line 17: Add to dependencies.

packages/api/test/hooks.js
```diff
@@ -46,7 +50,8 @@ export const mochaHooks = () => {
         wranglerConfigPath: true,
         wranglerConfigEnv: 'test',
         modules: true,
-        bindings: workerGlobals
+        bindings: workerGlobals,
+        fetchMock: fetchMock
```
alanshaw, line 54:
```suggestion
        fetchMock
```

alanshaw (approved):

packages/api/src/car.js
```diff
@@ -127,20 +134,50 @@ export async function handleCarUpload (request, env, ctx, car, uploadType = 'Car
...
     authKey: authToken?._id,
     contentCid,
     sourceCid,
     name,
     type: uploadType,
-    backupUrls: [`https://${env.s3BucketName}.s3.${env.s3BucketRegion}.amazonaws.com/${bucketKey}`],
-    pins: [],
+    backupUrls: [
+      `https://${env.s3BucketName}.s3.${env.s3BucketRegion}.amazonaws.com/${s3Key}` //,
+      // `https://${env.r2BucketName}.${env.accountId}.r2.cloudflarestorage.com/${r2bucketKey}`
```
olizilla, line 159: I'd like to stash a url for the r2 backup of the CAR here, but I'm not sure what URL to use. I'm guessing here based on some things I've seen in the r2 docs, but it doesn't look like the public url structure for r2 buckets is documented yet... also this would require us to duplicate our bucket config as env vars so we could get the "bucket  name" which is hidden from the worker normally as it just gets bound to an env var, and we'd also need an env var for the account id. Not a problem, but it makes our wrangler.toml a little gross, as the same values would need to be in two places.

vasco-santos: R2 public buckets did not land yet... So, there is really no way of getting it for now.

vasco-santos (approved):
LGTM! Great work @olizilla 

Left minor suggestions to get it to the finish line

packages/api/src/car.js
```diff
@@ -107,18 +109,23 @@ export async function carPost (request, env, ctx) {
...
   // Returns either the sum of the block sizes in the CAR, or the cumulative size of the DAG for a dag-pb root.
-  const { size: dagSize, rootCid } = await carStat(car)
-  const sourceCid = rootCid.toString()
+  const { size: dagSize, rootCid, structure } = await carStat(carBytes)
 
-  if (structure === 'Unknown' && rootCid.code === raw.code) {
-    structure = 'Complete'
-  }
+  const sourceCid = rootCid.toString()
+  const carCid = CID.createV1(0x202, await sha256.digest(carBytes))
```
alanshaw, line 121: Extract `0x202` as a `const`?

packages/api/src/car.js
```diff
@@ -107,18 +109,23 @@ export async function carPost (request, env, ctx) {
...
+  const s3Key = `raw/${sourceCid}/${user._id}/${toString(carCid.multihash.bytes, 'base32')}.car`
+  const r2Key = `${carCid.toString()}/${carCid.toString()}.car`
```
alanshaw, line 123: `toString` unnecessary and with it makes slightly more readable:

```suggestion
  const r2Key = `${carCid}/${carCid}.car`
```

packages/cron/test/pins.spec.js
```diff
@@ -59,7 +59,7 @@ const createPin = (cid, status = 'pinned') => {
   return pin
 }
 
-describe('cron - pins', () => {
+describe.skip('cron - pins', () => {
```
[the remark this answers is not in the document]
olizilla: Yes! I've made a list followup tasks in the PR description. The plan is to let the cron pin sync job keep running after this PR is deployed, until it has updated the last pin that was pinned to cluster. Then we can stop that cron and delete the code for it.

packages/api/wrangler.toml
```diff
@@ -27,20 +27,38 @@ deleted_classes = ["NameRoom0"]
...
 # name = "web3-storage-production"
 account_id = "fffa4b4363a7e5250af8357087263b3a" # Protocol Labs CF account
 route = { pattern = "https://api.web3.storage/*", zone_id = "7eee3323c1b35b6650568604c65f441e" }
 vars = { CLUSTER_API_URL = "https://web3.storage.ipfscluster.io/api/", ENV = "production", PG_REST_URL = "https://web3-storage-pgrest-prod.herokuapp.com", GATEWAY_URL = "https://ipfs.io" }
 
+[[env.production.r2_buckets]]
+# what's it called in r2?
+bucket_name = "carpark-prod-0"
+# what's it called as a property of the env object
+binding = "carpark"
```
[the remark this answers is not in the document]
olizilla: SCREAMING_SNAKE_CASE IS COOL! 📣🐍

THERE IS NO NEED TO MAKE OUR BINDINGS LOOK LIKE VARS WE PASSED IN, BUT I WILL, FOR CONSISTENCY, MAKE IT SHOUTY.

packages/api/src/car.js
alanshaw: You could always have a single `env.r2BucketUrl`.

I vote we put _something_ here - we should keep track of who uploaded what.

You could just make something up - all we'd need to do is create a worker to provide access:

```
prod-2.bucket.web3.storage
```

packages/api/src/car.js
```diff
@@ -127,20 +134,50 @@ export async function handleCarUpload (request, env, ctx, car, uploadType = 'Car
...
 
   /** @type {(() => Promise<any>)[]} */
-  const tasks = [async () => {
+  const tasks = []
+
+  // ask linkdex for the dag structure across the set of CARs in S3 for this upload.
+  const checkDagStructureTask = () => pRetry(async () => {
+    const url = new URL(`/?key=${s3Key}`, env.LINKDEX_URL)
+    const res = await fetch(url)
+    assert(res.ok)
```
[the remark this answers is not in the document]
alanshaw: yes but this is run in `waitUntil` so has no bearing on the response.

packages/api/src/car.js
olizilla: I think we want to be able to map between CARs in s3 and CARs in r2 _somehow_. Perhaps we could use fictional url for our r2 bucket that we could turn into a real worker that serves things from r2 if we need to. We can't use the s3 url directly at the moment either, as we don't have public access enabled

```js
`http://carpark.web3.storage/${bucketKey}`
```

packages/api/src/env.js
```diff
@@ -129,6 +133,8 @@ export function envAll (req, env, ctx) {
   // @ts-ignore
   env.MODE = env.MAINTENANCE_MODE || DEFAULT_MODE
 
+  env.ELASTIC_IPFS_PEER_ID = env.ELASTIC_IPFS_PEER_ID ?? 'bafzbeibhqavlasjc7dvbiopygwncnrtvjd2xmryk5laib7zyjor6kf3avm'
```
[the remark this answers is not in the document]
olizilla: my thinking was, it's not likely to change any time soon, so have a default value that we can override in the future if we need to...

packages/api/src/error-handler.js
```diff
@@ -8,7 +8,7 @@ import { HTTPError, PinningServiceApiError } from './errors.js'
  * @param {Request} request
  */
 export function errorHandler (err, { log }, request) {
-  console.error(err.stack)
+  console.error(err)
```
olizilla, line 11: this make it so you can actually see the error.cause in the console log when debugging. it's very helpful.

alanshaw (changes requested):

packages/api/src/car.js
olizilla: Done

packages/api/wrangler.toml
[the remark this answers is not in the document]
olizilla: DONE!

packages/api/src/car.js
olizilla: DONE! I have made CARPARK_URL an env var.

packages/api/src/car.js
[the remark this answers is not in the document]
olizilla: I've dropped the assert dependency as this is the only thing using it and it now throws a LinkdexError.

alanshaw (approved):

PR #1790, opened 2022-08-23
chore: add a script that generated many uploads

joshJarr:
Adds a script that generates a number of uploads and optionally generates pins to be used when testing large sets of paginated data.

### DIFF
--- a/packages/api/scripts/generate-upload-data.js
+++ b/packages/api/scripts/generate-upload-data.js
@@ -1,0 +1,102 @@
+/**
+ * This script will generate many uploads for a given user token using the web3.storage API endpoints.
+ * Handy when working with and testing features that require a pristine mass set of data, such as pagination.
+ * Arguments can be passed with a arg=value pattern, ie token='<YOUR TOKEN HERE>'. Passing a token is the only required argument.
+ * Not passing a value will default to a truthy value. ie appending `pin` argument is the same as `pin=true`.
+ * This script requires a users token to be passed or assigned to an environment variable API_TOKEN.
+ * Passable arguments are:
+ * - token: the token of the user these uploads will be authenticated with.
+ * - uploads: The amount of uploads that need to be added, defaulting to 25.
+ * - url: The URL of the API, defaulting to localhost.
+ * - pin: Adds another request for each upload to pin that upload generating pin results, defaulting to false.
+ * Example usage.
+ * From the API folder, run the following command using a generated token.
+ * - node scripts/generate-upload-data.js token='yourApiToken'
+ * This will generate 25 uploads for the given user. Append the pin argument if these uploads should also be pinned.
+ */
+
+import fetch, { Blob } from '@web-std/fetch'
+
+const requiredArgs = [
+  'token'
+]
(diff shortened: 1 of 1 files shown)


packages/api/scripts/generate-upload-data.js
```diff
@@ -0,0 +1,102 @@
...
+ * This script will generate many uploads for a given user token using the web3.storage API endpoints.
+ * Handy when working with and testing features that require a pristine mass set of data, such as pagination.
+ * Arguments can be passed with a arg=value pattern, ie token='<YOUR TOKEN HERE>'. Passing a token is the only required argument.
+ * Not passing a value will default to a truthy value. ie appending `pin` argument is the same as `pin=true`.
+ * This script requires a users token to be passed or assigned to an environment variable API_TOKEN.
+ * Passable arguments are:
+ * - token: the token of the user these uploads will be authenticated with.
+ * - uploads: The amount of uploads that need to be added, defaulting to 25.
+ * - url: The URL of the API, defaulting to localhost.
+ * - pin: Adds another request for each upload to pin that upload generating pin results, defaulting to false.
```
flea89, line 11: I wonder if we can make this slightly clearer. 
As usual "pin" is a confusing word 😢 

I would try to make it extremely obvious that here we refer to creating Pinning Service API requests here rather than just saying pin, which might be confused with cluster pinning.

maybe something along these lines:

```
psa_pin_requests: If true, on top of uploading the file, a PSA pin request is created for the given CID. 
```

What do you think? 
Feel free to update/reword.

packages/api/scripts/generate-upload-data.js
```diff
@@ -0,0 +1,102 @@
...
+  'token'
+]
+
+const optionalArgs = {
+  uploads: '25',
+  pin: false,
+  url: 'http://127.0.0.1:8787'
+}
+
+const argv = process.argv.slice(2).map((arg) => [
```
flea89, line 30: I reckon is ok to keep it as is, but FYI [sade](https://github.com/lukeed/sade) is used across the project to build command line scripts. ie, in scripts/cli.
That would take take of argv parsing, validation, help etc.

flea89 (changes requested):
This is useful, thanks @joshJarr 
Just a couple of minor comments

packages/api/scripts/generate-upload-data.js
joshJarr: Really good point! Yep this makes so much more sense, thanks

packages/api/scripts/generate-upload-data.js
joshJarr: Ahh sweet, yeah the args here are pretty simple, but if extending it then using sade to parse 'em would be amazing. Thanks!

joshJarr:
> This is useful, thanks @joshJarr Just a couple of minor comments

Thanks! Addressed now - let me know if you have any more feedback 💪

flea89 (approved):

flea89:
Thanks @joshJarr, lgtm

PR #1797, opened 2022-08-24
fix: update package-name for w3 release please

### LINKED ISSUE #1789: release-please is failing for w3 package
The `release-please` bot is failing to run properly for the `packages/w3` package. See [example GH Action run](https://github.com/web3-storage/web3.storage/runs/7954739477?check_suite_focus=true).

As a result of this, the release-please bot is creating weird PRs which:
* Contain release notes for things which have already been released;
* Revert the version number to `1.0.0` because it can't find the version of the previous release.

The first bad PR was [#1695](https://github.com/web3-storage/web3.storage/pull/1695). After discussion with @olizilla, and noticing that the release-please job had hit a rate limit, we thought it might just be a one-off problem with that release due to the transiient error. So I closed that release PR and then created a new bug fix ([#1777](https://github.com/web3-storage/web3.storage/issues/1777)) which would trigger a new release.

But the new release, [#1778](https://github.com/web3-storage/web3.storage/pull/1778) has got all the same problems.

My digging so far has led me to the following conclusions/thoughts:

### It works its way through _all_ the release tags
In the log out (see first link), we get a load of:
```⚠ Found release tag with component 'web3.storage', but not configured in manifest```
and if you look at the order of those, including the final
``` Unable to parse release name: ⁂```
you'll see that they match the order of the [releases for this repo](https://github.com/web3-storage/web3.storage/releases).

So it's clearly successfully fetching all the releases from GH, but when it comes across a `w3` release it gives:
```⚠ Unable to find expected version for path 'packages/w3' in manifest``` instead of parsing the release info/tag to get the version number (and other info?). I'm guessing that it's _probably_ getting this list of releases from the GH API, rather than from `git tag --list --sort creatordate`, so if it's then making another API request to fetch the details of each release, then this _might_ be related to #1776.

### Why can't it parse the release name?

I'm not sure. But I suspect that it might have something to do with `w3` vs. `"@web3-storage/w3"` (from package.json). The package name was changed in [21eac8a0d57f2cdfb5617e13bcd595ed71ce432f](https://github.com/web3-storage/web3.storage/commit/21eac8a0d57f2cdfb5617e13bcd595ed71ce432f). That was on the 22nd of July 2022, and the [last release](https://github.com/web3-storage/web3.storage/releases?q=w3&expanded=true) of the package was on the 19th of April 2022, so we haven't had a successful release since then.

The configuration for the `w3` package, in terms of release-please config, name in package.json, etc, seems to match that of our other packages, but maybe it's the naming pattern of the previous releases which aren't matching the right pattern. I haven't yet dug into exactly how release-please does its matching or exactly what it's matching against.

adamalton:
Possibly fixes #1789.

### DIFF
--- a/.github/workflows/w3.yml
+++ b/.github/workflows/w3.yml
@@ -61,7 +61,7 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           release-type: node
           monorepo-tags: true
-          package-name: w3
+          package-name: '@web3-storage/w3'
           changelog-types: '[{"type":"feat","section":"Features","hidden":false},{"type":"fix","section":"Bug Fixes","hidden":false},{"type":"chore","section":"Other Changes","hidden":false}]'
       - uses: actions/checkout@v2
         if: ${{ steps.tag-release.outputs.releases_created }}


adamalton:
Looking at the run of the 'release' job: https://github.com/web3-storage/web3.storage/runs/7992994109?check_suite_focus=true

This is still encountering the same problem as before: it's working it's way through _all_ the releases for all the packages in the web3.storage repo, and whenever it encounters a release for the w3 package it gets:
```⚠ Unable to find expected version for path 'packages/w3' in manifest```

So this has not helped.

adamalton:
If #1799 fixes the problem, *and* if when we do the _next_ `w3` release the problem doesn't come back, then we can just delete this PR.

If the problem comes back then _maaaybe_ this PR might help, but it also might not do. If the problem comes back then I think we should go back to the original issue (#1789) and have a re-think.

joshghent:
Looks like we can close this because the release-please has fixed itself.

PR #1806, opened 2022-08-25
fix: return correct issuer from magic token in `tryMagicToken`.

joshJarr:
Was getting an error on staging when logging in:
```
{"code":"ERROR_MAGIC_TOKEN_REQUIRED","message":"Must be logged in with magic.link"}
```

Looks like the issuer wasn't being passed correctly, this PR hopefully resolves this!🤞


francois-potato (approved):

olizilla (commented):
function jsDoc says
```
@returns {null | {issuer: string}} information about the api token
```
and calling code expects that shape too https://github.com/web3-storage/web3.storage/blob/4151e3bea7dce393ecdf3e080b74329ae82e703e/packages/api/src/user.js#L40

so looks legit to me. I don't know it it's possible, but it'd be ideal to add a test that covers this path if we can.

joshJarr:
Sorry @olizilla! @francois-potato reviewed swiftly as this was blocking some stuff - will keep track and address PR feedback
Yep - adding a test for this method would make a tonne of sense  - I'll make a fix on the restored branch & can also address any more feedback there too

olizilla:
Nice! Thanks

> I'll make a fix on the restored branch

A fresh PR from main is fine!

PR #1813, opened 2022-08-29
feat: add tsc checking to packages/api

gobengo:
Motivation:
* add a type checker and run it in ci to avoid bugs like [this](https://github.com/web3-storage/web3.storage/pull/1806) from the other day, and to be able to dev/refactor faster and with more confidence once we fix the type check errors that were already in here unchecked


yusefnapora:
:tada:  for actually running `tsc`! I totally assumed that we were already doing this... but the second best time to plant a tree is today, as they say :smile:

packages/api/src/auth.js
```diff
@@ -164,19 +170,19 @@ function authenticateMagicToken (
...
 
-  /** @type {import('@magic-sdk/admin').ParsedDIDToken} */
+  /** @type {import('@magic-sdk/admin').ParsedDIDToken | null} */
   let decodedToken = null
 
   try {
     decodedToken = env.magic.token.decode(token)
   } catch (error) {
-    console.warn('error decoding magic token', error.name, error.message)
+    console.warn('error decoding magic token', error)
```
olizilla, line 179: minor but this will change how errors get logged... but maybe this is better tho

```console
❯ node
Welcome to Node.js v18.8.0.
Type ".help" for more information.
> console.warn('foo', 'bar', 'baz')
foo bar baz
undefined
> console.warn('foo', {name: 'bar', message: 'baz' } )
foo { name: 'bar', message: 'baz' }
undefined
> console.warn('foo', new Error('baz') )
foo Error: baz
    at REPL2:1:21
    at Script.runInThisContext (node:vm:129:12)
    at REPLServer.defaultEval (node:repl:572:29)
    at bound (node:domain:433:15)
    at REPLServer.runBound [as eval] (node:domain:444:12)
    at REPLServer.onLine (node:repl:902:10)
    at REPLServer.emit (node:events:525:35)
    at REPLServer.emit (node:domain:489:12)
    at [_onLine] [as _onLine] (node:internal/readline/interface:425:12)
    at [_line] [as _line] (node:internal/readline/interface:886:18)
undefined
```

olizilla (approved):
Thanks for fixing this! 

Please raise an issue for a follow up to solve (as much as possible) the ts-ignore'd problems.

packages/api/src/auth.js
gobengo: @olizilla I thought it would be fine... technically something could be thrown without a `.name` or `.message`, but with some other `toString()`, which I think is why tsc complains that `error` is `unknown` unless you do an `error instanceof Error` check.

packages/api/tsconfig.json
```diff
@@ -0,0 +1,33 @@
+{
+  "compilerOptions": {
+    "outDir": "./dist",
+    "allowJs": true,
+    "checkJs": true,
+    "target": "esnext",
+    "lib": ["ESNext", "DOM", "DOM.Iterable"],
+    "strict": true,
+    "moduleResolution": "node",
+    "sourceMap": true,
```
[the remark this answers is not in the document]
gobengo: Only if `noEmit` is overridden by a cli flag, which is unlikely atm, but not out of the question.

gobengo:
@olizilla 

> Please raise an issue for a follow up to solve (as much as possible) the ts-ignore'd problems.

https://github.com/web3-storage/web3.storage/issues/1822

packages/api/tsconfig.json
```diff
@@ -0,0 +1,32 @@
+{
```
vasco-santos, line 1: @gobengo could we add in here the types for the actual CF workers? I think that brings a ton of benefits. Example:

 https://github.com/nftstorage/nftstorage.link/blob/main/packages/edge-gateway/tsconfig.json#L7

vasco-santos: We currently have all types for bindings specified in `env.js`, which is not cool. Adding CF Worker types and creating a proper ts file for the bindings injected makes things so much cleaner. You can also see https://github.com/nftstorage/nftstorage.link/blob/main/packages/edge-gateway/src/bindings.d.ts

gobengo: https://github.com/web3-storage/web3.storage/issues/1832

Thanks for the good suggestion. I'll do it when I have a sec

PR #1814, opened 2022-08-30
chore: point users to w3link in docs

dchoi27:


yusefnapora (approved):
nice, thanks @dchoi27

PR #1818, opened 2022-08-31
feat: PUT /user/payment API saves payment info in stripe.com

gobengo:
Motivation
* https://github.com/web3-storage/web3.storage/issues/1748

Parts
* [x] ability to create stripe customers for users and look them up
* [x] PUT /user/payment actually saves payment method id
* [x] GET /user/payment includes stripe.card card info for frontend
* [x] add db migration that will work


packages/api/src/utils/stripe.js
```diff
@@ -0,0 +1,124 @@
...
+  /**
+   * @param {{id: string}} user
+   * @returns {Promise<Customer>}
+   */
+  async getOrCreateForUser (user) {
+    const userIdMetadataProperty = 'web3StorageUserId'
+    // const searchQuery = `metadata["${userIdMetadataProperty}"]:"${user.id}"`
+    const searchQuery = `metadata["${userIdMetadataProperty}"]:"${user.id}"`
+    // console.log({ searchQuery })
+    const searchResponse = await this.stripe.customers.search({
```
gobengo, line 95: Unfortunately the search index doesn't update for like 5+ sec after writes. So this hope to do a version of this without a db change aint gonna work. I think I'll need a db change to store [user.id, stripeCustomer.id]

packages/api/src/user.js
```diff
@@ -78,7 +79,7 @@ async function loginOrRegister (request, env) {
 
   const parsed =
     data.type === 'github'
-      ? parseGitHub(data.data, metadata)
+      ? await parseGitHub(data.data, metadata)
```
vasco-santos, line 82: Not promise?

packages/api/src/utils/billing-types.ts
```diff
@@ -0,0 +1,59 @@
+export type StripePaymentMethodId = string;
```
vasco-santos, line 1: no strong feelings against this, but we are using jsdocs for types everywhere else in this repo. Would be great to uniform how we want to proceed with the whole team. I am up to go with types as in your PR, but prefer us to be consistent rather than having different styles all around.

Could you raise this with the team? (not blocking the PR in any way, we can revert back if we need to merge this ASAP)

packages/api/src/user.js
```diff
@@ -540,40 +541,48 @@ const notifySlack = async (
  * Get a user's payment settings.
  *
  * @param {AuthenticatedRequest} request
- * @param {import('./env').Env} env
+ * @param {Pick<import('./env').Env, 'billing'|'customers'>} env
  */
 export async function userPaymentGet (request, env) {
```
vasco-santos, line 546: Did we ship this already? I guess no right? This is a breaking change.

packages/api/src/utils/stripe.js
```diff
@@ -0,0 +1,378 @@
...
+   * @protected
+   * @param {StripeComForBillingService} stripe
+   */
+  constructor (stripe) {
+    /** @type {StripeComForBillingService}  */
+    this.stripe = stripe
+    /**
+     * @type {BillingService}
+     */
+    const instance = this // eslint-disable-line
```
vasco-santos, line 40: do we need this?

packages/api/src/utils/stripe.js
```diff
@@ -0,0 +1,378 @@
...
+
+  async getPaymentMethod (customerId) {
+    const response = await this.stripe.customers.retrieve(customerId, {
+      expand: ['invoice_settings.default_payment_method']
+    })
+    if (response.object !== 'customer') {
+      throw new Error('unable to get payment method')
+    }
+    if (response.deleted) {
+      throw new Error('unexpected response.deleted')
```
vasco-santos, line 51: Are these internal Errors or user input error?

packages/db/index.js
```diff
@@ -1274,4 +1274,52 @@ export class DBClient {
...
+
+  /**
+   * Get the Customer for a user
+   * @param {string} userId
+   * @returns {null|{ id: string }} customer
+   */
+  async getUserCustomer (userId) {
+    const { data, error } = await this._client
+      .from('user_customer')
+      .select(['customer_id'].join(','))
```
vasco-santos, line 1313: No need for the join

packages/db/index.js
```diff
@@ -1274,4 +1274,52 @@ export class DBClient {
...
+      .eq('user_id', userId)
+    if (error) {
+      console.error('getUserCustomer query error', error)
```
vasco-santos, line 1316:
```suggestion
```

packages/db/postgres/migrations/023-add-user-customer-table.sql
```diff
@@ -0,0 +1,7 @@
+-- add table to keep track of 0..1 customer id for each user
+CREATE TABLE IF NOT EXISTS user_customer
+(
+  id               BIGSERIAL PRIMARY KEY,
+  user_id          BIGINT NOT NULL UNIQUE REFERENCES public.user (id),
```
vasco-santos, line 5: We do queries for user_id, probably we should add an index for it? What you think?

vasco-santos (commented):
Good job @gobengo 

Did a first pass through this and left feedback. Let me know what you think

packages/api/src/utils/billing-types.ts
gobengo: > Could you raise this with the team? (not blocking the PR in any way, we can revert back if we need to merge this ASAP)

Def 👍 

> Would be great to uniform how we want to proceed with the whole team.

Personally I think it's not that important to be fully uniform about this. The jsdocs `@typedefs` are useful to create type aliases in the js files for intra-file use (it cuts down on many big `import(...)` in the jsdocs). But the jsdoc syntax is kludgy and not always able to express the whole type system, so then it makes sense to use .ts files. Even though there are ts files for some types, because those ts files are only linked to from jsdocs, all the js should run without transpilation still.

I don't mind at all establishing some kind of more strict norm, just presenting the case for incrementalism/agility over uniformity.

gobengo: https://github.com/web3-storage/web3.storage/discussions/1858

packages/api/scripts/cli.js
```diff
@@ -32,20 +33,23 @@ prog
...
+            })
+            build.onResolve({ filter: /^node-fetch$/ }, () => {
+              return { path: path.resolve(__dirname, 'fetch.js') }
+            })
+            build.onResolve({ filter: /^cross-fetch$/ }, () => {
+              return { path: path.resolve(__dirname, 'fetch.js') }
+            })
+          }
+        },
+        NodeModulesPolyfillPlugin()
```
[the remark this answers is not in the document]
gobengo: Added a comment explaining how 'stripe' deps on some node things (though I think we dont actually import those parts, e.g. we use the Fetch http adapater instead of the Node one)

packages/api/src/env.js
```diff
@@ -168,8 +175,61 @@ export function envAll (req, env, ctx) {
...
+ * @property {string} stripeSecretKey
+ */
+
+/**
+ * @param {Pick<Env, 'STRIPE_SECRET_KEY'|'db'>} env
+ * @returns {import('./utils/billing-types').BillingEnv & StripeDerrivedEnv}
+ */
+function createStripeBillingContext (env) {
+  const stripeSecretKey = env.STRIPE_SECRET_KEY
+  assert.ok(stripeSecretKey, 'env STRIPE_SECRET_KEY is required')
```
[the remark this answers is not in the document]
gobengo: sure

packages/api/scripts/cli.js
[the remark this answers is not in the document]
gobengo: (afaict it wasn't related to the use of `assert`)

packages/api/src/env.js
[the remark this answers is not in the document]
gobengo: 83431fc2fd01ab1fd0cea84b077e146c9dbac0f6

packages/api/src/utils/stripe.js
gobengo: @vasco-santos this is not related to user error. These would be legitimately unexpected errors that might indicate something else was wrong in the stripe.com-hosted customer db. It shouldn't need to keep anyone up at night, but it would be nice to have a modest alarm if these errors ever do come up, and throwing explicitly will give the debugger (probably me) a stack trace when that happens. Technically this case can happen (the type system told me), but it shouldn't, and if it does it is probably something that should be escalated to someone like us

packages/api/src/env.js
```diff
@@ -35,6 +39,7 @@ import { defaultBypassMagicLinkVariableName } from './magic.link.js'
  * @property {string} [LOGTAIL_TOKEN]
  * @property {string} MAINTENANCE_MODE
  * @property {string} [DANGEROUSLY_BYPASS_MAGIC_AUTH]
+ * @property {string} [STRIPE_SECRET_KEY]
```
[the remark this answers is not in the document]
gobengo: 9da3c8b601b13dd959df437f266bef030d30825b

packages/api/src/env.js
```diff
@@ -168,8 +175,61 @@ export function envAll (req, env, ctx) {
     )
   }
 
-  // via https://stripe.com/docs/api/payment_methods/object
-  // this can be used to mock realistic values of a stripe.com paymentMethod id
-  // after fulls tripe integration, this may not be needed on the env
-  env.mockStripePaymentMethodId = 'pm_1LZnQ1IfErzTm2rETa7IGoVm'
+  Object.assign(env, (env.ENV === 'test') ? createTestBillingEnv() : createStripeBillingContext(env))
```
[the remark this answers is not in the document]
gobengo: @vasco-santos great suggestion. I implemented as
475be014f8ef21ca50c1b56aa94c3e0c6ca7c045
3366de256e1d618a5188d387b90693f7de07afa2

packages/api/src/utils/stripe.js
gobengo: https://github.com/microsoft/TypeScript/issues/30156#issuecomment-952667829

gobengo: I made it nicer, one line, no eslint-disable 67deec440f4849cf36583e0c995e2bd31d882903

packages/api/src/utils/stripe.js
gobengo: I removed these two untyped `Errror`s. The first wasn't really needed. The second I added a specific `CustomerNotFound` type for, and added it to the function signature so tsc helps callers be ready for it or other errors. cd82b249164dbd8d373d653c9671f564bcba8e63

packages/db/index.js
gobengo: according to the docs (and tsc types, which aren't being checked in db package yet), it must be a string. https://supabase.github.io/postgrest-js/classes/PostgrestQueryBuilder.html#select

(not disagreeing that the join works now at runtime, but I'd prefer to keep it here so it doesn't have to change again if/when we enable type checking https://github.com/web3-storage/web3.storage/issues/1860 )

packages/db/postgres/migrations/023-add-user-customer-table.sql
gobengo: good idea! 3df2067c23a0df1465a2555e5dde1f9baaefd8dc

gobengo:
@vasco-santos Thanks for your review. I believe I addressed all of your comments. (I left an 🎉 emoji on comments I acted on). wdyt?

gobengo:
I integrated some frontend work by @adamalton  on top of this that actually results in a functional demo of the backend changes in this PR. It's really those changes too that we want to be deploying to staging and beyond. Let's consider this PR superceded by https://github.com/web3-storage/web3.storage/pull/1852 . @vasco-santos it's basically identical to what you reviewed already + changes based on your review.

packages/api/src/env.js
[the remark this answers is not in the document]
vasco-santos: Assert will throw `AssertionError` and in the error handler we will need to define what user error we want to send (and whether it should have 500 code or 4xx code). We could get all this with assert, but I think it would make things more complex with current code

packages/api/src/utils/stripe.js
vasco-santos: that sounds good. Thanks for details :)

PR #1823, opened 2022-08-31
feat: packages/api typecheck checks ./test dir

gobengo:
Motivation:
* followup to https://github.com/web3-storage/web3.storage/pull/1813
  * in that pr I didn't notice that `./test` dir would not be typechecked. This adds that

### DIFF
--- a/packages/api/test/maintenance.spec.js
+++ b/packages/api/test/maintenance.spec.js
@@ -62,7 +62,7 @@ describe('maintenance middleware', () => {
     invalidModes.forEach((m) => {
       assert.throws(() => block(() => {}, {
         MODE: m
-      }, /invalid maintenance mode/))
+      }), /invalid maintenance mode/)
     })
   })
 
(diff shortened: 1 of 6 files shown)


packages/api/test/maintenance.spec.js
```diff
@@ -62,7 +62,7 @@ describe('maintenance middleware', () => {
     invalidModes.forEach((m) => {
       assert.throws(() => block(() => {}, {
         MODE: m
-      }, /invalid maintenance mode/))
+      }), /invalid maintenance mode/)
```
olizilla, line 65: yikes! a fine example of "type checking is good, actually".

olizilla (approved):

packages/api/test/maintenance.spec.js
gobengo: @olizilla haha yeah, not to mention all the `assert(` I changed to `assert.equal(`. I think the former wasn't actually asserting what it intended (truthiness vs equality)

PR #1828, opened 2022-09-01
chore: add peering reference doc

### LINKED ISSUE #1725: How-to for peering with Web3.Storage
Once we have the final domain name for our E-IPFS node (in the next week or so, @francardoso93 is working on this and can follow up when ready), let's share the peer ID / info in a new section here under Reference, including information about what situations you'd consider peering (you're running your own IPFS nodes and/or gateway) and how to peer. It should also mention that it takes some time for new uploads' records to be picked up by the DHT (via indexers), and if you're using your own IPFS nodes to fetch data, you can circumvent this by peering.

Let's also stick a line in the Store docs that references this section as well as something to keep in mind!

Sister issue to: https://github.com/nftstorage/nft.storage/issues/2129

yusefnapora:
This adds a "Peering with Web3.Storage" doc in the reference section to tell people how and why to peer with our elastic IPFS instance.

Closes #1725


olizilla (approved):
Noice 👍

PR #1844, opened 2022-09-06
feat: new job to backup pins to s3

### LINKED ISSUE #794: Pinning - Backups

joshghent:
⚠️ Requires a new migration on the `psa_pins_request` table to add the new `backup_urls` column. ⚠️ 

### About this job
This new cron job which runs every 4 hours, grabs 10,000 pin requests, gets the car file and uploads it to s3.
It behaves in a similar way to [nftstorage/backup](https://github.com/nftstorage/backup).
It uses Dagula to grab the car file from the IPFS peer.


packages/cron/src/jobs/pins-backup.js
```diff
@@ -0,0 +1,239 @@
...
+
+  /**
+   * @param {import('@aws-sdk/client-s3').S3Client} s3
+   * @param {string} bucketName
+   * @param {import('./bindings').BackupContent} bak
+   */
+  async s3Upload (s3, bucketName, bak) {
+    const key = `complete/${bak.contentCid}.car`
+    const region = await s3.config.region()
+    const url = new URL(`https://${bucketName}.s3.${region}.amazonaws.com/${key}`)
```
flea89, line 88: If I'm not mistaken, with the current implementation, given we can have different PSA_PIN_REQUESTS pointing to the same `CID`, we would end up uploading the same car multiple time to S3 isn't it?
I'm not quite aware if we need a particular folder structre, but assuming this one the correct one (and the same cid would always endup in the same path) we should check if already exists, and if yes, avoid uploading it?

Also, it seems from here that we always use content_cid (and not source_cid) for the key meaning different cid versions are not something we should be caring about?

flea89 (changes requested):
This is getting together 🎉 
I did a really quick review pass and left some high-level comments for you.

Also noticed some types still need updating across the board.

packages/cron/src/jobs/pins-backup.js
```diff
@@ -0,0 +1,229 @@
...
+import formatNumber from 'format-number'
+import batch from 'it-batch'
+import { pipe } from 'it-pipe'
+import { CID } from 'multiformats'
+import * as raw from 'multiformats/codecs/raw'
+import { Readable } from 'stream'
+
+export default class Backup {
+  constructor (env) {
+    this.UPDATE_BACKUP_URL = 'UPDATE psa_pin_request SET backup_urls = $1 WHERE id = $2 AND content_cid = $3 RETURNING *'
```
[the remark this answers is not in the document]
joshghent: I see why this was confusing. Updated :)

packages/cron/src/jobs/pins-backup.js
```diff
@@ -0,0 +1,229 @@
...
+   * @param {Object} [options]
+   * @param {Date} [options.startDate]
+   * @param {(cid: CID) => Promise<boolean>} [options.filter]
+   */
+  async * getPinsNotBackedUp (db, options = {}) {
+    const { rows } = await db.query(this.GET_PINNED_PINS_QUERY, [
+      this.LIMIT
+    ])
+    if (!rows.length) return
+    const uploads = rows.filter(r => !r.url)
```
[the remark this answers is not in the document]
joshghent: Updated

packages/cron/test/pins-backup.spec.js
```diff
@@ -0,0 +1,198 @@
...
+    }
+  })
+
+  after(async () => {
+    await roPg.end()
+    await rwPg.end()
+  })
+
+  it.only('should attempt to backup pins to S3', async () => {
+    await backup.backupPins({ env, roPg, rwPg, cluster })
```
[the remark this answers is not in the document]
joshghent: it used to! but I've removed it now :)

.github/workflows/cron-backup-pins.yaml
```diff
@@ -0,0 +1,40 @@
...
+      - uses: bahmutov/npm-install@v1
+      - name: Run job
+        env:
+          DEBUG: '*'
+          ENV: ${{ matrix.env }}
+          STAGING_PG_CONNECTION: ${{ secrets.STAGING_PG_CONNECTION }}
+          STAGING_RO_PG_CONNECTION: ${{ secrets.STAGING_PG_CONNECTION }} # no replica for staging
+          PROD_PG_CONNECTION: ${{ secrets.PROD_PG_CONNECTION }}
+          PROD_RO_PG_CONNECTION: ${{ secrets.PROD_RO_PG_CONNECTION }}
+        run: yarn --cwd packages/cron start:pins:backup
```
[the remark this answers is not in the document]
joshghent: yeah good point, I've updated to use npm

packages/cron/src/bin/pins-backup.js
```diff
@@ -0,0 +1,22 @@
...
+import { getCluster, getPgPool } from '../lib/utils.js'
+
+async function main () {
+  const rwPg = getPgPool(process.env, 'rw')
+  const roPg = getPgPool(process.env, 'ro')
+  const cluster = getCluster(process.env)
+  const backup = new Backup(process.env)
+
+  try {
+    await backup.backupPins({ rwPg, roPg, cluster })
```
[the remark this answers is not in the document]
joshghent: done

packages/cron/src/jobs/pins-backup.js
```diff
@@ -0,0 +1,229 @@
...
+import AWS from 'aws-sdk'
+import debug from 'debug'
+import formatNumber from 'format-number'
+import batch from 'it-batch'
+import { pipe } from 'it-pipe'
+import { CID } from 'multiformats'
+import * as raw from 'multiformats/codecs/raw'
+import { Readable } from 'stream'
+
+export default class Backup {
```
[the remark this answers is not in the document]
joshghent: yeah it's a good point, it was kind of just that "double-check". But yes this can be safely removed.

packages/cron/src/jobs/pins-backup.js
```diff
@@ -0,0 +1,229 @@
...
+  constructor (env) {
+    this.UPDATE_BACKUP_URL = 'UPDATE psa_pin_request SET backup_urls = $1 WHERE id = $2 AND content_cid = $3 RETURNING *'
+    this.fmt = formatNumber()
+    this.SIZE_TIMEOUT = 1000 * 10 // timeout if we can't figure out the size in 10s
+    this.BLOCK_TIMEOUT = 1000 * 30 // timeout if we don't receive a block after 30s
+    this.REPORT_INTERVAL = 1000 * 60 // log download progress every minute
+    this.MAX_DAG_SIZE = 1024 * 1024 * 1024 * 32 // don't try to transfer a DAG that's bigger than 32GB
+    this.log = debug('backup:pins')
```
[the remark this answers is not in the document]
joshghent: We don't... but it's a pattern that is found in other areas of the system. So might be worth flagging.

packages/cron/src/jobs/pins-backup.js
```diff
@@ -0,0 +1,229 @@
...
+    this.env = env
+    this.LIMIT = env.QUERY_LIMIT !== undefined ? env.QUERY_LIMIT : 10000
+    this.GET_PINNED_PINS_QUERY = `
```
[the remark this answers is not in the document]
joshghent: Yes that is correct. There could be some extra work to add some de-duplication here. Not sure how common that case would be, but perhaps something for a V2.

packages/cron/src/jobs/pins-backup.js
joshghent: Ok so, I've updated it so it's `complete/${bak.contentCid}/${bak.sourceCid}.car` for the key
And then added it so if that key already exists then we won't re-upload it.

joshghent:
Ok, so before this is merged it needs a the migration running which adds a new column `backup_urls` to the `psa_pins_request`.
This cron job will need verifying on staging and make sure that it successfully backs up the pins.
After that, it can be safely merged to production.

packages/cron/src/jobs/pins-backup.js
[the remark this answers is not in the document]
joshghent: I've updated the system so it will first check if the file is in S3 before uploading. But changes to the query could be made also.

packages/cron/src/jobs/pins-backup.js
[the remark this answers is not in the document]
flea89: The comment is not pointing to the right line anymore, should be [line above](https://github.com/web3-storage/web3.storage/pull/1844/files#diff-4df8cd6f6a1194703527832b6e009d1515b1d8f708afe9a5af7c31f446f9aadeR25)[https://github.com/web3-storage/web3.storage/pull/1844/files#diff-4df8cd6f6a1194703527832b6e009d1515b1d8f708afe9a5af7c31f446f9aadeR25](url)

@alanshaw you might know more about this.

packages/cron/src/jobs/pins-backup.js
flea89: Ace!

.github/workflows/cron-backup-pins.yaml
```diff
@@ -0,0 +1,40 @@
+name: Cron Backup Pins
+
+on:
+  schedule:
+    - cron: '*/30 * * * *'
```
alanshaw, line 5: Why not just schedule for the max amount of time a job can run for 6h?

packages/cron/src/jobs/pins-backup.js
alanshaw: The current convention is `complete/${bak.sourceCid}.car`

packages/db/postgres/tables.sql
```diff
@@ -307,11 +307,13 @@ CREATE TABLE IF NOT EXISTS psa_pin_request
   meta            jsonb,
   deleted_at      TIMESTAMP WITH TIME ZONE,
   inserted_at     TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL,
-  updated_at      TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL
+  updated_at      TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL,
+  backup_urls     TEXT[]
```
alanshaw, line 311: Might be nice if this was `NOT NULL DEFAULT []` so you don't have to distinguish between null and empty.

packages/cron/src/jobs/pins-backup.js
```diff
@@ -0,0 +1,248 @@
...
+   * @param {number} [options.maxDagSize] Skip DAGs that are bigger than this.
+   */
+  exportCar (ipfs, options = {}) {
+    /**
+     * @param {AsyncIterable<import('./bindings').BackupCandidate>} source
+     * @returns {AsyncIterableIterator<import('./bindings').BackupContent>}
+     */
+    return async function * (source) {
+      for await (const candidate of source) {
+        yield { ...candidate, content: this.ipfsDagExport(ipfs, candidate.sourceCid, options) }
```
alanshaw, line 125: I think `this` will not be the expected object here.

packages/cron/src/jobs/pins-backup.js
```diff
@@ -0,0 +1,248 @@
...
+          { code: 'ERR_TOO_BIG' }
+        )
+      }
+
+      reportInterval = setInterval(() => {
+        const formattedTotal = bytesTotal ? this.fmt(bytesTotal) : 'unknown'
+        this.log(`received ${this.fmt(bytesReceived)} of ${formattedTotal} bytes`)
+      }, this.REPORT_INTERVAL)
+
+      for await (const chunk of ipfs.dagExport(cid, { timeout: this.BLOCK_TIMEOUT })) {
```
alanshaw, line 160: `dagExport` is not implemented on the IPFS Cluster client https://github.com/nftstorage/ipfs-cluster#api

packages/cron/src/jobs/pins-backup.js
```diff
@@ -0,0 +1,248 @@
...
+      for await (const candidate of source) {
+        yield { ...candidate, content: this.ipfsDagExport(ipfs, candidate.sourceCid, options) }
+      }
+    }
+  }
+
+  /**
+   * Export a CAR for the passed CID.
+   *
+   * @param {import('./ipfs-client').IpfsClient} ipfs
```
alanshaw, line 133: Does not exist!

packages/cron/test/pins-backup.spec.js
```diff
@@ -0,0 +1,164 @@
...
+import sinon from 'sinon'
+import fetch from '@web-std/fetch'
+
+global.fetch = fetch
+
+const env = {
+  DEBUG: 'backup:pins',
+  ENV: 'dev',
+  CLUSTER_API_URL: 'http://localhost:9094',
+  CLUSTER_IPFS_PROXY_API_URL: 'http://127.0.0.1:9095/api/v0/',
```
alanshaw, line 15: I think you probably want to connect directly to one of the nodes where the content is pinned. If you use the HTTP API via the DNS address you'll end up connecting to a random node and asking it to fetch the DAG (if it doesn't have it). There's no GC on these nodes so we want to avoid copying other nodes in the cluster where the data has not been allocated.

I'm also not certain the `api/v0/dag/export` route has been opened up.

In https://github.com/nftstorage/backup we use libp2p to connect to all nodes in the cluster and transfer via bitswap - only the nodes that are hosting the content will respond.

You could maybe do this with dagula nowadays if you look up which peer the content is hosted on.

alanshaw (changes requested):
You should check with @olizilla that the DB change is similar to what he's expecting to do.

packages/cron/src/jobs/pins-backup.js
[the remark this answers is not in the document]
alanshaw: My thinking was along the lines of that being the max in our ToS and wanting to have a cap at some point on the size of data we're willing to have uploaded/pinned. If you uploaded more than we've said is the max allowed then we shouldn't be obliged to store that.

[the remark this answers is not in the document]
flea89: But we don't have that limit for pins do we? I can't find the ToS for pins, but I might just be missing it?

Assuming there isn't one, there's a chance content bigger than that is stored there and users should expect it to be migrated?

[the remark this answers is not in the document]
flea89: I've commented out the size check part, because, unless we have strong reasons not to, we should move all the files to eipfs.

packages/api/src/car.js
```diff
@@ -6,6 +6,7 @@ import { toString, equals } from 'uint8arrays'
 import { LinkIndexer } from 'linkdex'
```
flea89, line 6: Not sure where these changes are coming from? I wonder if this is coming from a bad merge.
Pretty sure all changes in API package needs to be reverted to whatever we have in main

.github/workflows/cron-backup-pins.yaml
flea89: If I understand this correctly the job has 2 goals:
1. move historical pins to EIPFS
2. keep moving new psa requests to EIPFS until we move to pickup.

For **2** I guess it's ideal to keep moving stuff as promptly as possible (30 min make sense, even less than that?) while we know the first runs of the job will be super slow (since they will have to go through all the historical data).

Isn't a solution to satisfy both words to keep the `schedule` as is and set `concurrency` on the job?

packages/cron/src/jobs/pins-backup.js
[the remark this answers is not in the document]
flea89: Add logging to keep track of enormous dags

packages/api/src/car.js
flea89: Reverted changest to the API

packages/cron/src/jobs/pins-backup.js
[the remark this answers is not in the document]
flea89: I added 2 pieces of logging:
1. every time for whatever reason dag export fails we log the number of bytes read to that point
2. If a successful export is greater than 32TiB, we log it.

flea89:
@olizilla can you please give this a thorough review 🙏 
Let me know if you want to take it from here, or you will need more support from my end as well.

packages/cron/src/jobs/pins-backup.js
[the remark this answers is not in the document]
olizilla: 32TiB or 32 GiB? i think we want to know about pins >= 32 GiB

packages/cron/src/jobs/pins-backup.js
```diff
@@ -0,0 +1,301 @@
...
+
+    // Do not need to await this.
+    ;(async () => {
+      const abortController = new TimeoutController(20_000)
+      let bytesReceived = 0
+
+      let reportInterval
+      const libp2p = await getLibp2p()
+      try {
+        const dagula = await Dagula.fromNetwork(libp2p, { peer })
```
flea89, line 183: Cache instances for same peer location

packages/cron/src/jobs/pins-backup.js
```diff
@@ -0,0 +1,301 @@
...
+        }
+
+        this.log(`✅ Finished reading dag ${cid}. In total ${this.fmt(bytesReceived)}`)
+      } catch (err) {
+        this.log(`🚨 CID: ${cid} errored. Size of the dag is greater than ${bytesReceived} Bytes`)
+        readableStream.destroy(new Error('Dagula errored while reading'))
+        throw (err)
+      } finally {
+        if (bytesReceived > this.MAX_UPLOAD_DAG_SIZE) {
+          this.log(`⚠️ CID: ${cid} dag is greater than ${this.fmt(this.MAX_UPLOAD_DAG_SIZE)}`)
```
flea89, line 208: Add a per batch summary:
- failed cids (and their size)
- Successful (bigger than MAX_UPLOAD_DAG_SIZE)

Remove all unnecessary per cids logging, log just errors by default.

packages/cron/src/jobs/pins-backup.js
```diff
@@ -0,0 +1,301 @@
...
+  /**
+   * Get the size of an file on IPFS
+   * This is so we can limit the size of files we backup
+   *
+   * @param {import('@nftstorage/ipfs-cluster').Cluster} ipfs
+   * @param {import('multiformats').CID} cid
+   * @returns {Promise<number | undefined>}
+   */
+
+  // Given for PIN requests we never limited files size we shouldn't check this. ie.
```
flea89, line 228: Delete stale code

packages/cron/src/jobs/pins-backup.js
```diff
@@ -0,0 +1,301 @@
...
+      if (err.name === 'NotFound') {
+        this.log(`⏳ Uploading to ${url}`)
+
+        const upload = new Upload({
+          client: s3,
+          params: {
+            Bucket: bucketName,
+            Key: key,
+            Body: bak.content,
+            Metadata: { structure: 'Complete' }
```
flea89, line 128: Look into sending checksum for file, reject the upload if the bytes of car don't match the cid

packages/cron/src/jobs/pins-backup.js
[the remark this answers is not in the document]
flea89: yeah, I actually added  32 GiB in the code 👍  (see [here](https://github.com/web3-storage/web3.storage/pull/1844/files#diff-4df8cd6f6a1194703527832b6e009d1515b1d8f708afe9a5af7c31f446f9aadeR34))

packages/cron/src/jobs/pins-backup.js
flea89: I wonder if this is required.
Looking at the headers sent by the client 
```
  'content-type': 'application/xml',
  'content-length': '11985',
  Expect: '100-continue',
  host: '127.0.0.1',
  'x-amz-user-agent': 'aws-sdk-js/3.53.1',
  'user-agent': 'aws-sdk-js/3.53.1 os/darwin/21.6.0 lang/js md/nodejs/16.14.0 md/crt-avail api/s3/3.53.1',
  'amz-sdk-invocation-id': '25110079-acaf-425f-8933-3527fd8366c7',
  'amz-sdk-request': 'attempt=1; max=3',
  'x-amz-date': '20221031T122520Z',
  'x-amz-content-sha256': 'ebd8a0f42b66a7756aaee73e6275d918143525f125137b584e6b079b364a6b5f',
  authorization: 'AWS4-HMAC-SHA256 Credential=minioadmin/20221031/us-east-1/s3/aws4_request, SignedHeaders=amz-sdk-invocation-id;amz-sdk-request;content-length;content-type;host;x-amz-content-sha256;x-amz-date;x-amz-user-agent, Signature=90453c633c07234480f3319eb3c1b058d25b39a077eb3653063165c5bc137722'
}
```

you can see it sends `x-amz-content-sha256` which is the sha256 of the payload and implies the payload is signed (see [docs](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html).

If every chunk is hashed and verified I don't think we need the overall one? Or am I missing something?

flea89: I wonder if this is required.
Looking at the headers sent by the client 
```
  'content-type': 'application/xml',
  'content-length': '11985',
  Expect: '100-continue',
  host: '127.0.0.1',
  'x-amz-user-agent': 'aws-sdk-js/3.53.1',
  'user-agent': 'aws-sdk-js/3.53.1 os/darwin/21.6.0 lang/js md/nodejs/16.14.0 md/crt-avail api/s3/3.53.1',
  'amz-sdk-invocation-id': '25110079-acaf-425f-8933-3527fd8366c7',
  'amz-sdk-request': 'attempt=1; max=3',
  'x-amz-date': '20221031T122520Z',
  'x-amz-content-sha256': 'ebd8a0f42b66a7756aaee73e6275d918143525f125137b584e6b079b364a6b5f',
  authorization: 'AWS4-HMAC-SHA256 Credential=minioadmin/20221031/us-east-1/s3/aws4_request, SignedHeaders=amz-sdk-invocation-id;amz-sdk-request;content-length;content-type;host;x-amz-content-sha256;x-amz-date;x-amz-user-agent, Signature=90453c633c07234480f3319eb3c1b058d25b39a077eb3653063165c5bc137722'
}
```

you can see it sends `x-amz-content-sha256` which is the sha256 of the payload and implies the payload is signed (see [docs](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html).

If every chunk is hashed and verified I don't think we need the overall one? Or am I missing something?

packages/cron/src/jobs/pins-backup.js
flea89: Done!

Default logging (`DEBUG=backupPins:log` ) is quite succinct now, while the job can be run manually passing a more verbose `DEBUG=backupPins:*` through workflow inputs.

Example of default logging:
``` 
❯ DEBUG=backupPins:log npm test --workspace=packages/cron
```
Screenshot 2022-10-31 at 16 24 12

``` 
❯ DEBUG=backupPins:log npm test --workspace=packages/cron
```
Screenshot 2022-10-31 at 16 26 17

packages/cron/src/jobs/pins-backup.js
flea89: Done.

packages/cron/src/jobs/pins-backup.js
flea89: cc @olizilla

joshJarr (approved):

flea89 (approved):

Issue #1779, opened 2022-08-22
website: I can save my payment settings


gobengo:
Motivation:
* issue tracking e2e functionality of saving payment settings, refreshing, and seeing them loaded at /account/payment

gobengo:
first version shipped at https://web3.storage/account/payment/
#feedback-requested
will iterate in further issues

PR #1852, opened 2022-09-06
feat: issues/1779 - I can save my payment settings

gobengo:
Motivation:
* integrate the work required to accomplish this issue: https://github.com/web3-storage/web3.storage/issues/1779
  * backend: https://github.com/web3-storage/web3.storage/pull/1818
  * frontend (via @cmunns): frontend branch to be added here


packages/db/postgres/migrations/023-add-user-customer-table.sql
```diff
@@ -0,0 +1,9 @@
+-- add table to keep track of 0..1 customer id for each user
+CREATE TABLE IF NOT EXISTS user_customer
+(
+  id               BIGSERIAL PRIMARY KEY,
+  user_id          BIGINT NOT NULL UNIQUE REFERENCES public.user (id),
+  customer_id      TEXT NOT NULL UNIQUE
+);
+
+CREATE INDEX IF NOT EXISTS user_customer_user_id_idx ON public.user_customer (user_id);
```
vasco-santos, line 9:
```suggestion
CREATE INDEX CONCURRENTLY IF NOT EXISTS user_customer_user_id_idx ON public.user_customer (user_id);
```

In migration we should run concurrently to not lock anything. In this case, table is not existent yet, so should not be a problem, but also for future reference when adding indexes on already live things

packages/website/pages/account/payment.js
```diff
@@ -2,68 +2,121 @@
...
+//                 <span className="billing-plan-meter-used"></span>
+//               </div>
+//               100GB
+//             </div>
+//           </div>
+//         </div>
+//       )}
+//     </div>
+//   );
+// };
```
vasco-santos, line 45: is this to drop?

packages/db/test/billing.spec.js
```diff
@@ -0,0 +1,41 @@
+/* eslint-env mocha, browser */
+import assert from 'assert'
+import { DBClient } from '../index.js'
+import { token } from './utils.js'
+
+describe('billing db', () => {
+  it('can upsertUserCustomer to associate user with a stripe customer and then getUserCustomer to read', async () => {
+    assert.ok(true)
```
vasco-santos, line 8:
```suggestion
```

vasco-santos (approved):
Great work LGTM!

packages/api/src/utils/billing-types.ts
```diff
@@ -0,0 +1,66 @@
...
+    customer: CustomerId,
+    paymentMethodId: StripePaymentMethodId
+  ): Promise<void>;
+}
+
+export interface Customer {
+  id: string
+}
+
+export interface CustomersService {
```
yusefnapora, line 42: I really like the distinct types for `Customer` vs `User`. It seems like the kind of thing that often gets bundled together needlessly.

yusefnapora (approved):
Sorry it took a minute to get to reviewing this; I started a couple times yesterday but kept getting distracted by other stuff. Anyway, this looks really good to me, and I like the mock implementation with option to use the stripe test account.

It looks like we're not too tied to Stripe, which is nice even if we end up just using them forever.

packages/api/src/utils/billing-types.ts
gobengo: glad it seems sensible. the way things are configured in prod, there's definitely an important distinction because the system of record for Users is in the web3.storage db, but the system of record for the (sometimes) associated customer is stripe.com.

packages/website/pages/account/payment.js
gobengo: I'll make sure this gets removed in the fast-follow ui iteration, which we cut down a bit just to get a subset of it in here. @adamalton heads up, best to just remove code than comment it out. (you can always revert a commit that removes things).

PR #1854, opened 2022-09-07
fix: empty payload is client error

alanshaw:
This is currently raising alerts for on call folks but is a client error.


vasco-santos (approved):
Thanks man

ericssontadashi (approved):
Looks good! Thanks

PR #1871, opened 2022-09-09
fix: /account/payment tiny bugs - use env var on website, fix uncaught error when no default_payment_method in StripeBillingService

gobengo:
Motivation:
* small errors in /account/payment in staging/prod

Why wasn't this caught by testing?
* missing env var - I should have tested the website preview link
* uncaught error - this is tricky. We mock this service in all our tests, so tests wouldn't have caught it. Maybe test code coverage metric could have helped.  We can also later consider running e2e tests against a stripe account in testMode (instead of mocks)


cmunns (approved):

vasco-santos (approved):
LGTM

Issue #1869, opened 2022-09-08
I can choose a storage subscription price


gobengo:
Motivation:
* In #1779, we added a form for customers to enter a credit card paymentMethod
* Now we need to let them create a subscription using that paymentMethod.
  * in prod, we'll use stripe.com to store a [subscription](https://stripe.com/docs/api/subscriptions/update) with paymentMethod, product, price

gobengo:
`PUT /user/payment` will support
```
{
  "method": { "id": "pm_..." },
  "subscription": {
    "storage": { "price": "price_id_..." }
  }
}
```

PR #1878, opened 2022-09-10
feat: I can choose a storage pricing tier (issue 1869)

gobengo:
Motivation:
* #1869 
* kinda like https://trunkbaseddevelopment.com/ but not in 'main', use this branch instead. commit early and often and we can figure out together how to keep the ci checks passing

Parts
* [x] implement api contract stubbed in `userPaymentPut`
    [added test here](https://github.com/web3-storage/web3.storage/pull/1878/commits/8bfd82e89238fda023c10b76ac59358d587e0833#diff-0313c9c004c693d94c846d722a7ea9695c4c40867f3666b2f3706850d47f40eaR138)
    ```
    const desiredPaymentSettings = {
      method: { id: `pm_${randomString()}` },
      subscription: { storage: { price: `price_test_${randomString()}` } }
    }
    ```
* [x] @cmunns ensure /account/payment page can write the pricing tier to the PUT /account/payment api
* [x] @gobengo add quick hack where GET /account/payment?mockSubscription=true will include a mock subscription with storage price - this should unblock frontend fetching/rendering it for dev/testing
    [added here](https://github.com/web3-storage/web3.storage/pull/1878/commits/e121a6b7300b3cb59056b2b8d9b6d2b996441bd1#diff-0313c9c004c693d94c846d722a7ea9695c4c40867f3666b2f3706850d47f40eaR76)
    Opt-in to getting a realistic paymentSettings.subscription:
    * Request `GET /user/payment?mockSubscription=true`
    * Response
        ```
        {
          subscription: { storage: { price: 'price_mock_userPaymentGet_mockSubscription' } }
        }
        ```
* [x] create use case object `saveStorageSubscription` and use from `userPaymentPut`
* [x] can `saveStorageSubscription` with services backed by stripe.com APIs
* [x] @cmunns ensure /account/payment page can read the pricing tier from the GET /account/payment?mockSubscription=true api
* [x] can `getStorageSubscription(customer)`
* [x] `getStorageSubscription(customer) called from api userPaymentGet`
* [x] test for a new user choosing a paid tier for first time
* [x] test for upgrading from one paid tier to another
* [x] test for downgrading from one paid tier to another
* [x] test for downgrading from one paid tier to a free tier


packages/website/components/contexts/plansContext.js
```diff
@@ -1,27 +1,60 @@
-// temporary placeholder for stripe API data
-export const plans = [
+export const sharedPlans = [
```
gobengo, line 1: @cmunns I think I might need to make this driven by the server and not just the ui code.

e.g. on pageload the ui would need to do `GET /billing/plans/` and get back this information before rendering.

That way I can make sure it corresponds to prices/products in stripe that it knows how to bill for

packages/website/components/accountPlansModal/accountPlansModal.js
```diff
@@ -1,12 +1,46 @@
...
-// import Button from '../button/button.js';
-import GradientBackground from '../gradientbackground/gradientbackground';
-import { plans } from '../contexts/plansContext';
+import AddPaymentMethodForm from '../../components/account/addPaymentMethodForm/addPaymentMethodForm.js';
+import { API, getToken } from '../../lib/api';
+
+export async function putUserPayment(pm_id) {
+  const paymentSettings = {
+    method: { id: pm_id },
+    subscription: { storage: { price: `price_pro` } },
```
gobengo, line 15: This isn't going to work. This will reset the price to 'pro' each time.

packages/api/src/utils/stripe.js
```diff
@@ -396,13 +397,196 @@ export function createStripeBillingContext (env) {
...
+  }
+
+  /**
+   *
+   * @param {import('./billing-types').CustomerId} customerId
+   * @param {import('./billing-types').W3PlatformSubscription} subscription
+   * @returns {Promise<CustomerNotFound|void>}
+   */
+  async saveSubscription (customerId, subscription) {
+    const storageStripeSubscription = await this.getStorageStripeSubscription(customerId)
```
gobengo, line 482: This feels a bit gross to check whether there is already a subscription for the customer by going to stripe.com APIs.
1. its slower than checking our db
2. there's a race condition unless we use a db transaction (two concurrent requests will both see no previous subscription so both try to create one)
  * but even if I use the db... our postgrest/supbase setup doesn't support transactions I think. So it's mostly about #1 IMO

Alternatively we could keep track of the subscription in our db through a new [customer_id, subscription_id] 'join' table. But there are some other must-dos to get this ready for working across our separate staging/prod stripe accounts, so I want to do that first.

Is this approveable as is without making a db schema changes to keep track of subscription on our side?

yusefnapora: @gobengo and I just talked about this and agreed that it's a better use of time to make the stripe fixtures consistent between staging & prod envs, given that the race condition is likely to be rare and we can recover from an error with a manual fix if needed. Definitely worth adding the join table when things settle down though.

packages/api/src/utils/stripe.js
```diff
@@ -396,13 +397,196 @@ export function createStripeBillingContext (env) {
...
+   * @param {StripeApiForSubscriptionsService} stripe
+   * @protected
+   */
+  constructor (stripe) {
+    /** @type {StripeApiForSubscriptionsService} */
+    this.stripe = stripe
+    /**
+     * @type {import('./billing-types').SubscriptionsService}
+     */
+    const instance = this // eslint-disable-line
```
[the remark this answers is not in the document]
gobengo: it's one of the few ways to have tsc type-check that your class `implements` an interface. I found a shorter syntax, though, so will switch to that. Also it may not be necessary at all with the `@returns` in the static `create` func

cmunns (approved):
LGTM

packages/website/components/accountPlansModal/accountPlansModal.js
```diff
@@ -1,12 +1,60 @@
...
+import AddPaymentMethodForm from '../../components/account/addPaymentMethodForm/addPaymentMethodForm.js';
+import { API, getToken } from '../../lib/api';
+
+export async function putUserPayment(pm_id, plan_id) {
+  const storage = plan_id ? { price: plan_id } : null;
+  const paymentSettings = {
+    paymentMethod: { id: pm_id },
+    subscription: { storage: storage },
+  };
+  const res = await fetch(API + '/user/payment', {
```
gobengo, line 18: This use of fetch seems to encode the same api request as the other PUT [over here](https://github.com/web3-storage/web3.storage/pull/1878/files#diff-7cc8b048a95005ae2552d91b2be79b342228050c6ef92a204229debb1039f951R12).

It seems like the pattern in this part of the codebase has been to define functions that request the http api in `/packages/website/lib/api.js` and then use them from places like this (and the above link).

@cmunns Will you update the files that trigger this PUT request to both invoke the same funtion in `api.js` that encapsulates how to do the request. That way the 'save payment settings' api request logic is only in once place (e.g. [changes like this i just did](https://github.com/web3-storage/web3.storage/pull/1878/commits/7849b53b7e81ad5621d6085911cc43fd3519046e))

packages/website/pages/_app.js
```diff
@@ -15,7 +15,7 @@ import Footer from '../components/footer/footer.js';
  */
 const App = ({ Component, pageProps }) => {
   const { pathname } = useRouter();
-  const productRoutes = ['/login', '/account', '/account/payment', '/tokens', '/callback'];
+  const productRoutes = ['/login', '/account', '/account/payment', '/account/payment/plans', '/tokens', '/callback'];
```
gobengo, line 18: @cmunns That route doesn't exist eh?

packages/website/pages/account/index.scss
```diff
@@ -206,14 +373,26 @@
...
+  font-size: 1rem;
   display: flex;
-  gap: 1.2rem;
+  flex-direction: column;
+  justify-content: space-between;
+  gap: 0.5rem;
   margin-bottom: 1rem;
-}
+  line-height: 1.5;
+  // padding: 2px 18px;
```
gobengo, line 383: Will you remove lines that are commented out like this but not needed?

packages/website/pages/account/payment.js
```diff
@@ -1,60 +1,35 @@
+// @ts-nocheck
```
gobengo, line 1: Hey this seems like cheating :) @cmunns Are you opposed to using type-checking in this file?

packages/website/pages/account/payment.js
```diff
@@ -1,60 +1,35 @@
...
 
-import { useEffect, useState } from 'react';
+import { useState, useEffect } from 'react';
 import { Elements, ElementsConsumer } from '@stripe/react-stripe-js';
 import { loadStripe } from '@stripe/stripe-js';
 
-import Button from '../../components/button/button.js';
+import PaymentCustomPlan from '../../components/account/paymentCustomPlan.js/paymentCustomPlan.js';
+import PaymentTable from '../../components/account/paymentTable.js/paymentTable.js';
+// import PaymentHistoryTable from 'components/account/paymentHistory.js/paymentHistory.js';
```
gobengo, line 12: please remove any lines that aren't needed instead of commenting

packages/website/pages/account/payment.js
```diff
@@ -72,51 +47,115 @@ const PaymentSettingsPage = props => {
...
+
+  useEffect(() => {
+    if (savedPaymentMethod) {
+      const getPlan = async () => {
+        const userPlan = await getUserPaymentPlan();
+        if (userPlan?.subscription?.storage) {
+          try {
+            await setCurrentPlan(planList.find(plan => plan.id === userPlan.subscription.storage.price));
+          } catch {
+            throw new Error('MISSING PLAN');
```
gobengo, line 73: @cmunns Is this error more useful than the error it's catching, or was it just there from debugging? If it's from debugging, it seems like this error may have less info than the caught one, so it might be best to throw the original, or otherwise include more useful logging or error message

packages/website/pages/account/payment.js
```diff
@@ -72,51 +47,115 @@ const PaymentSettingsPage = props => {
...
+          }
+        } else {
+          setCurrentPlan(planList.find(plan => plan.id === null));
```
gobengo, line 76: @cmunns You could probably unify this with line 71 above like:
```
setCurrentPlan(planList.find(plan => plan.id === userPlan?.subscription?.storage?.price ?? null));
```

packages/api/src/user.js
```diff
@@ -561,38 +561,80 @@ const notifySlack = async (
...
   const requestBody = await request.json()
-  const paymentMethodId = requestBody?.method?.id
+  const paymentMethodId = requestBody?.paymentMethod?.id
   if (typeof paymentMethodId !== 'string') {
     throw Object.assign(new Error('Invalid payment method'), { status: 400 })
   }
-  const method = { id: paymentMethodId }
+  const subscriptionInput = requestBody?.subscription
+  if (typeof subscriptionInput !== 'object') {
+    throw Object.assign(new Error(`subscription must be an object, but got ${typeof subscriptionInput}`), { status: 400 })
```
yusefnapora, line 607: Hadn't seen this pattern for adding "ad-hoc" fields to an `Error` before. Pretty clever :+1:

packages/api/src/utils/logs.js
```diff
@@ -117,6 +117,10 @@ export class Logging {
   }
 
   async postBatch () {
+    if (process.env.NODE_ENV === 'development') {
```
yusefnapora, line 120: good call :slightly_smiling_face:

packages/api/test/stripe.spec.js
```diff
@@ -103,3 +105,171 @@ describe('StripeCustomersService', async function () {
     assert.equal(customer2ForUser2.id, customerForUser2.id, 'should return same customer for same userId2')
   })
 })
+
+describe('StripeSubscriptionsService', async function () {
+  it('can saveSubscription between prices using real stripe.com API', async function () {
+    // ensure stripeSecretKey and use it to construct stripe
+    const stripeSecretKey = process.env.STRIPE_SECRET_KEY
+    if (!stripeSecretKey) {
```
yusefnapora, line 113: that's cool - for some reason it never occurred to me that you could skip a test dynamically like that; I've always just added it to the `it` call if I knew it wasn't going to pass for some reason.

packages/website/components/account/paymentTable.js/paymentTable.js
```diff
@@ -0,0 +1,81 @@
...
+
+const PaymentTable = ({ plans, currentPlan, setPlanSelection, setIsPaymentPlanModalOpen }) => {
+  return (
+    <>
+      {currentPlan && (
+        <p className="billing-content-intro">
+          <span>
+            Your current plan is: <strong>{currentPlan.title}</strong>
+          </span>
+          <small>Billing Cycle: Aug 18 - Sept 18</small>
```
yusefnapora, line 11: This probably shouldn't be hard-coded :stuck_out_tongue:

packages/website/pages/account/payment.js
```diff
@@ -125,16 +164,19 @@ const PaymentSettingsPage = props => {
  * @returns {{ props: import('components/types').PageAccountProps}}
  */
 export function getStaticProps() {
-  const stripePublishableKey = process.env.NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY;
+  const STRIPE_PUBLISHABLE_KEY_ENVVAR_NAME = 'NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY';
+  const stripePublishableKey = process.env[STRIPE_PUBLISHABLE_KEY_ENVVAR_NAME];
   if (!stripePublishableKey) {
-    console.warn(`account payment page missing required process.env.NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY`);
+    throw new Error(
+      `account payment page missing requires truthy stripePublishableKey, but got ${stripePublishableKey}. Did you set env.${STRIPE_PUBLISHABLE_KEY_ENVVAR_NAME}?`
```
yusefnapora, line 171:
```suggestion
      `account payment page requires truthy stripePublishableKey, but got ${stripePublishableKey}. Did you set env.${STRIPE_PUBLISHABLE_KEY_ENVVAR_NAME}?`
```

yusefnapora (approved):
This looks great :100:  with the caveat that I didn't pull it down and try to run it, just read through the diff.

 @gobengo you did a great job testing the stripe API boundary IMO - I especially like the assertions on the number of API calls & verification that they were called with the right args.

Looks really nicely put together overall, great job guys.

packages/api/test/stripe.spec.js
gobengo: (FWIW me too, I think I copypastad the `it.skip()` otherwise I might not have thought about it)

packages/api/src/utils/logs.js
gobengo: It was slowing down my tests! :P

packages/website/components/account/paymentTable.js/paymentTable.js
gobengo: Oy. Good catch. Thank you!

gobengo:
works on prod

PR #1902, opened 2022-09-16
fix: db index upload type

vasco-santos:
We compute metrics in cron metrics based on upload type. As our upload table is getting bigger and bigger, this has become an issue. Adding an index for type to enable fast queries using that row


vasco-santos:
Index/migrations was already applied in prod and metrics cron is running. @ericssontadashi once you have a moment please approve this

ericssontadashi (approved):
LGTM

Issue #1901, opened 2022-09-16
Cron Heartbeat should be based on env


vasco-santos:
Metrics cron job has been failing for days without an alert. That's because staging was working given its lower DB load. They use same heartbeat so staging was just letting opsgeanie know that everything was fine

ericssontadashi:
PR merged

PR #1903, opened 2022-09-16
feat: split pins cron by environments

ericssontadashi:
[1901](https://github.com/web3-storage/web3.storage/issues/1901)


vasco-santos (approved):
LGTM thanks

PR #1908, opened 2022-09-20
feat: send timing info to logtail. Time r2 & s3.

olizilla:
- add timers for `putToR2` and `putToS3`
- tweak logger to send the timers as metadata to logtail so we may chart it.
- fix tsc errors in log.js

This updates the final request log we send to logtail to include a `timers` property, with a key for each timer name and it's duration in `ms`.

```js
{
  message: '',
  dt: '2022-09-20T14:02:42.960Z',
  level: 'info',
  metadata: {
    user: { id: '1' },
    request: {
      url: 'http://127.0.0.1:8787/car',
      method: 'POST',
      headers: [Object],
      cf: [Object]
    },
    cloudflare_worker: {
      version: '7.5.1',
      commit: 'eae75d2366d59b0cf16143723a5af6513d891f9e',
      branch: 'main',
      worker_id: 'krXpoggAwV',
      worker_started: 1663682561602
    },
    timers: { putToR2: 237, putToS3: 464, request: 494 },
    response: { headers: [Object], status_code: 200, duration: 494 }
  }
}
```

License: MIT

### DIFF
--- a/packages/api/src/car.js
+++ b/packages/api/src/car.js
@@ -292,7 +292,7 @@ async function putToS3 (env, key, carBytes, carCid, rootCid, structure = 'Unknow
     Bucket: env.S3_BUCKET_NAME,
     Key: key,
     Body: carBytes,
-    Metadata: { 
+    Metadata: {
       structure,
       rootCid,
       carCid: carCid.toString()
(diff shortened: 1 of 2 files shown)


alanshaw:
I'm not sure about the timers stuff in production - it only increments when async stuff (which, granted is exactly what we're measuring).

...but also you might get errors for multiple timers with the same name.

olizilla:
I'm not wild on this logging abstraction either, but we currently run `envAll` for each request and pass around a new env object with a new Logger on it for each request, so I think it should work, but, I don't love that we have both a stale env and a fresh env object floating around either.

https://github.com/web3-storage/web3.storage/blob/eae75d2366d59b0cf16143723a5af6513d891f9e/packages/api/src/index.js/#L162-L163

olizilla:
Open to other ideas on this too!

alanshaw (approved):

vasco-santos (approved):
LGTM

PR #1913, opened 2022-09-21
fix: bugs saving credit card at /account/payment for the first time

gobengo:
Motivation:
* fix bug discovered while testing on staging https://github.com/web3-storage/web3.storage/pull/1878
  * bug prevented adding a payment method for the first time in both 'add payment form' in lower left of payment page as well as the form in the accountPlanModal
  * I think this was introduced in [this commit](https://github.com/web3-storage/web3.storage/pull/1878/commits/8adff7beead71868cef8f7f2c9d813acab8c708d) after last PR review [approval](https://github.com/web3-storage/web3.storage/pull/1878#pullrequestreview-1112624754)


cmunns (dismissed):

cmunns (commented):
Early adopter is not getting set on null.

cmunns (approved):

PR #1918, opened 2022-09-21
Update enterprise tier questionaire

drewdelano:
Pricing page (pre-login): as a potential custom/enterprise user, I can enter the details for my storage needs


drewdelano:
@gobengo @heyjay44

packages/website/components/enterpriseTierRequest/enterpriseTierRequestModal.js
```diff
@@ -0,0 +1,83 @@
...
+import Modal from 'modules/zero/components/modal/modal';
+import CloseIcon from 'assets/icons/close';
+import Button from 'components/button/button.js';
+import { createEnterpriseTierRequest } from 'lib/api';
+import GradientBackground from 'components/gradientbackground/gradientbackground';
+
+const UserRequestModal = ({ isOpen, onClose }) => {
+  const [requesting, setRequesting] = useState(false);
+
+  async function handleCreateUserRequest(e) {
```
gobengo, line 12: It will be a bit more performant to wrap this with a `useCallback()` hook. https://reactjs.org/docs/hooks-reference.html#usecallback

packages/website/components/enterpriseTierRequest/enterpriseTierRequestModal.js
```diff
@@ -0,0 +1,83 @@
...
+    e.preventDefault();
+    const data = new FormData(e.target);
+
+    const authMethod = data.get('auth-method');
+    const links = data.get('links');
+    const dataVolume = data.get('data-volume');
+    const dataReadTypeAndFrequency = data.get('data-read-type-and-frequency');
+    const additionalInfo = data.get('additional-info');
+
+    if (authMethod && links && dataVolume && dataReadTypeAndFrequency) {
```
gobengo, line 22: I like the idea of checking these in the js even thought the html has `required`.
I think it can be even better by avoiding a situation where it might 'silently fail', which could make debugging quite a pain if things shift over time (e.g. if someone accidentally removes `required` in the html below).

I'd either throw in the unexpected case
```
if (authMethod && links && dataVolume && dataReadTypeAndFrequency) {
  throw new Error(`cannot build enterprise tier request because one of the required fields is falsy`);
}
```

Or if that feels aggressive, at least add an `else` block to this if with a `console.warn` or similar explaining to a future dev why the normal behavior was not attempted.

packages/website/lib/api.js
```diff
@@ -116,6 +116,20 @@ export async function createUnlimitedStorageRequest(authMethod, links, dataVolum
   );
 }
 
+export async function createEnterpriseTierRequest(authMethod, links, dataVolume, dataReadTypeAndFrequency, additionalInfo) {
+  await createUserRequest(
+    'StorageLimitBytes', // TODO: what should this be?
```
gobengo, line 121: It shouldn't be the same as the `createUnlimitedStorageRequest` IMO.

```suggestion
    'EnterpriseTierRequested',
```

gobengo (changes requested):

packages/website/lib/api.js
drewdelano: Sorry, yeah we talked about that earlier and I forgot to push the update

packages/website/components/enterpriseTierRequest/enterpriseTierRequestModal.js
drewdelano: I like the idea of throwing here.  I didn't put too much thought into this as I was just copying from the existing modal.

packages/website/components/enterpriseTierRequest/enterpriseTierRequestModal.js
drewdelano: I'll look into it.  I didn't put too much thought into this as I was just copying from the existing modal.

packages/website/lib/api.js
drewdelano: Cool, I'm going to wind up ripping this out anyway for the email stuff.  So it's about to be completely different.

packages/website/components/enterpriseTierRequest/enterpriseTierRequestModal.js
drewdelano: OK, I think I've fixed it, but I want to leave this comment unresolved until you verify it.

packages/website/components/enterpriseTierRequest/enterpriseTierRequestModal.js
drewdelano: OK, I think I've fixed it, but I want to leave this comment unresolved until you verify it.

packages/website/lib/api.js
drewdelano: OK, I've made a first attempt, but it looks like email is only available from the cron project which is probably going to be problematic.  I can try hoisting this into a shared project, if this cross project referencing isn't good (I don't feel good about it).

drewdelano:
Per meeting, I'm going to switch `Let's Chat` to a mailto link and we can address this later

PR #1919, opened 2022-09-21
feat: add TOS agree before confirm

cmunns:
Screen Shot 2022-09-21 at 3 25 04 PM

### DIFF
--- a/packages/website/components/accountPlansModal/accountPlansModal.js
+++ b/packages/website/components/accountPlansModal/accountPlansModal.js
@@ -55,6 +55,7 @@ const AccountPlansModal = ({
 }) => {
   const [isCreatingSub, setIsCreatingSub] = useState(false);
   const currentPlan = planList.find(p => p.id === planSelection.id);
+  const [hasAcceptedTerms, setHasAcceptedTerms] = useState(false);
   return (
     <div className="account-plans-modal">
       <Modal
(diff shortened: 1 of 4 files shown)


packages/website/components/accountPlansModal/accountPlansModal.js
```diff
@@ -76,10 +80,28 @@ const AccountPlansModal = ({
...
+        <div className="billing-card card-transparent">
+          <div className="billing-terms-toggle">
+            <input
+              type="checkbox"
+              id="agreeTerms"
+              checked={hasAcceptedTerms}
+              onChange={() => setHasAcceptedTerms(!hasAcceptedTerms)}
+            />
+            <label htmlFor="agreeTerms">
+              I agree to the{' '}
```
gobengo, line 92: FWIW instead of the `{' '}` in jsx I think you can just do `&nbsp;` in html

packages/website/components/accountPlansModal/accountPlansModal.js
```diff
@@ -76,10 +80,28 @@ const AccountPlansModal = ({
...
+              <a href="https://web3.storage/terms/" rel="noreferrer" target="_blank">
```
gobengo, line 93: We should probably use a url relative to this page so people see the version of that page from the same source code powering this page.

```suggestion
              <a href="/terms/" rel="noreferrer" target="_blank">
```

gobengo (approved):

PR #1925, opened 2022-09-22
chore: report error cause to sentry

olizilla:
adopt our forked toucan-js so we can start reporting error causes to sentry

i've set up a minimal fork over here with our standard release-please publishing flow: https://github.com/web3-storage/toucan-js

the original PR that i want to get deployed is here https://github.com/robertcepa/toucan-js/pull/104

and we've now got that rolled into our own npm package at https://www.npmjs.com/package/@web3-storage/toucan-js

License: MIT

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5575,6 +5575,99 @@
       "resolved": "packages/tools",
       "link": true
     },
+    "node_modules/@web3-storage/toucan-js": {
+      "version": "2.7.1",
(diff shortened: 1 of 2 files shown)


olizilla:
I applied my patch of the original module and tested it on staging and it works as expected https://github.com/robertcepa/toucan-js/pull/104#issuecomment-1253647742

vasco-santos (approved):
LGTM! This is niceee

PR #1928, opened 2022-09-22
ci: update api workflow to set production url to web3.storage not api…

gobengo:
….web3.storage

Motivation:
* this is a better URL to set for the 'production' environment.


yusefnapora (approved):
:+1:

PR #1930, opened 2022-09-22
chore: rm stripe.com:fixtures stuff from api workflow, it doesn't work

gobengo:
Motivation
* I shouldn't have moved so fast in adding this. It didn't work, and I missed that it didn't work in main, so also release-pleased to prod, and I think now the api workflow will always be broken for api releases, which is gross.
* This ci part isn't critical and I'm on a deadline, so I will improve things later by continuing ci work via [this](https://github.com/web3-storage/web3.storage/pull/1929), but not for a week or two


yusefnapora (approved):
Shame about the build fail, but it happens. I agree disabling for now is the way to go :+1:

PR #1945, opened 2022-09-25
feat: new users are initialized into free tier

gobengo:
* there wasn't previously a way that the api `loginOrRegister` function could determine when a user was created for the first time based on its invocation of `db.upsertUser()`
  * now `db.upsertUser()` returns `inserted` boolean in its return value
* when `loginOrRegister` does an `upsertUser()` that results in a new user insert, then it attempts to initialize the new user, e.g. by initializing their storage subscription
* only after this change is it a valid state for GET /account/payment paymentSettings can now come back like
    ```
    {
        paymentMethod: null,
        subscription: { storage: { price: 'free' } }
    }
    ```
  * Previously, there was never a time where subscription.storage would be non-null and also paymentMethod would be null.
* because of that last thing, the previous react hook setup on /account/payment page didn't work, so I touched that. While I was at it I removed what appeared to be redundant states/fetches.

### DIFF
--- a/packages/api/src/user.js
+++ b/packages/api/src/user.js
@@ -13,7 +13,7 @@ import { pagination } from './utils/pagination.js'
 import { toPinStatusResponse } from './pins.js'
 import { validateSearchParams } from './utils/psa.js'
 import { magicLinkBypassForE2ETestingInTestmode } from './magic.link.js'
-import { CustomerNotFound, getPaymentSettings, isStoragePriceName, savePaymentSettings } from './utils/billing.js'
+import { CustomerNotFound, getPaymentSettings, initializeBillingForNewUser, isStoragePriceName, savePaymentSettings } from './utils/billing.js'
 
 /**
  * @typedef {{ _id: string, issuer: string }} User
@@ -24,8 +24,35 @@ import { CustomerNotFound, getPaymentSettings, isStoragePriceName, savePaymentSe
  */
 
 /**
+ * @typedef {object} IssuedAuthentication
+ * @property {string} issuer
+ * @property {string} publicAddress
+ * @property {string} [email]
+ */
+
+/**
+ * @typedef {(req: Request) => Promise<null|IssuedAuthentication>} RequestAuthenticator
+ */
+
+/**
+ * Context needed to perform new user registration
+ * @typedef {object} UserRegistrationContext
+ * @property {object} magic
+ * @property {object} magic.utils
+ * @property {import('./env').Env['magic']['utils']['parseAuthorizationHeader']} magic.utils.parseAuthorizationHeader
+ * @property {import('./env').Env['MODE']} MODE
+ * @property {object} db
+ * @property {import('./env').Env['db']['upsertUser']} db.upsertUser
+ * @property {import('./env').Env['db']['getUser']} db.getUser
+ * @property {import('./env').Env['DANGEROUSLY_BYPASS_MAGIC_AUTH']} [DANGEROUSLY_BYPASS_MAGIC_AUTH]
+ * @property {RequestAuthenticator} authenticateRequest
+ * @property {import('../src/utils/billing-types').CustomersService} customers
+ * @property {import('../src/utils/billing-types').SubscriptionsService} subscriptions
+ */
+
+/**
  * @param {Request} request
- * @param {import('./env').Env} env
+ * @param {UserRegistrationContext} env
  * @returns {Promise<Response>}
  */
 export async function userLoginPost (request, env) {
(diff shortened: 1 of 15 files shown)


packages/api/test/fixtures/init-data.sql
```diff
@@ -14,11 +14,11 @@ VALUES ('test-pinning-user', 'test-pinning@user.com', 'test-pinning', 'test-pinn
 INSERT INTO public.user ( name, email, issuer, public_address)
 VALUES ('test-pinning-2-user', 'test-pinning2@user.com', 'test-pinning-2', 'test-pinning-2');
 
-INSERT INTO public.user (id, name, email, issuer, public_address)
-VALUES (6, 'test-pinning-and-restriction-user', 'test-pinning-and-restriction@user.com', 'test-pinning-and-restriction', 'test-pinning-and-restriction');
```
gobengo, line 18: Without these change, `db.upsertUser` would somtimes fail with an error that an upsert is trying to insert id=6 where and that violates the uniqueness constraint.

packages/api/src/user.js
```diff
@@ -62,16 +89,53 @@ function createMagicLoginController (env, testModeBypass = magicLinkBypassForE2E
...
+    ...metadata,
+    issuer: metadata.issuer,
+    publicAddress: metadata.publicAddress
+  }
+  return authentication
+}
+
+/**
+ * Initialize a new user that just registered.
+ * @param {object} ctx
```
gobengo, line 111: should be more specific type

packages/website/components/account/paymentTable.js/paymentTable.js
```diff
@@ -72,11 +70,12 @@ const PaymentTable = ({ plans, currentPlan, isEarlyAdopter, setPlanSelection, se
                     </Button>
                   )}
 
-                  {(currentPlan?.id === plan.id || (isEarlyAdopter && plan.id === 'earlyAdopter')) && (
```
gobengo, line 75: @cmunns It didn't seem we needed to check both `isEarlyAdopter` and `plan.id === 'earlyAdopter'`, but I could be missing something.

packages/api/src/user.js
```diff
@@ -62,16 +89,53 @@ function createMagicLoginController (env, testModeBypass = magicLinkBypassForE2E
...
+ * @param {UserRegistrationContext} env
+ * @returns {Promise<{ issuer: string }>}
  */
 async function loginOrRegister (request, env) {
   const data = await request.json()
-  const auth = request.headers.get('Authorization') || ''
 
-  const token = env.magic.utils.parseAuthorizationHeader(auth)
-  const metadata = await (createMagicLoginController(env).authenticate({ token }))
+  const authenticateRequest = env.authenticateRequest || createMagicLinkRequestAuthenticator(env)
```
gobengo, line 134: eventually I want to make this only come from `env`, and probably also be a CompositeRequestAuthenticator(...requestAuthenticators) that tries several in order.
* a la https://docs.djangoproject.com/en/4.1/topics/auth/customizing/#specifying-authentication-backends


That way we can support magic link in addition to other ways of authenticating requests with a standard abstraction. Then we can e.g. stop using magic-link-specific stuff in all our tests. (then we can drop magic when/if appropriate)

packages/api/src/user.js
```diff
@@ -87,8 +151,10 @@ async function loginOrRegister (request, env) {
   if (env.MODE === NO_READ_OR_WRITE) {
     return maintenanceHandler()
   } else if (env.MODE === READ_WRITE) {
-    // @ts-ignore
     user = await env.db.upsertUser(parsed)
+    if (user.inserted) {
+      await initializeNewUser(env, { ...user, id: user.id.toString() })
```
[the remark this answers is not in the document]
yusefnapora: I think you can write a PGSQL function that uses transactions and call it via postgrest. I haven't tried it personally tho.

packages/api/src/user.js
yusefnapora: I agree it makes sense to move this to `env` eventually, especially since `createMagicLinkRequestAuthentication` pulls everything it needs from `env` anyway.

yusefnapora (approved):
Makes sense to me. I think it might be possible to do it transactionally with a DB function, but that could be a rabbit hole...

packages/api/src/user.js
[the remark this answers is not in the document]
gobengo: @yusefnapora In this case, I don't think you can? (could be wrong). In this case, I'd want to BEGIN a transaction and do the upsert, then (in js with a network, not in sql) invoke some stripe apis, and only if the stripe api call succeeds would i COMMIT the transaction. I don't think you could do that in SQL fn alone, unless you can have two functions-- one to BEGIN the transaction, and then the second function async later calls COMMIT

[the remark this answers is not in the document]
yusefnapora: ah yeah, that makes sense. I think it's generally a Bad Thing to hold a transaction open while doing network I/O anyway, so I think the `inserted` flag is a better fit for this anyway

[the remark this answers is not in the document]
gobengo: yeah its a Bad Thing if there's a better way. but idk it also seems like the least bad thing compared to having partial application of the initialization and no way to rollback. It would be a problem under high writes probably, but IMO this part should be fairly low traffic on write.

fwiw i think async transactions isn't always bad thing e.g. https://en.wikipedia.org/wiki/Two-phase_commit_protocol

packages/website/components/account/paymentTable.js/paymentTable.js
gobengo: ah,,,, you're right.

PR #1950, opened 2022-09-26
feat: /account/login?redirect_to=/final/url should redirect to /final/url after authn

gobengo:
Motivation:
* Current if you're logged out, then try to access a page that requires auth like /account/payments, you'll be redirected through the authentication process and end up at /account (not the destination you originally wanted, always /account)

What this does
*  /account/login?redirect_to=/final/url should redirect to /final/url after authn
* /account/payment page getStaticProps redirectTo redirects to `/account/login?redirect_uri=/account/payment` so the /pricing page can link to /account/payment

### DIFF
--- a/packages/website/pages/account/payment.js
+++ b/packages/website/pages/account/payment.js
@@ -181,7 +181,7 @@ const PaymentSettingsPage = props => {
 /**
  * @returns {{ props: import('components/types').PageAccountProps}}
  */
-export function getStaticProps(arg) {
+export function getStaticProps() {
   const STRIPE_PUBLISHABLE_KEY_ENVVAR_NAME = 'NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY';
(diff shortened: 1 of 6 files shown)


packages/website/pages/account/payment.js
```diff
@@ -193,7 +193,7 @@ export function getStaticProps() {
     props: {
       title: 'Payment',
       isRestricted: true,
-      redirectTo: '/login/',
+      redirectTo: '/login/?redirect_uri=/account/payment',
```
drewdelano, line 196: I think this should just be `/account/payment`

gobengo: redirectTo is where the http response will redirect to when the end-user isn't authenticated. This page is served at /account/payment. If redirect_to=/account/payment, it would be an infinite redirect.

drewdelano: Ah ok, I misunderstood what was happening there.

drewdelano (approved):
LGTM -- `redirectTo` works a lot differently than i was expecting 😲

PR #1959, opened 2022-09-27
feat: early adopter UI

cmunns:
- [x] Added TOS to free tier pre-upload 
- [x] Updated enterprise plan form inputs
- [x] Added plan information to account page and removed default limit labels
- [x] Added ui blocking overlay to tiers when tier is early adopter
- [x] Updated tooltip text for tier table
- [x] Remove first and last name boxes from Enterprise Query Form
- [x] Early Adopters Account Page: their current storage limit will displayed and will stay the same as it was before and the progress bar will remain the same
- [x] (Nice to have) Early Adopters Account Page Account Page remove: orange add you cc payment warning above the matrix (Free Tiers continue to see this warning)
- [x] As a new user, on the payments page, when I select Lite or Expert, I must tick yes to “I have read and agree to the **web3.storage Terms of Service**" before being able to confirm  (bolded text is linked to [https://web3.storage/terms/](https://web3.storage/terms/)).
- [x]  As a new user, on the accounts page, when I go to upload my first file, I must tick yes to “I have read and agree to the **web3.storage Terms of Service**" before being able to upload my file (bolded text is linked to [https://web3.storage/terms/](https://web3.storage/terms/))


packages/website/components/account/paymentTable.js/paymentTable.js
```diff
@@ -79,6 +79,15 @@ const PaymentTable = ({ plans, currentPlan, setPlanSelection, setIsPaymentPlanMo
                 </div>
               </div>
             ))}
+            {currentPlan?.id === 'earlyAdopter' && (
+              <p className="early-adopter-ui-block">
+                {/* As an Early Adopter, you already get our lowest storage rate.
```
gobengo, line 84: if this isn't needed, can we remove it?

packages/website/components/account/storageManager/storageManager.js
```diff
@@ -102,29 +100,25 @@ const StorageManager = ({ className = '', content }) => {
...
                 })}
               </span>
-              &nbsp;of <span className="storage-number">{maxSpaceLabel}</span> used
+              {currentPlan?.id === 'earlyAdopter' ? (
+                <>
+                  &nbsp;of <span className="storage-number">{maxSpaceLabel}</span> used
+                </>
+              ) : (
+                <>
+                  &nbsp;of <span className="storage-number">{currentPlan?.base_storage}</span> used
```
gobengo, line 109: in general IMO we should stick with camelCase like `baseStorage` instead of `base_storage`

packages/website/components/accountPlansModal/accountPlansModal.scss
```diff
@@ -1,120 +1,122 @@
...
-    position: absolute;
-    top: 8px;
-    left: -2px;
-    width: calc(100% + 4px);
-    background: $malibu;
-    font-size: 12px;
-    font-weight: 600;
-    letter-spacing: 1px;
+  .billing-plans {
+    // padding: 3rem;
```
gobengo, line 27: if this isn't needed, let's remove it instead of leaving in the comments

packages/website/components/accountPlansModal/accountPlansModal.scss
```diff
@@ -1,120 +1,122 @@
...
+      background: $malibu;
+      font-size: 12px;
+      font-weight: 600;
+      letter-spacing: 1px;
+      color: #fff;
+      padding: 2px 2rem;
+      border-radius: 8px 8px 0 0;
+      text-align: left;
+      text-transform: uppercase;
+      // transform: translateY(-100%);
```
gobengo, line 59: remove comments?

packages/website/components/account/paymentTable.js/paymentTable.js
cmunns: yep, good catch. We were iterating quite fast yesterday :)

packages/website/components/contexts/plansContext.js
```diff
@@ -5,7 +5,7 @@ export const sharedPlans = [
     description: 'For those that want to take advantage of more storage',
     price: '$3/mo',
     base_storage: '15GiB',
-    additional_storage: '$0.20 / GiB',
+    additional_storage: '$0.22 / GiB',
```
gobengo, line 8: ... this implies a change to our stripe fixtures and maybe other backend things. I think it won't matter initially if we dont intend to charge for it, but what the user sees and agrees to does not correspond to how the billing vendor will (eventually) calculate billing.

packages/website/hooks/use-payment.js
```diff
@@ -0,0 +1,48 @@
+import { useState, useEffect } from 'react';
+
+import { userBillingSettings } from '../lib/api';
+import { plansAll } from '../components/contexts/plansContext';
+
+export const usePayment = () => {
```
gobengo, line 6: will you add a comment explaining what this is intended to be used for? How would another dev look at this and decide whether or not it's appropriate for their use case in another part of the codebase?

packages/website/hooks/use-payment.js
```diff
@@ -0,0 +1,48 @@
...
+   * @property {boolean} pinning_api
+   */
+
+  const [hasPaymentMethod, setHasPaymentMethod] = useState(false);
+  const [currentPlan, setCurrentPlan] = useState(/** @type {Plan | null} */ (null));
+
+  useEffect(() => {
+    const getUserPaymentInfo = async () => {
+      const userPlan = await userBillingSettings();
+      console.log(userPlan);
```
gobengo, line 27: `console`

packages/website/hooks/use-payment.js
```diff
@@ -0,0 +1,48 @@
...
+   * @property {string} bandwidth
+   * @property {string} block_limit
+   * @property {string} car_size_limit
+   * @property {boolean} pinning_api
+   */
+
+  const [hasPaymentMethod, setHasPaymentMethod] = useState(false);
+  const [currentPlan, setCurrentPlan] = useState(/** @type {Plan | null} */ (null));
+
+  useEffect(() => {
```
gobengo, line 24: I think doing ajax/state management in effects like this may not scale very far. Other parts of app use [react-query](https://tanstack.com/query/v4/?from=reactQueryV3&original=https://react-query-v3.tanstack.com/)

I don't think it's important to use atm, but heads up you might find it useful in future changes

packages/website/hooks/use-payment.js
```diff
@@ -0,0 +1,48 @@
...
+    const getUserPaymentInfo = async () => {
+      const userPlan = await userBillingSettings();
+      console.log(userPlan);
+      if (userPlan.paymentMethod) {
+        await setHasPaymentMethod(true);
```
gobengo, line 29: I'm not sure we really need state that just stores a boolean and 'has'. You can have state for the whole fetched paymentSettings, and you can derrive (`useMemo`) paymentMethod from that, and you can derrive `hasPaymentMethod` from that.

I find that `useMemo` is easier to reason about and maintain than `useEffect`, so try to use it whenever possible.

packages/website/pages/account/index.js
```diff
@@ -81,6 +81,7 @@ const Account = () => {
       <div className="page-container account-container">
         <h1 className="table-heading">{dashboard.heading}</h1>
         <div className="account-content">
+          {/* ui block here */}
```
gobengo, line 84: remove comment?

packages/website/components/contexts/plansContext.js
heyjay44: It can be kept at $0.20 / GiB - this was my error.

packages/website/pages/account/payment.js
```diff
@@ -136,6 +136,15 @@ const PaymentSettingsPage = props => {
             <h1 className="table-heading">Payment</h1>
           </div>
           <div className="billing-content">
+            {currentPlan?.id === 'free' && !savedPaymentMethod && typeof paymentSettings === 'undefined' && (
+              <div className="add-billing-cta">
+                <p>
+                  You don&apos;t have a paid plan. Please add a credit/debit card and select a plan to prevent storage
```
gobengo, line 142: I think the `&& typeof paymentSettings === 'undefined'` part does anything? `currentPlan.id` should never be 'free' when `typeof paymentSettings === 'undefined'` because `currentPlan` is derrived from `paymentSettings` (see line 113).

If that analysis is right though, then I think the harm is low though, because I think that last part of the `&&` will never matter...

cmunns: It looked like that was being used for "fetching user info", but certainly I meant to evaluate to !== because I don't want this to show while the user is being fetched. In both cases it wouldn't matter because id isn't free you are correct.

packages/website/hooks/use-payment.js
cmunns: Regarding this hook, it was previously merged code that I was just drying up, but I have updated it to be inline with your new approach on payment.js, ultimately it could all come from this hook maybe? but for now I'm not going to touch payment.js

yusefnapora:
I think the TOS thing from my last comment is caused by not setting up the stripe secrets locally. Will try that and do some more poking...

yusefnapora:
Got things working locally - my existing account shows up as "early adopter" and a new address shows "Free" and gives me the TOS prompt :+1:

packages/website/components/account/fileUploader/fileUploader.js
```diff
@@ -74,6 +77,20 @@ const FileUploader = ({ className = '', content, uploadModalState, background })
...
+    setHasAcceptedTerms(e.target.checked);
+    if (e.target.checked) {
+      localStorage.setItem('acceptedTerms', Date.now().toString());
+    } else {
+      localStorage.removeItem('acceptedTerms');
+    }
+  };
+  useEffect(() => {
+    const acceptedTermsLocalStorage = localStorage.getItem('acceptedTerms');
+    console.log(acceptedTermsLocalStorage);
```
yusefnapora, line 90: should probably rm this `console.log` or add some context so it's not just a random timestamp or null

packages/website/components/account/fileUploader/fileUploader.js
```diff
@@ -95,8 +112,23 @@ const FileUploader = ({ className = '', content, uploadModalState, background })
...
+                  web3.storage Terms of Service
+                </a>
+              </label>
+            </div>
+          )}
+
           <Dropzone
-            className="file-uploader-dropzone"
+            className={`file-uploader-dropzone ${
+              !hasPaymentMethod && currentPlan?.id === 'free' && !hasAcceptedTerms && 'disabled'
```
yusefnapora, line 130: not a blocker at all, but when conditional chains get this long I like to pull them out into local vars, e.g. `const dropZoneEnabled = thing && otherThing`.

yusefnapora (approved):
I played around a bit, and everything seems to work as described when I run locally and nav to `/account/payment` with an "early adopter" account and a "free" account.

I left a couple suggestions, but I don't think there's anything blocking or unsafe to merge in. BTW, the runtime error about pinning access shows up for me on main also, so I don't think it's related to this branch.

PR #1960, opened 2022-09-27
fix: add blog link with navigation fix

johnathonroach:
This PR adds a link to the new blog and addresses a formatting issue in the docs navigation

### DIFF
--- a/packages/website/modules/docs-theme/index.scss
+++ b/packages/website/modules/docs-theme/index.scss
@@ -10,6 +10,10 @@
   background-size: cover;
   word-break: break-word;
 
+  #navigation {
+    word-break: normal;
+  }
+
   #footer_squiggle {
     width: 25rem;
     left: -6.5rem;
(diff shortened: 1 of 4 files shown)


packages/website/modules/docs-theme/index.scss
```diff
@@ -10,6 +10,10 @@
   background-size: cover;
   word-break: break-word;
```
johnathonroach, line 11: This style was causing the issue with the nav on the docs page.

packages/website/modules/docs-theme/index.scss
```diff
@@ -10,6 +10,10 @@
...
 
+  #navigation {
```
johnathonroach, line 13: This CSS targets the navigation element setting the word-break to normal / no break.

JeffLowe (approved):
Checked on desktop and mobile (logged in and out state).  The issue on docs page causing the line breaks on nav items does not occur.  Issue appears resolved.

vasco-santos (approved):
LGTM thanks

PR #1974, opened 2022-09-29
fix: yank out faulty browser detection script

e-schneid:


jchris (approved):
thanks for removing this

alanshaw (approved):
Thank you 🙏

PR #1976, opened 2022-09-29
fix: (wip) add custom upsert_user function to distinguish new vs updated

yusefnapora:
This is a start at fixing the issue with the new `inserted` field returned by `upsertUser`.

The current implementation uses `xmax == 0` to mean the row was newly inserted, but this doesn't seem to work in production - all newly created users have a non-zero `xmax` in the DB.

This adds an `upsert_user` PGSQL function that does a `SELECT (COUNT(id) = 0) into inserted FROM public.user WHERE issuer = _issuer;` before doing the `INSERT`, which should set `inserted` to true only if there's no existing row with that `issuer`.

I haven't done extensive testing yet, but it seems to pass the existing `userLoginPost` test that checks whether new users get the free tier.

TODO:
- [ ] add migration sql file
- [ ] explicitly test the `upsert_user` function

### DIFF
--- a/packages/db/db-client-types.ts
+++ b/packages/db/db-client-types.ts
@@ -16,6 +16,7 @@ export type UpsertUserOutput = {
   id: number
   // whether the upsert inserted a new record (if falsy, it was updated)
   inserted: boolean
+  inserted_at: string
   issuer: string
 }
 
--- a/packages/db/index.js
+++ b/packages/db/index.js
@@ -106,32 +106,24 @@ export class DBClient {
   async upsertUser (user) {
     /** @type {{ data: definitions['user'], error: PostgrestError }} */
     const { data, error } = await this._client
-      .from('user')
-      .upsert(
-        {
-          name: user.name,
-          picture: user.picture,
-          email: user.email,
-          issuer: user.issuer,
-          github: user.github,
-          public_address: user.publicAddress
-        },
-        {
-          onConflict: 'issuer',
-          returning: 'representation'
-        }
-      )
-      .select('id,xmax,issuer')
-      .single()
+      .rpc('upsert_user', {
+        _name: user.name,
+        _picture: user.picture ?? '',
+        _email: user.email,
+        _issuer: user.issuer,
+        _github: user.github ?? '',
+        _public_address: user.publicAddress
+      })
+
+    const userData = data[0]
 
     if (error) {
       throw new DBError(error)
     }
-    const inserted = Boolean(data.xmax === '0')
     return {
-      id: data.id,
-      inserted,
-      issuer: data.issuer
+      id: userData.id,
+      inserted: userData.inserted,
+      issuer: userData.issuer
     }
   }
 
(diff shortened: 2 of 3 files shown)


packages/db/db-client-types.ts
```diff
@@ -16,6 +16,7 @@ export type UpsertUserOutput = {
   id: number
   // whether the upsert inserted a new record (if falsy, it was updated)
   inserted: boolean
+  inserted_at: string
```
gobengo, line 19: The inputs use `camelCase`, so perhaps the outputs should too? (i.e. `insertedAt`). IMO style nit and ez to change later to nonblocking

packages/db/index.js
```diff
@@ -106,32 +106,24 @@ export class DBClient {
...
     if (error) {
       throw new DBError(error)
     }
-    const inserted = Boolean(data.xmax === '0')
     return {
-      id: data.id,
-      inserted,
-      issuer: data.issuer
+      id: userData.id,
+      inserted: userData.inserted,
```
gobengo, line 125: If I'm reading this right, it seems like this property should be `inserted_at`? https://github.com/web3-storage/web3.storage/pull/1976/files#r984411987

packages/db/db-client-types.ts
gobengo: Oh.... actually do we really need this `inserted_at` col in the type at all? I don't see it in the return value? https://github.com/web3-storage/web3.storage/pull/1976/files#diff-ab5fce4fc61478316713f5b4fad807e7dac6e2e2100858f37bfb0b084543814fR123

I think I saw in slack that you tried another approach before involving a timestamp, so maybe this is a remanant of that that can be safely removed?

packages/db/postgres/functions.sql
```diff
@@ -415,3 +415,37 @@ FROM cargo.aggregate_entries ae
...
+  SET 
+    name = EXCLUDED.name,
+    picture = EXCLUDED.picture,
+    email = EXCLUDED.email,
+    github = EXCLUDED.github,
+    public_address = EXCLUDED.public_address
+  RETURNING u.id, u.issuer, inserted;
+
+END
+$$;
```
gobengo, line 451: VERY IMPRESSIVE SQLING TYSM! (I thought my figuring out the xmax thing was cool, this is next level)

packages/db/db-client-types.ts
yusefnapora: oh yeah, that's just cruft from the initial approach - thanks for catching it

jsdevel (approved):

alanshaw (approved):
LGTM

packages/db/postgres/functions.sql
```diff
@@ -415,3 +416,37 @@ FROM cargo.aggregate_entries ae
...
+$$
+#variable_conflict use_column
+DECLARE
+  inserted BOOLEAN;
+
+BEGIN
+  SELECT (COUNT(id) = 0) into inserted FROM public.user WHERE issuer = _issuer;
+
+  RETURN QUERY
+  INSERT INTO public.user AS u (name, picture, email, issuer, github, public_address)
```
drewdelano, line 440: probably not a big deal due to volume and how fast they would have to race:

but if called twice for the same user a milisecond or two apart won't it result in returning inserted = true for both?

yusefnapora: I think it's safe, since the `BEGIN` and `END` makes sure both statements happen in the same transaction.

PR #1977, opened 2022-09-30
fix: remove storage economics page

alanshaw:
Copy is currently invalid so removing until replacement is made.


alanshaw (commented):
LGTM

PR #1992, opened 2022-10-03
fix: top align testimonials

drewdelano:
https://www.notion.so/move-leftmost-testimonial-up-to-align-with-others-4134ff9af980487bb6b88b82b41cc684


drewdelano:
We decided to remove it instead

PR #1996, opened 2022-10-04
fix: "get started" only partially clickable

drewdelano:
https://www.notion.so/Make-get-started-on-Pricing-Page-more-easily-clickable-it-doesn-t-light-up-unless-your-cursor-hi-0f1e6f42515d46b88daca24ae1468565

The issue was the image was overlaying the button:
image

Proposed fix:
Disable `pointer-events` for the image so the browser acts like the image doesn't exist when you click where they overlap.

Post fix:
image


cmunns (approved):
Love pointer-events: none workarounds!

PR #2007, opened 2022-10-05
fix: missing prop for cc form in confirm modal

cmunns:
fixes this cc input
Screen Shot 2022-10-05 at 11 04 51 AM

### DIFF
--- a/packages/website/components/account/storageManager/storageManager.js
+++ b/packages/website/components/account/storageManager/storageManager.js
@@ -1,6 +1,7 @@
 import clsx from 'clsx';
 import filesz from 'filesize';
 import { useMemo, useEffect, useState, useRef } from 'react';
+import Link from 'next/link';
 
 import { useUser } from 'components/contexts/userContext';
 import { elementIsInViewport } from 'lib/utils';
@@ -85,6 +86,11 @@ const StorageManager = ({ className = '', content }) => {
 
   return (
     <div ref={storageManagerRef} className={clsx('section storage-manager-container', className)}>
+      <Link href={'account/payment'} passHref>
+        <a href={'account/payment'} className="storage-manager-payment-link">
+          {currentPlan?.id !== 'pro' ? 'Want more storage? Upgrade your plan here!' : 'Plans & Payment'}
+        </a>
+      </Link>
       <h6>Your Plan: {currentPlan?.title}</h6>
       <div className="storage-manager-space">
         <div className="storage-manager-used">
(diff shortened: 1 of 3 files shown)


packages/website/components/account/storageManager/storageManager.js
```diff
@@ -85,6 +86,11 @@ const StorageManager = ({ className = '', content }) => {
 
   return (
     <div ref={storageManagerRef} className={clsx('section storage-manager-container', className)}>
+      <Link href={'account/payment'} passHref>
+        <a href={'account/payment'} className="storage-manager-payment-link">
+          {currentPlan?.id !== 'pro' ? 'Want more storage? Upgrade your plan here!' : 'Plans & Payment'}
```
gobengo, line 91: Why is this the logic? We only want to show it if you're not on pro? Why not show it if you're on 'free' or 'null'? If it's not obvious, maybe explain with a comment or good variable name for the boolean? e.g. if we change the tier names or number of tiers later, how will we know to remember to change this or whether or not it makes sense to?

gobengo:
It's hard for me to tell what the motivation of this change is, i.e. what is it trying to fix? If that's described elsewhere, will you include it in the PR description? If not, will you describe a bit more what the intention of this change is, otherwise I'm not sure how to evaluate it.

packages/website/components/account/storageManager/storageManager.js
cmunns: If pro is the highest plan there is nothing to upgrade. It's honestly trivial we can remove it and just keep the same link text since pricing is in flux.

cmunns:
@gobengo here is the notion for the link text. https://www.notion.so/1c6ce217e4be45a28a59a6dfe5faac59?v=adbff0a36eca4d0486f66780c7417721&p=18f3d0f7d7624aefb37bff4d0175802b&pm=s

One of the downfalls of not being to link an issue quickly. Next time if I push a second commit I'll make sure I add the context from notion as well.

packages/website/components/account/storageManager/storageManager.js
```diff
@@ -85,6 +86,11 @@ const StorageManager = ({ className = '', content }) => {
 
   return (
     <div ref={storageManagerRef} className={clsx('section storage-manager-container', className)}>
+      <Link href={'account/payment'} passHref>
+        <a href={'account/payment'} className="storage-manager-payment-link">
```
drewdelano, line 90: I thought you only needed href on the outer link

drewdelano (approved):
LGTM other than the possible double href issue

packages/website/components/account/storageManager/storageManager.js
drewdelano: it works fine without the second href, but causes a linter error.  let's leave this as-is for now.

PR #2010, opened 2022-10-06
feat: display name and email in stripe

e-schneid:
Currently when customers are added to stripe, their name and email are missing.  This makes it difficult for the business to evaluate issues, since we'd need to look at our own database to determine which web3.storage user is associated with a stripe customer.

### DIFF
--- a/packages/api/src/utils/billing-types.ts
+++ b/packages/api/src/utils/billing-types.ts
@@ -42,7 +42,7 @@ export interface Customer {
 }
 
 export interface CustomersService {
-  getOrCreateForUser(user: BillingUser): Promise<Customer>
+  getOrCreateForUser(user: BillingUser, userCreationOptions?: UserCreationOptions): Promise<Customer>
 }
 
 export type StoragePriceName = 'free' | 'lite' | 'pro'
@@ -101,4 +101,9 @@ export type PaymentSettings = {
 export interface UserCustomerService {
   getUserCustomer: (userId: string) => Promise<null|{ id: string }>
   upsertUserCustomer: (userId: string, customerId: string) => Promise<void>
+}
+
+export interface UserCreationOptions {
+  email?: string
+  name?: string
 }
--- a/packages/api/src/user.js
+++ b/packages/api/src/user.js
@@ -16,7 +16,7 @@ import { magicLinkBypassForE2ETestingInTestmode } from './magic.link.js'
 import { CustomerNotFound, getPaymentSettings, initializeBillingForNewUser, isStoragePriceName, savePaymentSettings } from './utils/billing.js'
 
 /**
- * @typedef {{ _id: string, issuer: string }} User
+ * @typedef {{ _id: string, issuer: string, name?: string, email?: string }} User
  * @typedef {{ _id: string, name: string }} AuthToken
  * @typedef {{ user: User, authToken?: AuthToken }} Auth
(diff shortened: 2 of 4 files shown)


packages/api/src/utils/billing-types.ts
```diff
@@ -80,7 +80,9 @@ export interface SubscriptionsService {
 }
 
 export interface BillingUser {
```
gobengo, line 82: I don't think we should have the name and email required in here because not all billing things need it. A lot of times all that's needed is the ID. If we make it required, certain steps will either require passing mocked data or incur an unnecessary read from the db or session.

Consider the proposed solution I [wrote here](https://www.notion.so/bizops-can-see-name-and-email-in-stripe-com-dashboard-for-customers-9e263386e79746fabffaf38d66a0dfc6#1df8960aa93642e496c7cbd7c908816a) that @jsdevel acknowledged in notion?

packages/api/test/user.spec.js
```diff
@@ -448,7 +448,13 @@ describe('userLoginPost', function () {
     // now ensure it has desired subscription
     const gotUser = await env.db.getUser(user1Authentication.issuer, {})
     const gotSubscription = await env.subscriptions.getSubscription(
-      (await env.customers.getOrCreateForUser({ id: gotUser._id })).id
+      (
+        await env.customers.getOrCreateForUser({
+          id: gotUser._id,
+          name: 'test user',
```
gobengo, line 454: I don't think we should have `name` or `email` be required in all `getOrCreateForUser` invocations. e.g. for this test is it really needed?

packages/db/index.js
```diff
@@ -121,7 +121,9 @@ export class DBClient {
     return {
       id: userData.id,
       inserted: userData.inserted,
-      issuer: userData.issuer
+      issuer: userData.issuer,
+      name: userData.name,
+      email: userData.email
```
gobengo, line 126: Do these fields actually come back from the sql query? I don't think they will based on the underlying `upsert_user` [https://github.com/web3-storage/web3.storage/blob/main/packages/db/postgres/functions.sql#L449]

(I think if you wrote a test assertion that they are defined after calling `upsertUser`, it might fail).

Regardless, though, can we do this in a way that doesn't involve always returning this from upsertUser? It's an upsert, the caller of the function just passed in those fields, so it seems like there's no point in making the return type include them (and the underlying query more advanced).

gobengo:
When you make a PR will you write an issue description that includes your motivation? e.g. a description of what you expect the code will do or a link to an issue or backlog item? It helps in evaluating the patch 🙏

packages/api/src/utils/billing-types.ts
e-schneid: Sorry @gobengo , I hadn't added reviewers since I'd opened it up to initiate discussion with @jsdevel (I'll take another approach to this next time 😆 ).  I've since reviewed the notion doc and hopefully the changes are more in line with your planning.

packages/db/index.js
e-schneid: It was coincidental that this worked at all, since the name and email were coming from auth instead of the returned value here.  I've since updated this PR to align with your notion doc.  Thanks!

packages/api/test/user.spec.js
e-schneid: Updated.

yusefnapora (approved):

PR #2012, opened 2022-10-07
fix: login redirect should preserve query parameters

drewdelano:
https://www.notion.so/Ensure-redirect-from-Pricing-page-takes-you-to-post-login-modal-of-selected-tier-e9fb5131d7b2486596de0d8c17f77161

This PR is a follow up to a conversation BenGo (not tagging him here to avoid notification noise) and I had a couple of weeks ago.  The high level summary is:

Something like the pricing page could use a `requiresAuth: true` behavior.  This would function similar to the `redirectTo: '...'` behavior, except it would capture the url and handle the redirecting automatically.  This is important for tiered pricing because otherwise we can't include which tier the customer selected on the public pricing page before we redirect to the authenticated accounts payment page because the `redirectTo` must be static.


drewdelano:
FYI - @gobengo @heyjay44 starting work here.  updated notion too.

jsdevel (approved):

PR #2020, opened 2022-10-10
chore: use db ids > JS MAX_SAFE_INTEGER in test

### LINKED ISSUE #1367: Set SQL auto-incremement sequences to Javascript-unsafe integers for tests
A recurring gotcha is that we get bitten by the Javascript maximum integer when using SQL BIGINT columns, we unwittingly get hold of them as a `Number` at some point, which if the value is above the JS maximum integer value causes it to overflow and give us a different value.

We've got various measures and techniques in place to avoid this, but when running the tests the DB's auto-increment columns are usually initialised to `1`, so any bugs only surface when we deploy to staging, or worse, to production.

If we could initialise the sequences for these columns to a high value when running the tests, it would catch these bugs much earlier and avoid pain.

yusefnapora:
We got hit by this recently during the rollout of tiered pricing, so I decided to pick it up this morning. Closes #1367


vasco-santos:
Thanks @yusefnapora ❤️

PR #2028, opened 2022-10-12
feat: log terms of service acceptance

e-schneid:
This adds terms of service acceptance logging.  When a customer accepts the terms of service, we then send `web3.storage-tos-v1` as the agreement value to the API.  This then gets stored in a table with a foreign key of the userId.


packages/api/src/utils/billing.js
```diff
@@ -1,9 +1,19 @@
...
 /**
  * @typedef {import('./billing-types').StoragePriceName} StoragePriceName
  */
 
+/**
+ * @param {import('./billing-types').Agreement} agreement
+ * @returns {boolean}
+ */
+function isAgreement (agreement) {
+  return agreement === 'web3.storage-tos-v1'
```
jsdevel, line 14: right now this is fine because we aren't storing the ToS in the DB (need to have a future conversation about this cc @gobengo ).

packages/api/src/utils/billing.js
```diff
@@ -85,10 +100,16 @@ export function createMockUserCustomerService () {
   const upsertUserCustomer = async (userId, customerId) => {
     userIdToCustomerId.set(userId, customerId)
   }
+
+  const createUserTosAgreement = async (userId, agreement) => {
+    // noop
```
jsdevel, line 105: for my own sanity, are these needed?

e-schneid: It's possible I'm bloating the CustomerService type.  Maybe in general Terms of Service should be handled in its own type, since it doesn't really have anything to do with the stripe customer, and it's really not even strictly a billing thing in general (although we're applying them to only billing at the moment).  I'll consider a quick refactor

e-schneid: refactor done

packages/api/src/errors.js
```diff
@@ -117,6 +117,24 @@ export class MagicTokenRequiredError extends HTTPError {
...
 MagicTokenRequiredError.CODE = 'ERROR_MAGIC_TOKEN_REQUIRED'
 
+export class TosAgreementRequiredError extends HTTPError {
+  constructor (msg = 'Missing Terms of Service agreement') {
+    super(msg, 400)
+    this.name = 'TosAgreementRequired'
+    this.code = TosAgreementRequiredError.CODE
+  }
+}
+TosAgreementRequiredError.CODE = 'ERROR_TOS_AGREEMENT_REQUIRED'
```
gobengo, line 127: No need to change. I see this is how a lot of the rest of the file does it. But for my edification, is this different than including `static CODE = 'ERROR_TOS_AGREEMENT_REQUIRED'` in the class definition? I don't think so but could be wrong.

packages/api/src/user.js
```diff
@@ -700,13 +700,15 @@ export async function userPaymentPut (request, env) {
       billing: env.billing,
       customers: env.customers,
       user: { ...request.auth.user, id: request.auth.user._id },
-      subscriptions: env.subscriptions
+      subscriptions: env.subscriptions,
+      termsOfService: env.termsOfService
```
gobengo, line 704: In the long run, what do you think of using this service and table for 'agreements' or 'consents' beyond just terms of service agreements? It might be nice to model things that way. Curious for your thoughts, we can always iterate towards that after this PR if it makes sense to all.

packages/api/src/utils/billing-types.ts
```diff
@@ -91,19 +91,27 @@ export interface BillingEnv {
...
+  agreement?: Agreement
 }
 
 export interface UserCustomerService {
   getUserCustomer: (userId: string) => Promise<null|{ id: string }>
   upsertUserCustomer: (userId: string, customerId: string) => Promise<void>
 }
 
+export interface TermsOfServiceService {
+  createUserTosAgreement: (userId: string, agreement: Agreement) => Promise<void>
```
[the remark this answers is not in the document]
gobengo: where I'm coming from is it may not be long before there are agreements like 'privacy policy' or something that might not be exactly 'terms of service'. afaict they are similar but have important legal distinctions. But still would be nice to be able to share the agreement audit log across both. (which we can totally do without renaming anything, which is something I like about how you've already implemented it!)

packages/api/src/utils/billing.js
```diff
@@ -1,24 +1,40 @@
...
 
+import { InvalidTosAgreementError, TosAgreementRequiredError } from '../../src/errors.js'
+
 /**
  * @typedef {import('./billing-types').StoragePriceName} StoragePriceName
  */
 
+/**
+ * @param {import('./billing-types').Agreement} agreement
+ * @returns {boolean}
```
gobengo, line 11: It might be nice to type this as a type guard for tsc. https://goulet.dev/posts/type-guard-in-jsdoc/

I'm not positive this exact suggestion is the right syntax with the `import`, but might work. Otherwise you can use a typedef to alias agreement into this file like

```
/**
 * @typedef {import('./billing-types').Agreement} Agreement
 */
```

and then do 
```
 * @returns {agreement is Agreement}
```

```suggestion
 * @returns {agreement is import('./billing-types').Agreement}
```

If that's not aligned with the point of this function, ignore

packages/api/src/utils/billing-types.ts
```diff
@@ -91,19 +91,27 @@ export interface BillingEnv {
   billing: BillingService
   customers: CustomersService
   subscriptions: SubscriptionsService
+  termsOfService: TermsOfServiceService
 }
 
 export type PaymentSettings = {
   paymentMethod: null | PaymentMethod
   subscription: W3PlatformSubscription
+  agreement?: Agreement
```
gobengo, line 100: Why not make it required? Maybe there is a use case for it, but I can't necessarily think of one. Could be nice to err on the side of making it explicit everywhere.

packages/api/src/utils/billing.js
```diff
@@ -1,24 +1,40 @@
...
  * @param {import('./billing-types').BillingUser} ctx.user
  * @param {object} paymentSettings
  * @param {Pick<import('./billing-types').PaymentMethod, 'id'>} paymentSettings.paymentMethod
  * @param {import('./billing-types').W3PlatformSubscription} paymentSettings.subscription
+ * @param {import('./billing-types').Agreement} paymentSettings.agreement
  * @param {import('./billing-types').UserCreationOptions} [userCreationOptions]
  */
 export async function savePaymentSettings (ctx, paymentSettings, userCreationOptions) {
-  const { billing, customers, user } = ctx
+  if (!paymentSettings.agreement) throw new TosAgreementRequiredError()
```
gobengo, line 32: I like how you added these nice semantic errors here.
On the other hand, If this is always required, can we just make it required in the `PaymentSettings` type so it's not even compilable to call this function with an invalid value and get a runtime error?

(the caller of `savePaymentSettings` would then be responsible for getting an appropriate value, and could still throw the exact errors you did, but it might have more specific context to use to throw the error e.g. with a really useful error message).

packages/api/src/utils/billing.js
```diff
@@ -70,7 +86,7 @@ export async function getPaymentSettings (ctx) {
 }
 
 /**
- * @returns {import('./stripe').UserCustomerService & { userIdToCustomerId: Map<string,string> }}
+ * @returns {import('./stripe').UserCustomerService & import('./stripe').TermsOfServiceService & { userIdToCustomerId: Map<string,string> }}
```
gobengo, line 89: ooh can you instead just make a separate `createMockTermsOfServiceService` fn?
The only reason I added the extra `&` here before was to make the mock a little introspectable e.g. for some unit testing. I think it'd be nice to keep the two mock service types separate if possible (e.g. so consumers of these functions can do [ISP](https://en.wikipedia.org/wiki/Interface_segregation_principle))

packages/api/src/utils/billing.js
```diff
@@ -1,24 +1,40 @@
...
  * @param {import('./billing-types').UserCreationOptions} [userCreationOptions]
  */
 export async function savePaymentSettings (ctx, paymentSettings, userCreationOptions) {
-  const { billing, customers, user } = ctx
+  if (!paymentSettings.agreement) throw new TosAgreementRequiredError()
+  if (!isAgreement(paymentSettings.agreement)) throw new InvalidTosAgreementError()
+
+  const { billing, customers, user, termsOfService } = ctx
   const customer = await customers.getOrCreateForUser(user, userCreationOptions)
+  await termsOfService.createUserTosAgreement(user.id, paymentSettings.agreement)
```
gobengo, line 37: in practice the calls to customers and billing are going to hit the stripe apis so be slow compared to the ones that just use the sql db. Don't have to do it now, but eventually we might want to do some stuff in parallel here? I think technically there are three parallelizable steps. wdyt in theory? Don't have to do now but good to have in mind if this starts feeling slow.

```
function 
await Promise.all([
  termsOfService.createUserTosAgreement(user.id, paymentSettings.agreement),
  billing.savePaymentMethod(customer.id, paymentSettings.paymentMethod.id),
  (async () => {
    const customer = await customers.getOrCreateForUser(user, userCreationOptions);
    await ctx.subscriptions.saveSubscription(customer.id, paymentSettings.subscription)
  })(),
]);
```

packages/api/src/utils/stripe.js
```diff
@@ -429,7 +437,8 @@ export function createStripeBillingContext (env) {
   return {
     billing,
     customers,
-    subscriptions
+    subscriptions,
+    termsOfService
```
gobengo, line 441: Am I correct in understanding that the TermsOfServiceService doesn't depend at all on stripe?

If so, I think it's best that we don't include it in the return of this function. I think the way our tests get set up almost always have access to a sql db, so we don't even need to use a mock a lot of the times e.g. when integration testing the http api layer.

The main reason I used so many mocks for the stripe-specific stuff is because I didn't want our tests to have to invoke the stripe.com APIs (so I used mocks), whereas I think it's frequently okay to rely for a lot of our tests to rely on the sql db.

We can rename some things if it stop being intuitive, but the way I'd think about it is maybe the TermsOfService stuff isn't necessarily about 'BillingContext' as much as a separate kind of context that's a bit orthogonal (but often used together). The distinction being that everything in the BillingContext can be fulfilled by stripe.com, whereas the AgreementsContext doesn't use stripe at all and just uses a sqldb. (if in practice AgreementsContext only has one service in it, we might not need a 'context' interface at all)

packages/api/test/user-payment.spec.js
```diff
@@ -154,11 +155,46 @@ describe('userPaymentPut', () => {
...
+    const customers = createMockCustomerService()
+    const env = {
+      ...createMockBillingContext(),
+      billing,
+      customers
+    }
+    assert.rejects(
+      async () => await userPaymentPut(request, env),
+      /Missing Terms of Service/
+    )
```
gobengo, line 173: Nice! Thanks for testing this stuff. Love a good regression test ensuring there is a grokkable error message 🎉

packages/api/test/user-payment.spec.js
```diff
@@ -154,11 +155,46 @@ describe('userPaymentPut', () => {
...
+    const billing = createMockBillingService()
+    const customers = createMockCustomerService()
+    const env = {
+      ...createMockBillingContext(),
+      billing,
+      customers
+    }
+    assert.rejects(
+      async () => await userPaymentPut(request, env),
+      /Invalid Terms of Service/
```
gobengo, line 189: wdyt of this mentioning 'agreement' instead of 'Terms of Service' since that's what the field in the paymentSettings object calls it? e.g. it could be an error message like `Unable to save Payment Settings due to unknown agreement value "{paymentSettings.agreement}"`

Also is there a way for these tests we can ensure that the error being thrown will lead to an HTTP 400 response status code? That way even if the frontend doesn't know about the exact underlying error, it can at least use the http response status conventions to try to decide what to do or how to format a generic error for the end-user.

packages/db/index.d.ts
```diff
@@ -70,6 +70,7 @@ export class DBClient {
   createUserTag(userId: string, tag: UserTagInput): Promise<boolean>
   getUserTags(userId: string): Promise<UserTagInfo[]>
   upsertUserCustomer(userId: string, customerId: string): Promise<void>
+  createUserTosAgreement(userId: string, agreement: string): Promise<void>
```
gobengo, line 73: what if we just called it `createUserAgreement` ?

packages/db/index.js
```diff
@@ -1295,6 +1295,30 @@ export class DBClient {
     }
   }
 
+  /**
+   * @param {string} userId
+   * @param {string} agreement
```
gobengo, line 1300: Will you add a `@returns` ?

packages/api/src/utils/billing-types.ts
gobengo: When I got to the website I remembered the use case. If the end-user is only entering their credit card for later use but not necessarily as part of choosing a storage subscription they might get charged for, we don't always render the tos checkbox (I think?).

https://github.com/web3-storage/web3.storage/pull/2028/files#r998202155

So I'm in favor of leaving this optional insofar as we still need to accomodate that UI use case.

packages/website/components/account/addPaymentMethodForm/addPaymentMethodForm.js
```diff
@@ -49,7 +50,7 @@ const AddPaymentMethodForm = ({ setHasPaymentMethods, setEditingPaymentMethod, c
         if (error) throw new Error(error.message);
         if (!paymentMethod?.id) return;
         const currStorageSubscription = planIdToStorageSubscription(currentPlan);
-        await userBillingSettings(paymentMethod.id, currStorageSubscription);
+        await userBillingSettings(constants.TERMS_OF_SERVICE_VERSION, paymentMethod.id, currStorageSubscription);
```
gobengo, line 53: hmm. Did the user definintely consent to this TOS at this point? e.g. if the end-user is just adding their credit card (in the lower left of the /account/payment page) but not necessarily choosing a plan and checking the agreement checkbox, we probably don't want to record that they consented to anything.

Realizing this made me revisit this other comment above https://github.com/web3-storage/web3.storage/pull/2028/files#r998202155

packages/website/components/accountPlansModal/accountPlansModal.js
```diff
@@ -8,12 +8,14 @@ import Modal from '../../modules/zero/components/modal/modal';
 import CloseIcon from '../../assets/icons/close';
 import AddPaymentMethodForm from '../../components/account/addPaymentMethodForm/addPaymentMethodForm.js';
 import { API, getToken } from '../../lib/api';
+import constants from '../../lib/constants.js';
 
-export async function putUserPayment(pm_id, plan_id) {
+export async function putUserPayment(agreement, pm_id, plan_id) {
```
gobengo, line 13: I think it's a good idea to add a jsdoc w/ type for the `agreement` value. I've ended up fixing several frontend bugs in all this that would have been caught by the typechecker. wdyt?

gobengo: e.g. I find myself wondering if `agreement` is allowed to be undefined, null, some other value here?

packages/website/lib/api.js
```diff
@@ -307,15 +307,17 @@ export async function deletePinRequest(requestid) {
 
 /**
  * Gets/Puts saved user plan and billing settings.
+ * @param {string} agreement
```
gobengo, line 310: Can we type this more specifically like you did in the api with a well-known list of agreement values? I think it will really help avoid bugs getting to prod because we'll be able to check things at compile time.

I don't really like how this function sometimes does a GET and sometimes does a PUT, but didn't want to nitpick at the time. But I think in both cases it's actually fine to allow this to be undefined. e.g. you might wanna use the API just to save a payment method, but without recording an agreement or subscription price.

packages/api/src/utils/billing.js
gobengo, line 32: I *think* we only need to require this when the user is actually saving their paymentSettings to be a subscription.storage value that isn't `null`, and maybe not even when it's `free`. Not exactly sure, it's a business/legal call but this is the function that is meant to encode that 'business logic' (which we should test well once we understand it to prevent regressions).

So if `paymentSettings.subscription.storage === null` we probably don't need to require an agreement.

packages/website/lib/api.js
gobengo: I hadn't had time yet but it seems like we should have a new small package in the monorepo for things that are shared across the API and the website, e.g. the known agreement types.

It would make it easier to type these kinds of things in sync. For now since we only have one agreement type I might just do
```suggestion
 * @param {'web3.storage-tos-v1'} agreement
```

(if anyone has time though we should go ahead and add the new package, but it takes a bit wrestling with the monorepo linking sometimes)

packages/website/pages/account/payment.js
```diff
@@ -67,7 +68,7 @@ const PaymentSettingsPage = props => {
       setIsFetchingPaymentSettings(true);
       setNeedsFetchPaymentSettings(false);
       try {
-        setPaymentSettings(await userBillingSettings());
+        setPaymentSettings(await userBillingSettings(constants.TERMS_OF_SERVICE_VERSION));
```
gobengo, line 71:
```suggestion
        setPaymentSettings(await userBillingSettings());
```

I dont think we should need to pass an agreement value just to GET the payment settings.

packages/website/tsconfig.json
```diff
@@ -4,7 +4,11 @@
     "allowJs": true,
     "checkJs": true,
     "target": "esnext",
-    "lib": ["ESNext", "DOM", "DOM.Iterable"],
+    "lib": [
```
gobengo, line 7: I"m glad you checked this in. My local kept reformatting this too but I wasn't sure if it was something weird on my setup. 🙏

packages/api/test/user-payment.spec.js
```diff
@@ -154,11 +155,46 @@ describe('userPaymentPut', () => {
       customers.mockCustomers.map(c => c.id).includes(billing.paymentMethodSaves[0].customerId),
       'billing.paymentMethodSaves[0].customerId is in customers.mockCustomers')
   })
+  it('throws an error if no agreement', async function () {
```
gobengo, line 158: Can we add a test not for the error conditions but also to ensure that the agreement is saved?

One way of doing it would be to invoke userPaymentPut with an `env.termsOfService` that is a mock and ensure that the correct agreement value was passed to a method on that TermsOfServiceService.

(If we had a way 'from the outside' of knowing things had been saved, e.g. userPaymentGet returning a list of agreements, then we might not even need a mock, you could do userPaymentPut with an agreement and then userPaymentGet to ensure it had been saved, but eh it's not worth adding new functionality to userPaymentGet just for the test, so that's why the mocks are nice for unit testing)

packages/api/src/errors.js
e-schneid: It's the same.  I also felt it was an odd way of writing it, but it follows the existing pattern.

packages/api/src/utils/billing-types.ts
e-schneid: I actually made it optional because the type was being reused on the `GET`, but that method probably shouldn't return the agreement (it might be preferable to have separate types).  But if we want to make it optional I should probably only `throw` when it's invalid (currently it'll throw if it's missing, so it's not truly optional).  I'll update.

packages/api/src/utils/billing-types.ts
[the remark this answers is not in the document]
e-schneid: This makes sense.  Updating.

However I want to note that we currently have a `terms_of_service` table (rather than a generic `agreements` table).  It makes sense to me to model agreement logging in separate tables, since it seems unlikely we'd want to query for all agreements for a given user (regardless of their type), but I wanted to call this out in case you had other thoughts.  _Should_ we change the `terms_of_service` table to `agreements` and put _all_ agreement logging in that one table?

packages/website/tsconfig.json
e-schneid: I've actually been reverting this change for awhile, and pushing this up was a total accident.  I'm glad I'm not the only one!  I will leave it in there.

packages/api/src/utils/billing.js
e-schneid: updating

packages/api/src/utils/billing.js
gobengo: I did this. ✅

packages/api/src/utils/stripe.js
gobengo: It still feels odd to me to construct the agreements service in this stripe-specific context, but I spent some time refactoring it and hit my timebox. This'll work well in practice, and we can refactor if/when needed.

packages/api/test/user-payment.spec.js
```diff
@@ -334,7 +371,9 @@ describe('savePaymentSettings', async function () {
   it('saves subscription using billingService', async () => {
     const desiredPaymentSettings = {
       paymentMethod: { id: `pm_mock_${randomString()}` },
-      subscription: { storage: { price: storagePriceNames.lite } }
+      subscription: { storage: { price: storagePriceNames.lite } },
+      /** @type {import('src/utils/billing-types.js').Agreement} */
```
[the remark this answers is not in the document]
gobengo: https://github.com/web3-storage/web3.storage/pull/2028/commits/122cc70f1f4058cb6a51e4d1eb3ddaea55490811

packages/db/index.d.ts
gobengo: eric renamed it :) ty

packages/db/index.js
gobengo: https://github.com/web3-storage/web3.storage/pull/2028/commits/7631d7a57bb35237590b3c8596d689a81367d4f3

packages/website/components/account/addPaymentMethodForm/addPaymentMethodForm.js
gobengo: fixed via https://github.com/web3-storage/web3.storage/pull/2028/commits/eb506e338c403f970ce0bb6c19fcc35d92ea04ec

packages/website/components/accountPlansModal/accountPlansModal.js
gobengo: fixed in https://github.com/web3-storage/web3.storage/pull/2028/commits/eb506e338c403f970ce0bb6c19fcc35d92ea04ec

(I got rid of this `putUserPayment` function and replaced it with the very similar fn in the api module)

packages/website/hooks/use-payment.js
```diff
@@ -32,7 +33,7 @@ export const usePayment = () => {
       setIsFetchingPaymentSettings(true);
       setNeedsFetchPaymentSettings(false);
       try {
-        setPaymentSettings(await userBillingSettings());
+        setPaymentSettings(await userBillingSettings(constants.TERMS_OF_SERVICE_VERSION));
```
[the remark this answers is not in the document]
gobengo: Updated in https://github.com/web3-storage/web3.storage/pull/2028/commits/eb506e338c403f970ce0bb6c19fcc35d92ea04ec

packages/website/lib/api.js
gobengo: I typed this more explicitly (I think in https://github.com/web3-storage/web3.storage/pull/2028/commits/eb506e338c403f970ce0bb6c19fcc35d92ea04ec)

packages/website/pages/account/payment.js
gobengo: I did it

packages/api/test/user-payment.spec.js
gobengo: I added a lot of tests, including the unit test I described above.

I also added some tests for some various use cases of using this http api (and updated business logic accordingly), including:
* Just using the api to update the default payment method, which happens on /account/payment in lower left (but not in context of making agreements or choosing plan)
* using api to update the plan, which requires the agreement

gobengo:
As I added tests for the different ways of invoking the http api and what should require agreements (vs not), I had to change even more than I realized in backend code (not your stuff, more like my old stuff). Basically because before the two places the frontend did PUT /user/payment were kinda the same. But now they're kinda different because when the AddPaymentForm just wants to update your payment method, that should require an agreement (ie because we don't even show that agreement to the end user in that part of the UI).

The other thing that's major is I got rid of the uniquness requirement on the agreement table. Now we'll record the timestamp of each time the user agrees to the thing. It might be more than we need, but it won't be a ton of data that I think should be problematic. The main reason I did this is I think the code that was there before would have had an uncaught error the second time savePaymentSettings was called with same agreement due to the uniqueness constraint in the db, and this was a quick way of avoiding that without having to do more clever queries.

gobengo:
@e-schneid any concerns about the updates I made?

packages/db/index.js
```diff
@@ -1295,6 +1295,25 @@ export class DBClient {
...
 
+  /**
+   * @param {string} userId
+   * @param {import('./db-client-types').AgreementKind} agreement
+   * @returns {Promise<void>}
+   */
+  async createUserAgreement (userId, agreement) {
+    const { error } = await this._client
+      .from('agreement')
+      .insert({
```
e-schneid, line 1306: As an insert, won't `error` be populated if the same user adds the same agreement?  Then this will throw and it doesn't seem like it's being caught.  It's possible this intended (also possible I'm misreading), but just wanted to call it out just in case.

Everything else looks great!  I love the solution.

gobengo: I got rid of the uniqueness constraint in tables and in the migration. So... I think it won't error in that case eh?
here is the relevant test: https://github.com/web3-storage/web3.storage/pull/2028/commits/5c84b5bb82279e876d8ca5f47b3497a0ee2482c5

Or am I misunderstanding?

gobengo (approved):
go team 🎉

PR #2034, opened 2022-10-14
feat: enterprise let's chat form

drewdelano:
https://www.notion.so/Custom-Plan-let-s-chat-enterprise-form-58358bd0f51a489089b8c15a798bf2df


drewdelano:
There's a styling conflict with the kwes-form default styling.  There doesn't seem to be a way to turn it off easily in kwes-form either.  Maybe @cmunns can help point me in the right direction?

drewdelano:
@heyjay44 can you take a look at this and let me know if this is close enough?

drewdelano:
Talked in the sync meeting.  I'm going to fix it.

gobengo:
Lookin good! (@drewdelano I think I saw when you demoed yesterday and it is way better now!)

Screen Shot 2022-10-18 at 6 55 09 PM

packages/website/components/enterpriseTierRequestModal/enterpriseTierRequestModal.js
```diff
@@ -0,0 +1,76 @@
...
+
+const UserRequestModal = ({ isOpen, onClose }) => {
+  useEffect(() => {
+    kwesforms.init();
+  }, []);
+
+  useLayoutEffect(() => {
+    const form = document.getElementById('enterpriseKwesForm');
+    form?.addEventListener('kwSubmitted', function () {
+      // do we need to do any custom logic?
```
gobengo, line 17: 🤷  I don't think so? @drewdelano what did you have in mind?

If nothing in particular, will you remove this?

I think because it does a DOM query by element id, if there are several of these on the same page, it's not guaranteed that it will select the HTMLElement that corresponds to this component instance. (there may also be react-specific concerns too as I think react reserves the right to destroy/recreate those HTML elements whenver it wants which would remove the listener).

because of all that I try to just avoid using `id` attributes altogether in reusable components.

packages/website/components/enterpriseTierRequestModal/enterpriseTierRequestModal.js
```diff
@@ -0,0 +1,76 @@
...
+import kwesforms from 'kwesforms';
+
+import Modal from 'modules/zero/components/modal/modal';
+import CloseIcon from 'assets/icons/close';
+import Button from 'components/button/button.js';
+import GradientBackground from 'components/gradientbackground/gradientbackground';
+
+const UserRequestModal = ({ isOpen, onClose }) => {
+  useEffect(() => {
+    kwesforms.init();
```
gobengo, line 11: For my own edification, what does this do? i.e. what would be bad if we didn't call it? Does it set up form validation js or something?

packages/website/components/enterpriseTierRequestModal/enterpriseTierRequestModal.js
```diff
@@ -0,0 +1,76 @@
+import { useEffect, useLayoutEffect } from 'react';
+import kwesforms from 'kwesforms';
+
+import Modal from 'modules/zero/components/modal/modal';
+import CloseIcon from 'assets/icons/close';
+import Button from 'components/button/button.js';
+import GradientBackground from 'components/gradientbackground/gradientbackground';
+
+const UserRequestModal = ({ isOpen, onClose }) => {
```
gobengo, line 9: I was a little confused about this name. And I see later we import it as `EnterpriseTierRequestModal`. wdyt of calling it that in here too?

Might also be good to have a jsdoc explaining what it's for, that way another dev that finds it can understand if it's a good fit for what they're doing. jsdoc for the two props types could also help avoid bugs with another dev calling it wrong. (I've fixed some website bugs that would have been caught by having prop types in jsdoc)

packages/website/pages/pricing.js
```diff
@@ -132,96 +135,107 @@ const Card = props => {
...
-              callToActionUrl="/account/payment?plan=lite"
-            >
-              <li className="pricing-bullet-1">
-                <i>Everything from the Free tier, plus...</i>
-              </li>
-              <li className="pricing-bullet-2">
-                Additional storage for personal usage or projects requiring lower data volumes
-              </li>
-            </Card>
+              <Card
```
gobengo, line 164: GitHub shows the space before the `<` a bit darker green, and only on this card. Not sure what that means, but there might be an extra space in there?

packages/website/pages/pricing.scss
```diff
@@ -15,6 +15,144 @@
...
+      small {
+        margin: 0px;
+      }
+
+      input,
+      textarea {
+        padding: 10px 15px;
+        border-radius: 4px;
+        font-size: 1rem;
+        // background-color: rgba(117, 122, 155, 0.15);
```
gobengo, line 120: fyi these are unused

packages/website/styles/layouts/userRequestsModals.scss
```diff
@@ -52,18 +52,18 @@
   > * {
     position: relative;
   }
-}
 
-.user-request-modal__heading {
-  font-size: 1.5rem;
-}
+  .user-request-modal__heading {
```
gobengo, line 56: I also see the darker green here in github? Might not be worth worrying about but just fyi.

packages/website/components/enterpriseTierRequestModal/enterpriseTierRequestModal.js
gobengo: (if I'm confused, someone else might be. Could be good to put a one-line comment for others that are new to kwesforms)

gobengo (changes requested):
99% there. @drewdelano will you
* [ ] add jsdocs to the enterpriseTierRequestForm
* [ ] remove the unused kwSubmitted form handler (attached to id in a way that might be a footgun if someone else uses it later)

other things like the whitespace stuff are just fyi.

packages/website/components/enterpriseTierRequestModal/enterpriseTierRequestModal.js
drewdelano: Yeah, I think you're right.  I copied it from the older kwes-form usages.

packages/website/components/enterpriseTierRequestModal/enterpriseTierRequestModal.js
drewdelano: Yeah, I was just copying from the older implementation.  I have no idea.

packages/website/styles/layouts/userRequestsModals.scss
drewdelano: I moved the `}` from like 55 in the old diff down

packages/website/pages/pricing.js
drewdelano: Looks like VS Code fixed some formatting here.  As far as I can tell this is correct based on the other `<Card />`s

packages/website/styles/layouts/userRequestsModals.scss
drewdelano: The problem I mentioned yesterday turned out to not be an issue with kwes-forms, but with the fact that we hadn't scoped our styles to the component instance and so the styling was bleeding across all of them

gobengo: @drewdelano good catch :)

gobengo (approved):

PR #2035, opened 2022-10-17
feat: add SATNAV and DUDEWHERE index buckets

alanshaw:
* `SATNAV` stores CARv2 indexes for CAR files uploaded to the service
* `DUDEWHERE` stores a mapping of root data CID => CAR CID(s)

This enables [freeway](https://github.com/web3-storage/freeway) to lookup CAR CID(s) for a given root data CID and UnixFS export data directly from R2.


olizilla (approved):
Behold! It does the things!

Screenshot 2022-10-17 at 14 26 34

_pictured:_ miniflare having written a CAR file, a DUDEWHERE key, and a SATNAV index via the [`--r2-persist`](https://miniflare.dev/storage/r2#persistence) flag

packages/api/src/car.js
```diff
@@ -122,7 +124,9 @@ export async function handleCarUpload (request, env, ctx, car, uploadType = 'Car
 
   await Promise.all([
     putToS3(env, s3Key, carBytes, carCid, sourceCid, structure),
-    putToR2(env, r2Key, carBytes, carCid, sourceCid, structure)
+    putToR2(env, r2Key, carBytes, carCid, sourceCid, structure),
+    writeSatNavIndex(env, carCid, carBytes),
+    writeDudeWhereIndex(env, sourceCid, carCid)
```
olizilla, line 129: idle thought: a shared AbortController so we bail on all if any has failed... or maybe just write the indexes only if the CAR upload has succeeded... I can't think of anything that gets worse if we end up with DUDEWHERE links and car indexes but no CAR, other than we could have failed on the first read (no entry in DUDEWHERE) rather than second read.

vasco-santos (commented):
RAD 🎉

PR #2037, opened 2022-10-17
fix: rm staging url in PSA docs

### LINKED ISSUE #2024: Remove mention of staging in docs
Under the "delete a pin" section in this page https://web3.storage/docs/how-tos/pinning-services-api/, we mention `api-staging` in the url. Can we replace that by simply `api`

image

yusefnapora:
closes #2024 by changing the API endpoint from `api-staging` to `api`.


joshJarr (approved):
LGTM, thanks!

PR #2043, opened 2022-10-17
fix: properly center align the loading icon within login/logout button

### LINKED ISSUE #2002: "Log in" / "Log out" button loading icon is not centered horizontally or vertically
The top right button that says "Log in" or "Log out" has a loading icon on it while the website tries to determine login state. This icon is pretty off-center. I think it should be centered both horizontally and vertically.
image

e-schneid:
Closes #2002 

Context: center aligns the loading icon within login/logout button


johnathonroach (approved):

gobengo (approved):
Not sure of the context exactly but I can see how this might fix a bug

PR #2048, opened 2022-10-18
feat: show tier modal on account page after logging in

jsdevel:
This shows the user the plan they wanted from the home page after logging in.

### DIFF
--- a/packages/website/components/accountPlansModal/accountPlansModal.js
+++ b/packages/website/components/accountPlansModal/accountPlansModal.js
@@ -31,10 +31,15 @@ export async function putUserPayment(pm_id, plan_id) {
 }
 
 /**
+ *
+ * @typedef {import('../../components/contexts/plansContext').Plan} Plan
+ */
+
+/**
  * @param {object} obj
  * @param {any} obj.isOpen
  * @param {any} obj.onClose
- * @param {any} obj.planSelection
+ * @param {undefined|Plan} obj.planSelection
  * @param {any} obj.planList
  * @param {any} obj.stripePromise
  * @param {any} obj.setCurrentPlan
@@ -58,7 +63,7 @@ const AccountPlansModal = ({
   stripePromise,
 }) => {
   const [isCreatingSub, setIsCreatingSub] = useState(false);
-  const currentPlan = planList.find(p => p.id === planSelection.id);
+  const currentPlan = planList.find(p => p.id === planSelection?.id);
   return (
     <div className="account-plans-modal">
       <Modal
--- a/packages/website/pages/account/payment.js
+++ b/packages/website/pages/account/payment.js
@@ -2,6 +2,8 @@
  * @fileoverview Account Payment Settings
  */
 
+import { parse as queryParse } from 'querystring';
+
 import { useState, useEffect, useMemo } from 'react';
 import { Elements, ElementsConsumer } from '@stripe/react-stripe-js';
 import { loadStripe } from '@stripe/stripe-js';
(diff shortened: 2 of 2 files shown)


packages/website/pages/account/payment.js
```diff
@@ -2,6 +2,7 @@
  * @fileoverview Account Payment Settings
  */
 
+import { parse as queryParse } from 'querystring';
```
gobengo, line 5: I think this will bundle in a browser polyfill of this thing from nodejs stdlib. wdyt of using the DOM URLSearchParams constructor? https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams

jsdevel: @gobengo i looked and it looks like we're already bundling it.  i can follow up with a tech debt PR if you want to refactor all locations in the codebase to use `URL`.
querystring

packages/website/pages/account/payment.js
```diff
@@ -187,6 +197,9 @@ const PaymentSettingsPage = props => {
           onClose={() => {
             setIsPaymentPlanModalOpen(false);
             setHasAcceptedTerms(false);
+            if (planQueryParam) {
+              history.pushState({}, '', window.location.href.replace(/plan=[^&]+&?/, ''));
```
gobengo, line 201: Is this inteded to wipe the query param?

If so, I think it's kinda nice to keep it around. That way the user can share the URL to that tier to themselves or others (e.g. via user-agent share functionality, copy url bar and paste elsewhere, etc)

Just curious, what is the motivation of the pushState and, if I'm reading it right, clearing the query param? If it's not strictly needed we might as well not do it.

jsdevel: so, if the user closes the modal, it's their intention _not_ to see it again.  if they refresh the browser after closing the modal and then see the modal again, i figured it would be a little annoying.  if they click the browser back button they'll see the modal and query param again.  lmk what you want.

gobengo: ah, it's onclose. Nice. Sry for missing that.

gobengo (approved):
Approach looks great. Approving so as to not block, but probably best to ensure the tsc check passes before merging. re-request review when it's green if you want a pair of eyes on whatever the fix is.

packages/website/pages/account/payment.js
gobengo: >  if they click the browser back button they'll see the modal and query param again with my current implementation.

awesome

packages/website/components/accountPlansModal/accountPlansModal.js
```diff
@@ -30,11 +30,16 @@ export async function putUserPayment(pm_id, plan_id) {
...
+/**
+ *
+ * @typedef {import('../../components/contexts/plansContext').Plan} Plan
+ */
+
 /**
  * @param {object} obj
  * @param {any} obj.isOpen
  * @param {any} obj.onClose
- * @param {any} obj.planSelection
```
gobengo, line 37: nice, I didn't realize all these `any` were in there :P

packages/website/pages/account/payment.js
```diff
@@ -45,19 +46,36 @@ import GeneralPageData from '../../content/pages/general.json';
  * @property {PaymentMethodCard} card
  */
 
+function removePlanQueryParam() {
```
gobengo, line 49: love how the function name is self-documenting 🙏

gobengo (approved):

jsdevel:
> Approach looks great. Approving so as to not block, but probably best to ensure the tsc check passes before merging. re-request review when it's green if you want a pair of eyes on whatever the fix is.

yea, pushed up some changes to make the build pass.  fingers crossed...

Issue #1987, opened 2022-10-02
I cant pay im getting: failed to get/put user billing info: {}


Shuberth:
Hi, i cant finalize a new payment ussing my credit card, im  getting this: failed to get/put user billing info: {}

dchoi27:
Hey, are you doing this in the website? Mind trying again to see if the problem persists?

alanshaw:
@gobengo have you seen this before?

@Shuberth is your github account the account you're signed in with and trying to add payment (just so we can look you up and see what happened)?

heyjay44:
The issue has been resolved in PR #1988

PR #2058, opened 2022-10-20
feat: AddPaymentMethodForm shows a more end-user-friendly error message when it encounters an error

gobengo:
Motivation:
* provide a more helpful error message in situations like this https://github.com/web3-storage/web3.storage/issues/1987

How:
* If the AddPaymentForm encounters an error, it doesn't render the error message (that is probably more intended for a developer to debug than the end-user), it renders in the html a message that gives the user an affordance to get in touch with us if the error keeps happening, and doesn't show any dev details. Dev details still appear in the dev console to aid debugging.

### DIFF
--- a/packages/website/components/account/addPaymentMethodForm/addPaymentMethodForm.js
+++ b/packages/website/components/account/addPaymentMethodForm/addPaymentMethodForm.js
@@ -2,7 +2,7 @@ import { useState } from 'react';
 import { CardElement, useElements, useStripe } from '@stripe/react-stripe-js';
 
 import Loading from '../../../components/loading/loading';
-import { userBillingSettings } from '../../../lib/api';
+import { APIError, defaultErrorMessageForEndUser, userBillingSettings } from '../../../lib/api';
 import Button from '../../../components/button/button';
 import { planIdToStorageSubscription } from '../../contexts/plansContext';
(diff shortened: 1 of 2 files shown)


packages/website/components/account/addPaymentMethodForm/addPaymentMethodForm.js
```diff
@@ -52,12 +52,12 @@ const AddPaymentMethodForm = ({ setHasPaymentMethods, setEditingPaymentMethod, c
...
         setHasPaymentMethods?.(true);
         setEditingPaymentMethod?.(false);
-        setPaymentMethodError('');
+        setPaymentMethodError(null);
       } catch (error) {
-        let message;
-        if (error instanceof Error) message = error.message;
-        else message = String(error);
-        setPaymentMethodError(message);
+        if (!(error instanceof APIError)) {
```
drewdelano, line 57: will Sentry see these?  if not, do we want to explicitly send them to Sentry?

gobengo: We don't have sentry at all on the frontend! but I want to (search slack dag.house)... will handle in separate issue/priority. Good idea/concern though :)

PR #2062, opened 2022-10-21
fix: clarity on web3.storage PSA

dchoi27:


yusefnapora (approved):
Looks good to me - it's good to call out that you don't need the PSA for things uploaded through us.

PR #2069, opened 2022-10-26
perf: decouple user used storage sql function

vasco-santos:
This PR aims to optimize performance for calculating user used storage. It includes:
- decouple `user_used_storage` into 2 functions, one for regular uploads and other for psa
- both regular uploads and psa now ignore Pin status. If it is not Pinned, we won't know the dag size.
  - this should be a big boost given Pin table is by far the largest table we have
  - an accepted side effect by us is that non completed uploads that get stalled will be accounted
- changes psa sql indexes so that a sequential scan is avoided
  - seq_scan on psa_pin_request is done to check the deleted_at and to join with auth_key table
  - `psa_pin_request_search_idx` should perform better and includes the 3 columns used
  - other psa indexes were only used in this flow

### DIFF
--- a/packages/api/test/user.spec.js
+++ b/packages/api/test/user.spec.js
@@ -33,7 +33,7 @@ describe('GET /user/account', () => {
     assert(res.ok)
     const data = await res.json()
     assert.strictEqual(data.usedStorage.uploaded, 32000)
-    assert.strictEqual(data.usedStorage.psaPinned, 10000)
+    assert.strictEqual(data.usedStorage.psaPinned, 710000)
   })
 })
 
--- a/packages/db/index.js
+++ b/packages/db/index.js
@@ -6,7 +6,8 @@ import {
   normalizePins,
   normalizeDeals,
   normalizePsaPinRequest,
-  parseTextToNumber
+  parseTextToNumber,
+  safeNumber
 } from './utils.js'
 import { ConstraintError, DBError, RangeNotSatisfiableDBError } from './errors.js'
 
@@ -316,19 +317,28 @@ export class DBClient {
    * @returns {Promise<import('./db-client-types').StorageUsedOutput>}
    */
   async getStorageUsed (userId) {
-    /** @type {{ data: { uploaded: string, psa_pinned: string, total: string }, error: PostgrestError }} */
-    const { data, error } = await this._client
-      .rpc('user_used_storage', { query_user_id: userId })
-      .single()
+    const [userUploadedResponse, psaPinnedResponse] = await Promise.all([
+      this._client
+        .rpc('user_uploaded_storage', { query_user_id: userId })
+        .single(),
+      this._client
+        .rpc('user_psa_storage', { query_user_id: userId })
+        .single()
+    ])
 
-    if (error) {
-      throw new DBError(error)
+    if (userUploadedResponse.error) {
+      throw new DBError(userUploadedResponse.error)
+    } else if (psaPinnedResponse.error) {
+      throw new DBError(psaPinnedResponse.error)
     }
 
+    const uploaded = parseTextToNumber(userUploadedResponse.data)
+    const psaPinned = parseTextToNumber(psaPinnedResponse.data)
+
     return {
-      uploaded: parseTextToNumber(data.uploaded),
-      psaPinned: parseTextToNumber(data.psa_pinned),
-      total: parseTextToNumber(data.total)
+      uploaded,
+      psaPinned,
+      total: safeNumber(uploaded + psaPinned)
     }
   }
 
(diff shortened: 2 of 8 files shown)


packages/db/postgres/functions.sql
```diff
@@ -265,16 +271,30 @@ BEGIN
...
                   c.dag_size
         ) AS pinned_content), 0)
     );
+  return (total)::TEXT;
+END
+$$;
+
+CREATE TYPE stored_bytes AS (uploaded TEXT, psa_pinned TEXT, total TEXT);
 
+CREATE OR REPLACE FUNCTION user_used_storage(query_user_id BIGINT)
```
vasco-santos, line 287: keeping this function because `users_by_storage_used` uses it. So, just refactored to re-use partial functions created

packages/api/test/user.spec.js
```diff
@@ -32,8 +32,9 @@ describe('GET /user/account', () => {
     })
     assert(res.ok)
     const data = await res.json()
+    console.log('TOTAL', data.usedStorage)
     assert.strictEqual(data.usedStorage.uploaded, 32000)
-    assert.strictEqual(data.usedStorage.psaPinned, 10000)
+    assert.strictEqual(data.usedStorage.psaPinned, 710000)
```
vasco-santos, line 36: Needed updated as now we are counting fixtures that were not pinned status from https://github.com/web3-storage/web3.storage/blob/main/packages/api/test/fixtures/init-data.sql#L156-L167

packages/db/test/user.spec.js
```diff
@@ -183,23 +183,6 @@ describe('user operations', () => {
...
-
-    // Create Upload not pinned yet. It should not be counted yet.
-    await createUpload(client, user._id, authKey._id, await randomCid(), {
-      dagSize: 1000000,
-      pins: initialPinsNotPinned
-    })
-
-    const usedStorageWithFailed = await client.getStorageUsed(user._id)
-    assert.strictEqual(usedStorageWithFailed.uploaded, dagSize1 + dagSize2, 'used storage should not count unpinned')
-
```
vasco-santos, line 202: These would be accounted now, so really no need to add them given there is no exception anymore

packages/db/test/utils.js
```diff
@@ -127,37 +127,13 @@ export async function createUserWithFiles (dbClient, options = {}) {
   const pinRequests = 3
   const dagSize = Math.ceil(((percentStorageUsed / 100) * storageQuota) / (uploads + pinRequests))
 
-  // Create a failed upload.
```
vasco-santos, line 130: Failed + Not pinned are now counted too

packages/db/index.js
```diff
@@ -316,19 +317,28 @@ export class DBClient {
    * @returns {Promise<import('./db-client-types').StorageUsedOutput>}
    */
   async getStorageUsed (userId) {
-    /** @type {{ data: { uploaded: string, psa_pinned: string, total: string }, error: PostgrestError }} */
-    const { data, error } = await this._client
-      .rpc('user_used_storage', { query_user_id: userId })
-      .single()
+    const [userUploadedResponse, psaPinnedResponse] = await Promise.all([
+      this._client
```
flea89, line 321: What is the rationale behind splitting the 2 up?
Is it to better understand where the bottleneck is?

flea89 (commented):
Nice one! This should hopefully improve perf substantially!

As discussed offline, it'd be great to make it obvious to users that failed/uncompleted uploads might count against their quota.
Create a ticket for that [here](https://github.com/web3-storage/web3.storage/issues/2081)?

packages/db/utils.js
```diff
@@ -122,3 +122,13 @@ export function parseTextToNumber (numberText) {
   }
   return num
 }
+
+/**
+ * @param {Number} num
+ */
+export function safeNumber (num) {
```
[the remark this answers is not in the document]
alanshaw: This is not parsing anything - just ensuring it's within the safe integer range.

Maybe `validateSafeInteger`?

alanshaw (approved):
LGTM

packages/db/utils.js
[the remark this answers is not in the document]
flea89: what I meant is 
```
export function parseTextToNumber (numberText) {
  const num = Number(numberText)
  validateSafeInteger(num)
  return num
}

/**
 * @param {Number} num
 */
export function validateSafeInteger (num) {
  if (!Number.isSafeInteger(num)) {
    throw new Error('Invalid integer number.')
  }
  return num
}
```

to avoid duplicating the validation in 2 functions... 

Really minor ;)

packages/db/index.js
vasco-santos: Goal is to decrease the request time to be less likely of getting to the 30s timeout of heroku dyno. With Alan's account that has a lot of data this was taking so much.

Also, we were talking about having a view per product usage, and we will be able to perform queries specific for each thing.

PR #2070, opened 2022-10-26
feat: add metadata to upload error reports

olizilla:
add metadata to upload error reports in sentry so we can debug when we see an uptick in concurrent upload errors.

the `sentry.addExtras` function lets us set additional metadata on error reports and shows up like the screenshot below:
Screenshot 2022-10-26 at 15 39 54

> setExtras: Set an object that will be merged sent as extra data with the event.
– https://github.com/robertcepa/toucan-js#features

License: MIT


vasco-santos (approved):
Nice 👍🏼 did not know about this feature

PR #2077, opened 2022-10-27
fix: ensure prod api deploy can only happen from main

olizilla:
Update the api workflow so that the prod deploy job only runs on main and if there is a release created flag from release_please

This brings the api workflow inline with the website one: https://github.com/web3-storage/web3.storage/blob/53e6da1dd62831f85735ed2b403427e731ee863d/.github/workflows/website.yml/#L242

this should prevent weird deploys happening just due to a timing error, and means we wont see a github deployment for non-main api builds... note: we will still lots of prod deploy builds as we also provide the production env to website builds, and github doesn't know about out api vs website distinction.

License: MIT

### DIFF
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -88,15 +88,6 @@ jobs:
       name: production
       url: https://web3.storage
     steps:
-      - uses: GoogleCloudPlatform/release-please-action@v3
-        id: tag-release
-        with:
-          path: packages/api
-          token: ${{ secrets.GITHUB_TOKEN }}
-          release-type: node
-          monorepo-tags: true
-          package-name: api
-          changelog-types: '[{"type":"feat","section":"Features","hidden":false},{"type":"fix","section":"Bug Fixes","hidden":false},{"type":"chore","section":"Other Changes","hidden":false}]'
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/api.yml
```diff
@@ -58,10 +76,14 @@ jobs:
...
-    name: Release
+  
+  deploy-production:
+    name: Deploy Production
+    if: (github.event_name == 'push' && github.ref == 'refs/heads/main' && needs.changelog.outputs.releases_created) || inputs.force_release
     runs-on: ubuntu-latest
-    needs: test
+    needs:
+      - test
+      - changelog
```
vasco-santos, line 86: release please step below should go away right?

.github/workflows/api.yml
```diff
@@ -76,17 +98,12 @@ jobs:
           package-name: api
           changelog-types: '[{"type":"feat","section":"Features","hidden":false},{"type":"fix","section":"Bug Fixes","hidden":false},{"type":"chore","section":"Other Changes","hidden":false}]'
       - uses: actions/checkout@v2
-        if: ${{ steps.tag-release.outputs.releases_created }}
```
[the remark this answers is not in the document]
olizilla: In this PR, we make deploy-production `need` changelog` and that the event be a push to main, so the whole job is skipped unless we really are ready to do a prod deploy

.github/workflows/api.yml
olizilla: ah yes

olizilla: Fixed

vasco-santos (approved):
LGTM
Thanks sir ❤️

PR #2089, opened 2022-11-01
fix: do not render account blocked modal unless account restricted

### LINKED ISSUE #1998: Blocked modal overlay flashes briefly sometimes when browsing web3.storage website
Sometimes, when clicking through the web3.storage website, a modal pops up for a split second. I've seen this on pretty much all pages of the website, including the docs.

From what I can read in the brief time it's up, I'm pretty sure it's this: https://github.com/web3-storage/web3.storage/blob/main/packages/website/components/account/accountBlockedModal/accountBlockedModal.js

Potentially we are making it appear by default, rather than just making it appear if the account is supposed to be blocked? Also it might not be this exact modal (the word "blocked" is definitely in it), but if it is, we should consider not making it appear on any page except the account page (potentially we call it whenever you're in the logged in state).

e-schneid:
Closes #1998 

There's a reported issue where the Account Blocked modal displays briefly when navigating the site.  I haven't yet been able to reproduce this, but currently this modal always renders (and is just hidden with CSS).  This PR is an optimization to prevent the modal from rendering at all.  It will hopefully resolve the issue, but it will also reduce the size of the prerendered HTML.

### DIFF
--- a/packages/website/components/account/accountBlockedModal/accountBlockedModal.js
+++ b/packages/website/components/account/accountBlockedModal/accountBlockedModal.js
@@ -16,7 +16,7 @@ const AccountBlockedModal = ({ hasAccountRestriction }) => {
     }
   }, [hasAccountRestriction, modalState]);
 
-  return (
+  return modalState[0] ? (
     <div className="account-blocked-modal">
       <Modal
         className=""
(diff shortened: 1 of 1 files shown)


cmunns (approved):
looks good. nice idea.

johnathonroach (approved):

PR #2095, opened 2022-11-04
wip: scaffolding for "v2" docs

yusefnapora:
This is a very early pass at the new version of the docs site that covers w3up & friends.

So far, I've just copied over the existing content and taken a sledgehammer to it, removing things that obviously don't apply, e.g. Go client, the "query" how-to, etc. There are a great many **TODO** comments sprinkled throughout, and the "net new" docs that didn't have an existing article are all outlines at best. Here be :dragon:s, etc. 

I'll take a pass this afternoon at mapping the TODOs to existing tickets in Notion and making new ones, with the primary goal of sorting out things that are immediately actionable (no API changes on the horizon, etc).

If anyone wants to click through the deploy preview, I've stuck the new stuff at the `/docs/v2` route, but it's not linked anywhere from the home page or the existing docs yet, so you need to add the `v2` to the URL manually. 

Some things to figure out:
- [ ] Do we want to use "v2" in the URL? Does the new API have a "marketing name"?
- [ ] What's the lowest-friction "quickstart" flow we can put in front of people?
  - Can we get people into a codesandbox instead of making them copy / paste scripts?
    - And / or: go fork this template repo & follow these instructions
  - If we're having them write code, we need to explain the UCAN registration flow
    - Unless there is / will be a way to register via the UI console? Not sure what's possible here.

cc @heyjay44 & @dchoi27 - don't know if this is worth reviewing in depth yet, but if you have any ideas about the Qs above, LMK :)


dchoi27:
> Do we want to use "v2" in the URL? Does the new API have a "marketing name"?
^ the new API is `w3up` but don't think we want to tie the docs to it since the docs should apply to the other products in the platform. I think `v2` works fine :)

As for the rest of the questions...they're really good ones! Let's chat about them on Wednesday's migration call

PR #2101, opened 2022-11-09
feat: add tooltip to dashboard storage totals

### LINKED ISSUE #2081: Improve how we communicate what contributes to the storage quota to users
## Challenge
Once we merge #2069 we will start counting failed/incompleted uploads toward the calculated storage quota for a user. These failed uploads are shown in the table.
This might not be obvious to users, and, I wonder if  the paid plans makes this a bit more sensitive.

## Suggested solution 
Let's add a tooltip to explain to the user what accrues to the total, making sure they are aware that our current storage calculation also includes incompleted uploads.

### Suggested copy
```
This is the total of all your uploads, including the ones that never completed.
You can remove uploads via the table below, through your preferred client library, the CLI or the HTTP APIs.
```
We tried to keep it as compact as possible, given this goes into a  tooltip.

### How it looks in place 
Here's a screenshot of how this could look in the dashboard

Image

### Questions
We are wondering if we should suggest to users to reach out if they need help with bulk deletion, given there's no trivial way to do it with the APIs right now.
It could be something along the lines of:
```
Please get in touch with support@web3.storage if you need bulk deleting files.
```

At the same time, not sure if that's required and might create more work for you @dchoi27 

What do you think?

Follow up on #2033

flea89:
Resolves #2081

### DIFF
--- a/packages/website/content/pages/app/account.json
+++ b/packages/website/content/pages/app/account.json
@@ -76,7 +76,8 @@
       },
       "loading": "Loading...",
       "max_space_tib_label": "TiB",
-      "unlock_label": "1 TiB+"
+      "unlock_label": "1 TiB+",
+      "tooltip_total": "<div> This is the total of all your uploads, including the ones that never completed.</div> You can remove files and incomplete uploads at any point by using the table below, your preferred client library, the CLI or the HTTP APIs. <div>Please get in touch with <a href='mailto:support@web3.storage'>support@web3.storage</a> if you need help bulk deleting files.<div>"
     },
     "file_manager": {
       "tabs": [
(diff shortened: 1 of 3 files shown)


packages/website/content/pages/app/account.json
```diff
@@ -76,7 +76,8 @@
       },
       "loading": "Loading...",
       "max_space_tib_label": "TiB",
-      "unlock_label": "1 TiB+"
+      "unlock_label": "1 TiB+",
+      "tooltip_total": "<div> This is the total of all your uploads, including ones marked \"Failed\" (you can read more about what the status means <a href='https://web3.storage/docs/how-tos/get-status/'>here</a>). </div> You can remove uploads via the table below, through your preferred client library, the CLI or the HTTP APIs.<div>"
```
adamalton, line 80: What's the purpose of the trailing `<div>` on the end of this?

adamalton:
I've tested this out on the staging link and it looks great. I've just added one question about a trailing `<div>` in the HTML, but other than that this looks good to merge.

packages/website/content/pages/app/account.json
flea89: I updated the copy and missed that closing tag. 
Great catch

adamalton: 👍 fix.

joshJarr (approved):

alanshaw (approved):

PR #2103, opened 2022-11-10
feat: logins of already-existing users result in the user's customer contact info being updated (e.g. in stripe.com dashboard)

gobengo:
Motivation:
* business wants the stripe dashboard to be useful as a CRM of sorts, i.e. have pretty up to date contact info on the customer records there
  * before this PR, we only write that contact info for new user signups. This PR makes logins of previously-existing-users also result in an update of the customer contact info in stripe.
* https://www.notion.so/on-user-login-update-stripe-com-customer-with-latest-name-email-fed8d970688f4bfd8281dc6cc31581bc

Todo
* [x] write main business logic, tests that work against mock (not against stripe apis)
* [x] unit test for StripeCustomerService updateContact
* [x] integration test for StripeCustomerService updateContact
* [x] manual test login locally after stripe, ensure no errors

Challenges
* testing everything locally against an old user is hard because localdev always has an emptyish database of users.

How to Test on staging/production
* log in with email should work without error
* log in with github should work without error
* in prod, @gobengo can test with his oldish user and ensure that re-logging in will update the corresponding name/email of customer in stripe


yusefnapora (approved):
Looks solid to me and nicely tested.

Out of curiosity, is it actually possible for an email-registered user to change their email but keep the same issuer? I noticed you have a test for it, but wasn't aware that was a thing 😄 Either way, the test is a good idea and makes sense to include.

gobengo:
@yusefnapora  tysm!

> Out of curiosity, is it actually possible for an email-registered user to change their email but keep the same issuer? I noticed you have a test for it

afaik it's not possible. But it's kind of an implementation detail of how magic.link works, which I don't really understand! :P
But it seems like reasonable business logic to encode as a regression test. and if we learn otherwise it's hopefully easy to adjust it

packages/api/src/utils/stripe.js
```diff
@@ -235,19 +292,55 @@ export function createMockStripeCardPaymentMethod () {
   }
 }
 
-/** @returns {StripeComForCustomersService} */
+/**
+ * Create a mock stripe sdk instance that can be used with the StripeCustomersService
+ * @returns {StripeComForCustomersService}
+ */
 export function createMockStripeForCustomersService () {
```
travis, line 299: this looks good and makes sense, AND is starting to get to the level of complexity where it seems worth investigating whether we can use Stripe's API in "test mode" (https://stripe.com/docs/keys#test-live-modes) to get some end-to-end testing into place. 

I dug around in the Stripe API docs enough to convince myself it would probably be worth investigating this but not enough to be sure they actually implement this particular API functionality in test mode - I would definitely be down to do a quick spike on this if you think it makes sense to keep digging into this!

travis: adding - if I do find this is possible, I'd want to add it to a test suite that primarily gets run during CI so we can still run the existing tests offline/on a plane. it seems like `test:e2e` in the top level `package.json` might be a good place for this - seems like I'd need to add a similar `test:e2e` script to the `api` project, does that sound right?

travis (approved):
this all looks great to me! see my comment re: Stripe's "test mode" API and investigating a version of the CustomerService that uses it for end-to-end testing - if you think that's worth digging into I'd be happy to spend some time today/next week spiking this out

packages/api/src/utils/stripe.js
```diff
@@ -185,10 +187,62 @@ export class StripeCustomersService {
...
+   */
+  async getContact (customerId) {
+    let stripeCustomer
+    try {
+      stripeCustomer = await this.stripe.customers.retrieve(customerId)
+    } catch (error) {
+      if (hasOwnProperty(error, 'code')) {
+        switch (error.code) {
+          case 'resource_missing':
+            return new CustomerNotFound('customer not found')
```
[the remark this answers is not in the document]
gobengo: @travis in short, it's for type safety of the 'error' result value types. afaik tsc won't typecheck the type of what gets `throw`n. It's definitely unusual compared to the whole world of js, but I like the pattern of having 'result types' that are union of both 'happy path' and specific error tyeps. iirc @yusefnapora seemed fine with it too in earlier review. When I used it in here originally I was a bit inspired by how @Gozala was using them in ucanto. https://github.com/web3-storage/ucanto/blob/b752b398950120d6121badcdbb639f4dc9ce8794/packages/interface/src/lib.ts#L267

PR #2106, opened 2022-11-15
feat: load faq content from Netlify

johnathonroach:
This PR updates the FAQ page to fetch content from the Netlify blog and content site.

### DIFF
--- a/packages/website/components/accordionblock/accordionblock.js
+++ b/packages/website/components/accordionblock/accordionblock.js
@@ -1,3 +1,5 @@
+import ReactMarkdown from 'react-markdown';
+
 import { ReactComponent as Chevron } from '../../assets/icons/chevron.svg';
 import ZeroAccordion from 'ZeroComponents/accordion/accordion';
 import ZeroAccordionSection from 'ZeroComponents/accordion/accordionSection';
@@ -18,7 +20,11 @@ export default function AccordionBlock({ block }) {
               <div className="accordion-header-text">{section.heading}</div>
             </ZeroAccordionSection.Header>
             <ZeroAccordionSection.Content>
-              <div className="accordion-content-text" dangerouslySetInnerHTML={{ __html: section.content }}></div>
+              {block.markdown?.includes('content') ? (
+                <ReactMarkdown>{section.content}</ReactMarkdown>
+              ) : (
+                <div className="accordion-content-text" dangerouslySetInnerHTML={{ __html: section.content }}></div>
+              )}
             </ZeroAccordionSection.Content>
           </ZeroAccordionSection>
         ))}
--- a/packages/website/pages/faq.js
+++ b/packages/website/pages/faq.js
@@ -5,6 +5,7 @@ import BlockBuilder from '../components/blockbuilder/blockbuilder.js';
 import { initFloaterAnimations } from '../lib/floater-animations.js';
 import GeneralPageData from '../content/pages/general.json';
 import AccordionBlock from '../components/accordionblock/accordionblock';
+import Loading from '../components/loading/loading';
 
 export default function Home() {
   const [faqSections, setFaqSections] = useState();
@@ -122,6 +123,7 @@ export default function Home() {
               <div className="grid">
                 <div className="col-12 column-1">
                   <div className="column-content">
+                    {!faqSections && <Loading size="large" message="Fetching Latest FAQs..." />}
                     {faqSections && (
                       <AccordionBlock
                         block={{
--- a/packages/website/pages/faq.js
+++ b/packages/website/pages/faq.js
@@ -9,7 +9,7 @@ import AccordionBlock from '../components/accordionblock/accordionblock';
 export default function Home() {
   const [faqSections, setFaqSections] = useState();
   /** @type [any, null | any] */
-  const [markdownProps, setMarkdownProps] = useState([]);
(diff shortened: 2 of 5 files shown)


cmunns (approved):
@johnathonroach how much trouble would it be to add a loading indicator. The first load on the preview url had a bit of delay with no indication anything was happening.

Approved otherwise.

packages/website/pages/faq.js
```diff
@@ -1,13 +1,102 @@
...
 import GeneralPageData from '../content/pages/general.json';
+import AccordionBlock from '../components/accordionblock/accordionblock';
 
 export default function Home() {
+  const [faqSections, setFaqSections] = useState();
+  /** @type [any, null | any] */
+  const [propsToParseAsMarkdown, setPropsToParseAsMarkdown] = useState([]);
   const sections = FAQPageData.page_content;
   const animations = FAQPageData.floater_animations;
+  const netlifyCMSHost = process.env.NEXT_PUBLIC_NETLIFY_CMS_ENDPOINT || 'https://blog.web3.storage';
```
jsdevel, line 15: can we remove the default value of `https://blog.web3.storage` here?  i'm concerned that we may end up forgetting to add it in the env config.  or, perhaps that doesn't matter since this is the correct URL in prod?

johnathonroach: Go that fallback removed.

e-schneid: I added back the fallback.  It's consistent with how it's handled in nft.storage, and I'm not sure we need to manage an env var for this since we're using the same domain across environments.  We also link directly to blog.web3.storage in other parts of the site, so it's unlikely we'll ever want to change this.

e-schneid (approved):

packages/website/pages/faq.js
```diff
@@ -1,13 +1,102 @@
...
+  /** @type [any, null | any] */
+  const [propsToParseAsMarkdown, setPropsToParseAsMarkdown] = useState([]);
   const sections = FAQPageData.page_content;
   const animations = FAQPageData.floater_animations;
+  useEffect(() => {
+    fetch(`${process.env.NEXT_PUBLIC_NETLIFY_CMS_ENDPOINT || 'https://blog.web3.storage'}/api/partials/faq/faq`)
+      .then(async response => await response.text())
+      .then(text => {
+        const obj = JSON.parse(text);
+        const faqs = obj.props.partial.meta['faq-list'].reduce((sections, faq) => {
```
gobengo, line 21: You might want to accomodate a case where `obj` did not parse to the expected shape, because e.g. if `obj.props` is not an object, this will have an uncaught `TypeError`

packages/website/pages/faq.js
```diff
@@ -1,13 +1,102 @@
...
+            content: faq.answer,
+          });
+
+          return sections;
+        }, []);
+        setFaqSections(faqs);
+        setPropsToParseAsMarkdown(['content']);
+      })
+      .catch(e => {
+        // @ts-ignore
```
gobengo, line 39: Why swallow the error here? Can you only swallow certain expected errors, and throw if you encounter an unexpected error?

Can you avoid `@ts-ignore`?

packages/website/components/accordionblock/accordionblock.js
```diff
@@ -18,7 +20,11 @@ export default function AccordionBlock({ block }) {
               <div className="accordion-header-text">{section.heading}</div>
             </ZeroAccordionSection.Header>
             <ZeroAccordionSection.Content>
-              <div className="accordion-content-text" dangerouslySetInnerHTML={{ __html: section.content }}></div>
+              {block.markdown?.includes('content') ? (
+                <ReactMarkdown>{section.content}</ReactMarkdown>
+              ) : (
+                <div className="accordion-content-text" dangerouslySetInnerHTML={{ __html: section.content }}></div>
```
gobengo, line 26: I know this was already there, so I'm not saying we have to change it now, but this really is quite dangerous. If possible over time it seems like we should be able to avoid it, e.g. by passing `props.children` as a `React.Element` for this content instead of raw html strings.

packages/website/pages/faq.js
e-schneid: are we OK with the catch clause handling this error?

packages/website/pages/faq.js
e-schneid: updated to log the error and also updated to create a type for FaqSection to remove the ts-ignore.

packages/website/components/accordionblock/accordionblock.js
e-schneid: I could import a new package (for example DOMPurify) and pass it through there first.  Would that work?

gobengo: I don't think that's worth adding dep on dompurify vs avoiding authoring/passing-around html strings in the first place. JSX helps with sanitization, we just have to use it.

packages/website/pages/faq.js
gobengo: sure, if there was a comment in the catch handler that that's what you expect to catch, I prob wouldn't have added my original comment

e-schneid: comment added

e-schneid (approved):

gobengo (approved):

PR #2107, opened 2022-11-15
feat: add `w3 name get-key` command

### LINKED ISSUE #1835: Add `w3 name get <keyId>` command
## TLDR

I propose adding the `w3 name get <keyId>` command, to retrieve and reuse the signing key associated with an IPNS name. Currently, the revision can be updated only from the environment where the IPNS name was initially created.

I took the initiative and already implemented the solution [here](https://github.com/nvrabii/web3.storage/tree/get-signing-key). If the feature proposal is approved, I'm open to taking further guidance on how to contribute to the project.

## Why add this command?

I was building my website, intending to deploy it on IPFS (Web3.Storage). From an [article](https://betterprogramming.pub/modern-way-to-host-a-site-on-ipfs-7941b65d27c3), I learned how to deploy changes in 3 commands, but I thought I could do better.

When I was trying to update the CID associated with the taken IPNS name via a GitHub action, I could not do that without providing the corresponding signing key. I searched on Internet how to retrieve the key, but I could not find another solution as to modify the `w3`'s code, in order to `console.log` the signing key. Once I retrieved the key, I added it to the repo's secrets, so that my workflow adds the new build to IPFS with a followed update of the IPNS revision.

I don't think that my use case is unique in its kind. Many developers who would like to automate their deployments will have to "move" the signing key from the local to the deployment context.

## Why not add this command?

Maybe there is another way to retrieve the signing key, and this feature is not as useful as I thought.

## Proposed solution

Add the `w3 name get <keyId>` command to the `w3` CLI.

## Code changes

1. Add the command to [w3/bin.js](https://github.com/nvrabii/web3.storage/blob/acb5b7d2a74ba557018a4d3a754f5b9ff73fbafa/packages/w3/bin.js#L72-L75)
2. Add the handler to [w3/name.js](https://github.com/nvrabii/web3.storage/blob/acb5b7d2a74ba557018a4d3a754f5b9ff73fbafa/packages/w3/name.js#L50-L60)
3. Update [w3/README.md](https://github.com/nvrabii/web3.storage/blob/acb5b7d2a74ba557018a4d3a754f5b9ff73fbafa/packages/w3/README.md?plain=1#L90-L94)

## How to test

From what I've seen in the repo, `w3 name ...` commands are [not tested](https://github.com/web3-storage/web3.storage/blob/main/packages/w3/test/bin.spec.js). The code for `Name.getKey(keyId)` reuses the key presence check from [`Name.publish()`](https://github.com/web3-storage/web3.storage/blob/96ff423d778543d0fda4e5f9da6a7ff2417f4587/packages/w3/name.js#L21-L24).

## Risks

There are no reasons to "hide" the signing key in the `config` because it can be easily retrieved by adding a `console.log` either in the `publish` or `create` functions (as a workaround to the existing limitations in the CLI). 

Backward compatibility will be respected because the changes only extend the interface.

adamalton:
Fixes #1835.

### DIFF
--- a/packages/w3/README.md
+++ b/packages/w3/README.md
@@ -1,9 +1,9 @@
 <h1 align="center">⁂<br/>web3.storage</h1>
 <p align="center">The CLI for web3.storage</p>
 
-## Getting started 
+## Getting started
 
-Install the CLI from npm 
+Install the CLI from npm
 
 ```console
 $ npm install -g @web3-storage/w3
@@ -99,7 +99,7 @@ Working with name records simply updates the Web3.Storage cache of data.
 
 Resolve the current IPNS record revision for the passed name.
 
-### `w3 name get <keyId>`
+### `w3 name get-key <keyId>`
 
 **❗️Experimental** this command may not work, may change, and may be removed.
 
--- a/packages/w3/README.md
+++ b/packages/w3/README.md
@@ -99,6 +99,12 @@ Working with name records simply updates the Web3.Storage cache of data.
 
 Resolve the current IPNS record revision for the passed name.
 
+### `w3 name get <keyId>`
+
+**❗️Experimental** this command may not work, may change, and may be removed.
+
+Retrieve the signing key associated with the passed name.
+
 ### `w3 name rm <keyId>`
 
 **❗️Experimental** this command may not work, may change, and may be removed.
--- a/packages/w3/bin.js
+++ b/packages/w3/bin.js
@@ -74,6 +74,11 @@ cli.command('name resolve <keyId>')
   .example('name resolve k51qzi5uqu5dlcuzv5xhg1zqn48gobcvn2mx13uoig7zfj8rz6zvqdxsugka9z')
   .action(Name.resolve)
 
+cli.command('name get <keyId>')
+  .describe('Retrieve the signing key associated with the passed name.')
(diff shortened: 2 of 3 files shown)


nvrabii (approved):

joshJarr (approved):
LGTM, great shout.

adamalton:
@alanshaw having both an `import` and an `export` command makes sense. Do you think that these should only be for doing one key at a time, or do you think that it would actually be a nicer user experience to be able to export _all_ of your name keys at once into a single file, and then import them in bulk too? Or do you think that the common use case of `w3 name` functionality is typically only one or two names at a time?

alanshaw:
IMO just one at a time...multiple would mean that there has to be a format to the exported file and I don't want to have to define that, or trust users to pick out a particular key from it. You also can't give it directly to something else that is able to read the keys (like programmitcally to the library) without also making changes there.

alanshaw:
...you could also add `w3 name export --all` _later_ and augment `w3 name import` to be able to read one or many...

adamalton:
Now refactored as per Alan's suggestion, and tests added for additional fun and profit.

francois-potato (approved):

packages/w3/package.json
```diff
@@ -26,6 +26,7 @@
     "@ipld/car": "^3.1.16",
     "conf": "^10.1.1",
     "enquirer": "^2.3.6",
+    "fs": "^0.0.1-security",
```
alanshaw, line 29:
```suggestion
```

Issue #2030, opened 2022-10-13
Ensure web analytics receives needed KPI data & prototype dashboards


JeffLowe:
Initial request was to vet and ensure Countly analytics system was connected throughout sites.  However, decision was made to move away from countly and find/use an alternative.  Plausible and Simple are being vetted.  One should be selected and used. 
Eg current Countly Metrics ([https://countly.ipfs.io/dashboard#/60fe7ce4d7c3b807836603d7/](https://countly.ipfs.io/dashboard#/60fe7ce4d7c3b807836603d7/))

- [x] vet Plausible analytics
- [ ] vet Simple Analytics
- [ ] ensure we use a Dag.House account (opposed to  a protocol labs account)
- [ ] Ensure this Data is present in analytics system so that it can be easily tracked
        - Heaviest web traffic sources
        - Time spent on docs
        - Blog post visits / time spent on post
        - Blog subscriptions

- [ ] Ensure we are capturing the data for, and the following funnels are clear in report
        - Conversion / churn from new users landing pre-login and creating account (also should include time to create account)
        - Conversion / churn rate for new users from pre-login pricing page to creating account to putting in their credit card to signing up for a plan

- [x]  Resolve Weird buggy things (likely requires ensuring that cta's pass more info into countly to clearly label what was clicked...at minimum)
        - What are the steps in this funnel? [[https://countly.ipfs.io/dashboard#/60fe7ce4d7c3b807836603d7/funnels/detail/59297a90c81b7c19b27c63c0a6620079/{}](https://countly.ipfs.io/dashboard#/60fe7ce4d7c3b807836603d7/funnels/detail/59297a90c81b7c19b27c63c0a6620079/%7B%7D)](https://countly.ipfs.io/dashboard#/60fe7ce4d7c3b807836603d7/funnels/detail/59297a90c81b7c19b27c63c0a6620079/%7B%7D)
- [ ] create minimal dashboards in analytics system displaying above KPIs so others can go in and know how to build the ones they need.

JeffLowe:
@dchoi27 Do we also need to build out minimal dashboards in Countly, or  just ensure the data get's into Countly correctly so it can be used/referenced?

dchoi27:
minimal dashboards would be appreciated so others can go in and know how to build the ones they need to (sometimes the tag names are hard to decipher)

JeffLowe:
[...]

PR #2110, opened 2022-11-17
feat: simple analytics

cmunns:
Swap countly for simple analytics.


cmunns:
Here is a custom event getting fired. I was unable to test implementation on safari and firefox bc of a cors error which I believe will be solved on prod.
Screenshot 2022-11-21 at 12 35 18 PM

johnathonroach (approved):

hugomrdias:
do we have any write up about simple vs plausible ?

hugomrdias:
@cwaring this may be interesting original issue https://github.com/web3-storage/web3.storage/issues/2030

packages/website/modules/zero/components/accordion/accordionSection.js
```diff
@@ -44,6 +45,15 @@ function AccordionSection({ active, toggle, toggleOnLoad, reportUID, slug, disab
     reportUID(uid);
   }, [reportUID, uid]);
 
+  useEffect(() => {
+    if (open && header) {
+      const headerLabel = header.props.children.find(child => {
```
travis, line 50: this is a litttttle funky - hard for me to tell whether it would be possible to use a `Ref` (or whether that would be cleaner) but curious if you considered that rather than searching through children? not a blocker, but figured it's worth raising for posterity

packages/website/pages/_app.js
```diff
@@ -27,6 +28,16 @@ const App = ({ Component, pageProps }) => {
     <AppProviders authorizationProps={{ ...pageProps }}>
       <Metadata {...pageProps} />
       <RestrictedRoute {...pageProps}>
+        <Script src="https://track.web3.storage/latest.js" data-hostname="web3.storage" data-ignore-pages="/callback" />
```
travis, line 31: what's the best place to understand what these scripts are doing? it would be helpful to have a comment in here with a pointer for folks who come later and are trying to figure out where scripts come from and how they work

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
```
travis, line 1: should this be checked in? genuinely unfamiliar with conventions for this team/repo so this is a "I am not sure, teach me" question not a "I am skeptical" one ;-p

travis (approved):
lgtm - had a couple questions but they are mostly for my information - happy to see this merge as-is if you think that makes sense!

.vscode/settings.json
cmunns: I actually have a similar question. I was getting vscode plugin conflicts on all the imports and this seemed to solve it. I thought it may help but others but if this was just me I probably should ignore the settings from getting checked in.

Issue #1953, opened 2022-09-26
Possible Repainting issue on website impacting performance


JeffLowe:
Something appears to be making the browser repaint the entire window when i scroll, which gives the site an unstable feeling.

This seems to only be a noticeable issue on the web3.storage website, not the blog nor the docs sites. 

## Results from quick device testing
- Not as noticeable on iphone 8 ios 15.6.1 Safari (actually, barely perceptable)
- Same device( iphone 8 ios 15.6.1), Chrome browser more noticeable.
- Pixel6 Chrome latest, probably the most noticeable.  
- Pixel 6 firefox, I cannot see an issue
- Desktop macos 12.6 Chrome latest, I see the issue on web3 website when I scroll quickly.
- Desktop macos 12.6 Safari, the web3 website repainting issue very noticeable while scrolling.

JeffLowe:
One cause is likely the json rendering.  Will need to factor that out from legacy code.

JeffLowe:
@e-schneid reference related fix on blog https://github.com/web3-storage/web3.storage-content/pull/56 & https://github.com/web3-storage/web3.storage-content/pull/48

PR #2115, opened 2022-11-18
fix: begin refactoring away from json content files

e-schneid:
See [ticket](https://github.com/web3-storage/web3.storage/issues/1953)

This is an initial pass to begin removing the json content files (and BlockBuilder, etc) in favor of more standard React, to make it a bit easier for developers to make changes and also hopefully improve performance.  

To facilitate smaller components, this also creates a new `page-components` directory to keep components that are specifically for use on a given route (in Next.js we don't want to these files living under `/pages`).


packages/website/components/layout/index.js
```diff
@@ -0,0 +1,77 @@
...
+/**
+ *
+ * @param {Object} props
+ * @param {React.ReactNode} props.children
+ * @param {string} [props.variant]
+ * @param {string} [props.id]
+ *
+ * @returns
+ */
+export function Sectionals({ children, variant, id }) {
```
e-schneid, line 16: The existing code used the name "sectionals," and then each of these contained a number of "sectional" components, and within "sectional" you'll find "column."  The naming needs a little help, I think, and I'm open to ideas.  Maybe "Container" for the outermost box, then "Section," and then "Column"?

Issue #2117, opened 2022-11-21
Improve the handling of the PSA authorisation error in the FE


alexandrastoica:
Currently when a user does not have pinning access, the error is displayed in console.

We should improve the UX to either show how to gain such access and handle the access error more gracefully.

Issue #1947, opened 2022-09-26
Redundant call made to fetch pins on account page load


joshJarr:

mbommerez:
Note: we need to double check these criteria before starting work on them

alexandrastoica:
Last two points are in https://github.com/web3-storage/web3.storage/pull/2147

PR #2122, opened 2022-11-24
fix: do not make pins request if the user does not have access #2117

alexandrastoica:


adamalton (approved):
LGTM

alanshaw:
The server should be verifying this. Is this not the case already?

The user should not reach this view if they don't have access - they won't see the tabs and the table will not be loaded.

alexandrastoica:
@alanshaw thanks for the review, some context:

> The server should be verifying this. Is this not the case already?

Yes it is.

> The user should not reach this view if they don't have access - they won't see the tabs and the table will not be loaded.

The user does reach the view, more specifically the `PinRequestsTable` in `FileManager` is displayed if pin requests count > 0. To get the count, the `FileManager` uses the `usePinRequests` hook which tries to fetch the pin requests and throws an error (which is not handled) if the user does not have access. This leads to an error in the console log in prod which I was trying to avoid for the meantime with this PR, while I was working on https://github.com/web3-storage/web3.storage/issues/1947. 

I marked this as draft and I'm doing some better improvements here https://github.com/web3-storage/web3.storage/pull/2147, let me know if there's anything that doesn't make sense, or any suggestions :)

PR #2124, opened 2022-11-25
feat: allow to pass mailprovider as a workflow input

flea89:
Add option to run email storage cron job without sending real emails.


adamalton (approved):
LGTM

PR #2125, opened 2022-11-28
fix: allow update PSA pin requests with same CID

### LINKED ISSUE #1547: It should be possible to update/replace a pin with the same CID
There are some cases where it's desired to update the `name`, `meta`, and `origins` fields of a pin request with a given CID using the replace pin endpoint.

Currently, the web3.storage pinning API returns the following 400 error when attempting to update those fields with the same CID.

https://github.com/web3-storage/web3.storage/blob/bfa88cbb330274b287de90ca7d4b244919f18a5f/packages/api/src/pins.js#L253-L256

Related: https://github.com/ipfs-shipyard/pinning-service-compliance/issues/124

flea89:
Resolve #1547


alexandrastoica:
LGTM

alanshaw (approved):
LGTM

PR #2147, opened 2022-12-05
fix: improve pins vs uploads experience on account page #2117 #1947

alexandrastoica:
Please also see https://github.com/web3-storage/web3.storage/issues/1947 for reference.

This PR: 
- avoids fetching pin requests when the user does not have PSA access enabled (see https://github.com/web3-storage/web3.storage/issues/2117)
- displays pinning table in the user account page if the user has PSA access, even though there's 0 pins currently, so that the user is aware they can access the pinning service
-  fixes bug where the table tab is selected based on the param `?table=pinned`: the previous implementation was not waiting until the pin request was done, hence removing the param and not selecting the table. This PR fixes it by checking if the user has PSA access enabled and setting the tab accordingly.


adamalton (approved):
This looks great. I've tested it locally and it meets all the user story requirements.

PR #2148, opened 2022-12-05
feat: listen for "Enter" key on login input field

### LINKED ISSUE #2143: "Sign Up/ Login" Button should react to enter key
When I'm entering my email in the "Enter your email" field, I want to be able to press enter to trigger the "Sign Up/ Login" button.

yusefnapora:
This closes #2143 by adding an `onKeyDown` handler to the email input field on the login form. If the Enter key was pressed, it calls `onLoginWithEmail` to trigger the login request.

### DIFF
--- a/packages/website/pages/login/index.js
+++ b/packages/website/pages/login/index.js
@@ -91,6 +91,7 @@ const Login = () => {
             className={clsx('login-email', errors.email && 'error')}
             placeholder={pageContent.form_placeholder}
             onChange={useCallback(e => setFormData({ email: e.currentTarget.value }), [])}
+            onKeyDown={e => e.key === 'Enter' && onLoginWithEmail()}
           />
           <Button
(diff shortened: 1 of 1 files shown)


packages/website/pages/login/index.js
```diff
@@ -91,6 +91,7 @@ const Login = () => {
             className={clsx('login-email', errors.email && 'error')}
             placeholder={pageContent.form_placeholder}
             onChange={useCallback(e => setFormData({ email: e.currentTarget.value }), [])}
+            onKeyDown={e => e.key === 'Enter' && onLoginWithEmail()}
```
alanshaw, line 94: This should be in a `<form>` element with an `onSubmit` handler. Then the key press listener is not required.

alanshaw (changes requested):

PR #2151, opened 2022-12-06
fix: bump 'playwright' version & fix intermittent e2e test failure

adamalton:
A story in two parts…

### Part 1 - Playwright installation was failing

This PR was started to fix an error with `playwright install-deps` which is showing up here: https://github.com/web3-storage/web3.storage/actions/runs/3620217854/jobs/6124601471

It seems that the dependencies which are required by the version of Playwright that we were using are no longer available on `ubuntu-latest` because `ubuntu-latest` is now a newer version. When that changed, this broke (is my educated guess). There are vague clues pointing at this being the case in this issue: https://github.com/microsoft/playwright/issues/19234

Bumping the playwright version seems to fix it. So that's the first part of this PR.

But fixing that made the Playwright tests run again. Which then uncovered another problem…

### Part 2 - The Playwright tests were failing

The `test-e2e` step in `.github/workflows/website.yml` was failing for the Firefox value of the matrix with this error:

```
page.goto: NS_ERROR_ABORT
    =========================== logs ===========================
    navigating to "http://localhost:4000/", waiting until "load"
    ============================================================

       7 |
       8 | test.beforeEach(async ({ page }) => {
    >  9 |   await page.goto('/');
         |              ^
      10 | });
      11 |
      12 | test.describe('/account/payment', () => {
```

[Example CI run](https://github.com/web3-storage/web3.storage/actions/runs/3630975522/jobs/6140045696).

After much confusion (why did it only fail for Firefox and not Chrome? why did it only fail for the first test? how come you can see many successful requests being made to the server before that error occurs?), the painfully simple answer is:

After running `npm start` we weren't waiting for the servers (web or API) to be ready before we ran the browser tests.

So when the `beforeEach` hook ran on the first test, it got that error. The test would be retried (including retrying the failed `beforeEach` hook), the retries would also fail (because the server still wasn't ready), then the other tests would run and pass (server now ready), then only after all the tests had run did you see the output of the one that failed. In Chrome, the test was marked as flakey (i.e. it passed after one or more retries). Presumably Chrome was taking longer to start up, giving the server more time to be ready, and hence it passed on the retry.

### Happily ever after

This PR fixes both of the above issues. Yay! It also adds some improved logging so that you can see a lot more about what's happening when the e2e tests are being run, which should make debugging of this stuff much easier in future.

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -116,7 +116,7 @@ jobs:
 
       # add playwright-report to web3.storage
       - name: add playwright-report to web3.storage
-        if: inputs.add_playwright_report_to_web3_storage && steps.check_website_playwright_report.outputs.files_exists == 'true'
+        if: ${{ inputs.add_playwright_report_to_web3_storage }} && steps.check_website_playwright_report.outputs.files_exists == 'true'
         uses: web3-storage/add-to-web3@v2
         id: add_playwright_report_to_web3_storage
         with:
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -88,7 +88,7 @@ jobs:
 
       - run: npx playwright install --with-deps ${{ matrix.browser }}
         working-directory: packages/website
-      
+
       # run e2e tests
       - id: website_test_e2e
         continue-on-error: true # when this fails, we still want to do some cleanup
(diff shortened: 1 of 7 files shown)


adamalton:
Now that the installation of Playwright is fixed, we're getting an error running the browser tests. At first it looks like it's just Firefox that's failing, but actually, if you look at the [report for Firefox](https://bafybeib7ct2y6735ecmkhwcmqpezyndq4gpxrh7vkee62olqxqmbeq3baa.ipfs.w3s.link/#?testId=8620d1f3ef5a3848857b-186294de2ad8169151f6) and the [report for Chromium](https://bafybeichduddcmmarnnqo23j5u3xdsplhzncveddysepwcnbyytbnp5enu.ipfs.w3s.link/#?testId=8620d1f3ef5a3848857b-cd65b748a17e2c33c49c) they're both getting the same issue.

I _think_ that the problem is that it's timing out waiting for the root of the site to load, but I'm not 100% sure.

yusefnapora (approved):
Awesome work, thanks @adamalton!

PR #2152, opened 2022-12-06
feat: allow displaying of custom stripe prices

e-schneid:
This continues to only allow the _selection_ of a standard plan, but if a user is associated with a non-standard plan in Stripe we will now return this data (specifically tier information and metadata) to the UI in order to populate custom cards.  This will be necessary as we migrate early adopters to custom plans as well as for custom plan functionality in general.

image

This also includes an API change to ensure that all of our users have at least a subscription to the Free tier.  We want to migrate away from modeling early adopters as users without a subscription and instead explicitly put them into a tier.

Update UIs
See [here](https://www.notion.so/UIs-for-custom-plans-post-migration-df237239d19049e999e2284478ed372d)

- [x] Handle description metadata from Stripe
- [x] Include Stripe metadata to prevent selecting other plans in the UI
- [x] Add "View your plan details here." text

### DIFF
--- a/packages/api/src/utils/billing-types.ts
+++ b/packages/api/src/utils/billing-types.ts
@@ -66,7 +66,8 @@ export interface PriceTier {
 
 export interface CustomStoragePrice {
   id: string;
-  bandwidth?: string;
+  // Monthly bandwidth in GiB
+  bandwidth?: number;
   label: string;
   // True if price is prioritized.  It could be used to disable selection of alternative prices.
   isPreferred: boolean;
--- a/packages/api/src/utils/billing-types.ts
+++ b/packages/api/src/utils/billing-types.ts
@@ -59,7 +59,22 @@ export interface CustomersService {
 
 export type StoragePriceName = 'free' | 'lite' | 'pro'
 
-export type StoragePrice = StoragePriceName | Pick<Stripe.Response<Stripe.Price>, 'id'|'metadata'|'tiers'>
+export interface PriceTier {
+  flatAmount: number | null;
+  unitAmount: number | null;
+  upTo: number | null;
+}
+
+export interface CustomStoragePrice {
+  id: string;
+  bandwidth?: string;
+  label: string;
+  isPreferred: boolean;
+  description?: string;
+  tiers: Array<PriceTier> | null;
+}
+
+export type StoragePrice = StoragePriceName | CustomStoragePrice;
 
 /**
  * A subscription to the web3.storage platform.
--- a/packages/api/src/utils/billing-types.ts
+++ b/packages/api/src/utils/billing-types.ts
@@ -68,6 +68,7 @@ export interface CustomStoragePrice {
   id: string;
   bandwidth?: string;
   label: string;
+  // True if price is prioritized.  It could be used to disable selection of alternative prices.
   isPreferred: boolean;
   description?: string;
   tiers: Array<PriceTier> | null;
--- a/packages/api/src/utils/billing-types.ts
+++ b/packages/api/src/utils/billing-types.ts
@@ -1,4 +1,5 @@
 import { StripePriceId } from "./stripe";
+import Stripe from 'stripe';
 
 export type StripePaymentMethodId = string;
 export type CustomerId = string;
@@ -58,6 +59,8 @@ export interface CustomersService {
 
 export type StoragePriceName = 'free' | 'lite' | 'pro'
 
+export type StoragePrice = StoragePriceName | Pick<Stripe.Response<Stripe.Price>, 'id'|'metadata'|'tiers'>
+
 /**
  * A subscription to the web3.storage platform.
  * This may be a composition of several product-specific subscriptions.
@@ -66,12 +69,12 @@ export interface W3PlatformSubscription {
   // details of subscription to storage functionality
   storage: null | {
     // the price that should be used to determine the subscription's periodic invoice/credit.
-    price: StoragePriceName
+    price: StoragePrice
   }
 }
 
 export type NamedStripePrices = {
-  priceToName: (priceId: StripePriceId) => undefined | StoragePriceName
+  priceToName: (priceId: string) => undefined | StoragePriceName
   nameToPrice: (name: StoragePriceName) => undefined | StripePriceId
 }
 
--- a/packages/api/src/utils/billing-types.ts
+++ b/packages/api/src/utils/billing-types.ts
@@ -1,5 +1,4 @@
 import { StripePriceId } from "./stripe";
-import Stripe from 'stripe';
 
 export type StripePaymentMethodId = string;
 export type CustomerId = string;
--- a/packages/website/components/account/currentBillingPlanCard/currentBillingPlanCard.js
+++ b/packages/website/components/account/currentBillingPlanCard/currentBillingPlanCard.js
@@ -1,27 +1,30 @@
+import { formatAsStorageAmount, formatCurrency } from '../../../lib/utils';
+
 const CurrentBillingPlanCard = ({ plan }) => {
   return (
     <div className="billing-card card-transparent">
       {plan !== undefined && (
         <div key={plan.title} className="billing-plan">
           <div className="billing-plan-overview">
-            <h4 className="billing-plan-title">{plan.title}</h4>
-            <div className="billing-plan-amount">{plan.price}</div>
+            <h4 className="billing-plan-title">{plan.metadata['UI Label']}</h4>
+            <div className="billing-plan-amount">{formatCurrency(plan.tiers[0]?.flat_amount / 100, true)}/mo</div>
           </div>
 
-          <p className="billing-plan-desc">{plan.description}</p>
+          <p className="billing-plan-desc">{plan.metadata['Description']}</p>
           <p className="billing-plan-limit">
             {plan.id !== 'free' && (
               <span>
-                <span>Base Storage Capacity:</span> <strong>{plan.baseStorage}</strong> for{' '}
-                <strong>{plan.price}</strong>
+                <span>Base Storage Capacity:</span> <strong>{formatAsStorageAmount(plan.tiers[0]?.up_to)}</strong> for{' '}
+                <strong>{formatCurrency(plan.tiers[0]?.flat_amount / 100, true)}/mo</strong>
               </span>
             )}
             <span>
-              <span>Additional Storage:</span> <strong>{plan.additionalStorage.split(' ')[0]} per GiB</strong> after{' '}
-              {plan.baseStorage.split(' ')[0]}
+              <span>Additional Storage:</span>{' '}
+              <strong>{formatCurrency(plan.tiers[1]?.unit_amount / 100)} per GiB</strong> after{' '}
+              {formatAsStorageAmount(plan.tiers[0]?.up_to)}
             </span>
             <span>
-              <span>Bandwidth:</span> <strong>{plan.bandwidth}</strong>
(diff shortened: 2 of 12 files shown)


jsdevel (approved):

packages/website/lib/utils.js
```diff
@@ -12,6 +12,31 @@ export const formatTimestamp = timestamp => {
   return formattedDate;
 };
 
+/**
+ * Returns the amount in TiBs if the amount is a multiple of 1024, otherwise returns the amount in GiBs
+ * @param {number} amountInGiB
+ * @returns {string}
+ */
+export const formatAsStorageAmount = amountInGiB =>
+  amountInGiB % 1024 === 0 ? `${amountInGiB / 1024} TiB` : `${amountInGiB} GiB`;
```
drewdelano, line 21:
```suggestion
  amountInGiB > 1024 ? `${amountInGiB / 1024} TiB` : `${amountInGiB} GiB`;
```

drewdelano: maybe I'm missing the why here...but this seems like something where once it goes past 1 TiB you only would want to measure it in TiB

e.g. 5.1 TiB vs 5,121 GiB

drewdelano: Or are we not worried about this being a problem because it's all from data we can manicure?

packages/api/src/utils/billing-types.ts
```diff
@@ -58,6 +59,8 @@ export interface CustomersService {
 
 export type StoragePriceName = 'free' | 'lite' | 'pro'
 
+export type StoragePrice = StoragePriceName | Pick<Stripe.Response<Stripe.Price>, 'id'|'metadata'|'tiers'>
```
gobengo, line 62: Can you do this without depending on `stripe` package for types, that way this API can be implemented against non-stripe billing providers?

That way we can also hide stripe-specific implementation details from consumers of this HTTP API, e.g. from the frontend. i.e. in this PR the frontend code will throw runtime errors if the custom storage price object `metadata` property doesn't have keys for 'UI Label', 'Description', and 'Bandwidth'. We want  stripe-specific parsing happens in order to produce something in `billing-types` so that anything consuming billing-types (e.g. our frontend) is abstracted from stripe. It also allows us to keep the api json objects using consistent naming e.g. thisCase vs snake_case.

One way of doing that is to define a `CustomStoragePrice` type with eactly what you need:

```
type CustomStoragePrice = { 
  // in StripeSubscriptionsService, parse this from metadata['UI Label']
  name: string; 
  // in StripeSubscriptionsService, parse this from metadata['Description']
  description: string;
  // parse from stripe 'tiers' info
  tiers: Array<{ unit: 'byte', max: number, price: ['usd', number] }>
}

and in the response
```
price: StoragePriceName | CustomStoragePrice
```

and then no need to have this file import from npm package `stripe`.

File: packages/website/components/account/currentBillingPlanCard/currentBillingPlanCard.js [hunk 3]
```diff
@@ -1,27 +1,30 @@
...
+
 const CurrentBillingPlanCard = ({ plan }) => {
   return (
     <div className="billing-card card-transparent">
       {plan !== undefined && (
         <div key={plan.title} className="billing-plan">
           <div className="billing-plan-overview">
-            <h4 className="billing-plan-title">{plan.title}</h4>
-            <div className="billing-plan-amount">{plan.price}</div>
+            <h4 className="billing-plan-title">{plan.metadata['UI Label']}</h4>
```
gobengo, line 9: Will you encapsulate the stripe metadata processing inside `StripeSubscriptionService`? More here https://github.com/web3-storage/web3.storage/pull/2152/files#r1065213416

If we do that, then:
1) this frontend code has no stripe specifics, it only depends on billing-types types
2) this code won't have errors like:
    * If someone sets a customer to a custom price, but sets metadata 'UI label' instead of 'UI Label', the end-user will see 'undefined' here
    * if someone sets a customer to a custom price with no tiers, the end-user will see '/mo'

gobengo: Ideally whenever someone changes a 'custom price' in stripe (including metadata), we either:
* get a big loud error via sentry as soon as `StripeSubscriptionService` encounters it an tries to parse the price into a `StoragePrice`
* get no error and feel confident that everyone consuming the StoragePrice has everything they will need (i.e. no need here to handle case where `tiers` is an empty array by putting optional chaining operators in here, so need for this ui code to handle `NaN` currencies in formatCurrency)

packages/website/components/account/paymentTable.js/paymentTable.js
```diff
@@ -32,30 +50,36 @@ const PaymentTable = ({ plans, currentPlan, setPlanSelection, setIsPaymentPlanMo
             </div>
             {plans.map(plan => (
               <div
-                key={plan.title}
-                className={`billing-card card-transparent ${currentPlan?.id === plan.id ? 'current' : ''}`}
+                key={plan.metadata['UI Label']}
```
gobengo, line 53: since metadata is loosely typed, this `plan.metadata['UI Label']` can be undefined, and then weird things may start happening in the loop with many `key={undefined}`

packages/website/components/account/paymentTable.js/paymentTable.js
```diff
@@ -1,14 +1,32 @@
...
-const PaymentTable = ({ plans, currentPlan, setPlanSelection, setIsPaymentPlanModalOpen }) => {
   return (
     <>
       {currentPlan && (
         <p className="billing-content-intro" data-testid="currentPlanIndicator">
           <span>
-            Your current plan is: <strong data-testid="currentPlan.title">{currentPlan.title}</strong>
+            Your current plan is:{' '}
+            <strong data-testid="currentPlan.metadata['UI Label']">
+              {currentPlan?.metadata['UI Label'] ?? 'Custom'}
```
gobengo, line 28: Can you avoid adding metadata-dependent view logic in the jsx here by first parsing the thing-with-metadata to something else with explicit typing so you dont need to put defaults like 'Custom' in here?

packages/website/components/contexts/plansContext.js
```diff
@@ -17,77 +17,94 @@
...
- * @property {StoragePrice|EarlyAdopterPlanId} id
- * @property {string} title
- * @property {string} description
- * @property {string} price
- * @property {string} baseStorage
- * @property {string} additionalStorage
- * @property {string} bandwidth
- * @property {string} blockLimit
+ * @property {string} id
+ * @property {import("@stripe/stripe-js").Metadata} metadata
```
gobengo, line 30: I think it was better before when this was using a plan type that doesn't depend on stripe, and isn't as vaguely defined as `Stripe.Metadata`

packages/website/lib/utils.js
```diff
@@ -12,6 +12,31 @@ export const formatTimestamp = timestamp => {
...
+/**
+ * Formats an amount (in dollars) as a currency
+ * @param {number} amount
+ * @param {boolean} shouldTruncateToDollar
+ * @returns {string}
+ */
+export const formatCurrency = (amount, shouldTruncateToDollar = false) => {
+  if (isNaN(amount)) return '';
+
+  const formatter = new Intl.NumberFormat('en-US', {
```
gobengo, line 32:
```suggestion
  const formatter = new Intl.NumberFormat(undefined, {
```

Do we want to always use that locale? I thought the point of `Intl` was to use the locale of the user agent (browser)

gobengo (changes requested):

packages/website/lib/utils.js
e-schneid: My thought was it would be a loss of precision, but we don't currently have plans over 1TiB that aren't divisible by 1024.  I'll update it to display an integer in TiB if divisible by 1024, otherwise I'll round to 2 decimal places

packages/website/components/account/currentBillingPlanCard/currentBillingPlanCard.js
e-schneid: > get no error and feel confident that everyone consuming the StoragePrice has everything they will need

I was opting for this one.  

That's related to my original rationale behind not doing any additional mapping.  This metadata doesn't pertain to any business rules and really only relates to view logic and text.  Our API, in my original opinion, was kept more flexible if staying agnostic to what's actually kept in this metadata.  Mapping this data seemed like unnecessary obfuscation, and if we switch to another billing provider later then presumably they'd offer a similar way to store this sort of metadata which could _then_ be mapped to the shape of Stripe's metadata (i.e., it seemed premature to map it to our own interface).

But that said, I also agree with what you're saying and so I wrote it that way as well (or similar to -- let me know if there's anything I missed in your intent), and it does make things tidier in the UI.

packages/website/components/account/paymentTable.js/paymentTable.js
e-schneid: updated (see [here](https://github.com/web3-storage/web3.storage/pull/2152#discussion_r1065217803))

packages/website/components/contexts/plansContext.js
e-schneid: updated (see [here](https://github.com/web3-storage/web3.storage/pull/2152#discussion_r1065217803))

packages/api/src/utils/billing-types.ts
e-schneid: updated (see [here](https://github.com/web3-storage/web3.storage/pull/2152#discussion_r1065217803))

packages/api/src/utils/billing-types.ts
```diff
@@ -58,6 +59,23 @@ export interface CustomersService {
...
+  flatAmount: number | null;
+  unitAmount: number | null;
+  upTo: number | null;
+}
+
+export interface CustomStoragePrice {
+  id: string;
+  bandwidth?: string;
+  label: string;
+  isPreferred: boolean;
```
gobengo, line 71: Will you add a comment explaining what this means?

packages/api/src/utils/billing-types.ts
```diff
@@ -58,6 +59,23 @@ export interface CustomersService {
...
 
+export interface PriceTier {
+  flatAmount: number | null;
+  unitAmount: number | null;
+  upTo: number | null;
+}
+
+export interface CustomStoragePrice {
+  id: string;
+  bandwidth?: string;
```
gobengo, line 69: Will you add a comment explaining what this means? I'm a little surprised the bandwidth is expressed as a string vs something like a number amount + (explicit or implicit) a unit involving bytes and time.

e.g.
```
bandwidth: {
  monthlyLimit: { bytes: number }
}
```

packages/website/components/account/currentBillingPlanCard/currentBillingPlanCard.js
gobengo: >  This metadata doesn't pertain to any business rules and really only relates to view logic and text.

If that's true, then one might argue the website-specific view data shouldn't be stored in stripe metadata. Otherwise someone may come along and read/write from that metadata without knowing it's tightly coupled to the website frontend requirements.
I think it'd be ideal if we treat data in stripe *as business data* and not just view-related, but I hear that wasn't your intent, though, and that you prefer to do it another way.

gobengo (approved):

packages/api/src/utils/billing-types.ts
e-schneid: updated

packages/api/src/utils/billing-types.ts
e-schneid: since we're mapping this, I agree that it does make sense as a number now, and we may use this bandwidth later to control more than just the display (supporting what you were saying before re: business logic etc)

PR #2159, opened 2022-12-12
fix: reject not supported codecs

### LINKED ISSUE #2154: pinning service API needs to reject IPLD codecs we don’t support
- pinning service API needs to reject IPLD codecs we don’t support (for both web3.storage and NFT.Storage)
-  document that we support remote pinning the following codecs raw dag-json dag-cbor dag-pb in the root node and outside of that, we’ll reject (here https://web3.storage/docs/how-tos/pinning-services-api/ and https://nft.storage/docs/how-to/pinning-service/)
- also add to the API reference docs of each
- easiest way we can get this enforced in the API for now is to return non-200 responses to the pinning service API  (decode the cid in full, get its codec, compare, reject)

flea89:
Resolves #2154

### DIFF
--- a/packages/api/src/utils/psa.js
+++ b/packages/api/src/utils/psa.js
@@ -165,7 +165,7 @@ export function transformAndValidate (payload) {
       if (!SUPPORTED_CODECS.includes(c.code)) {
         let message = PSACodecNotSupported.MESSAGE
         if (c.code === libp2pKeyCodec) {
-          message = `${message} If you're trying to pin using an IPNS record that isn't supported yet. Please +1 [this githib issue](https://github.com/web3-storage/web3.storage/issues/2155) if you want it to be.`
+          message = `${message} If you're trying to pin using an IPNS record that isn't supported yet. Please +1 [this github issue](https://github.com/web3-storage/web3.storage/issues/2155) if you want it to be.`
         }
         throw new PSACodecNotSupported(message)
       }
--- a/packages/api/src/utils/psa.js
+++ b/packages/api/src/utils/psa.js
@@ -1,6 +1,18 @@
 import { Validator } from '@cfworker/json-schema'
-import { PSAErrorInvalidData, PSAErrorRequiredData } from '../errors.js'
+import { PSACodecNotSupported, PSAErrorInvalidData, PSAErrorRequiredData } from '../errors.js'
 import { normalizeCid } from '../utils/cid.js'
+import * as raw from 'multiformats/codecs/raw'
+import * as pb from '@ipld/dag-pb'
+import * as dagJson from '@ipld/dag-json'
+import * as dagCbor from '@ipld/dag-cbor'
+import { CID } from 'multiformats/cid'
+
+const SUPPORTED_CODECS = [
+  raw.code,
+  pb.code,
+  dagCbor.code,
+  dagJson.code
+]
 
 /**
  * @typedef {'queued' | 'pinning' | 'failed' | 'pinned'} apiPinStatus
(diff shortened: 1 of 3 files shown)


packages/api/test/pin.spec.js
```diff
@@ -562,6 +568,53 @@ describe('Pinning APIs endpoints', () => {
       assert.strictEqual(error.details, INVALID_CID)
     })
 
+    it('throws error if not supported codec', async () => {
+      const hash = await sha256.digest(Buffer.from(`${Math.random()}`))
```
francois-potato, line 572: Nit: could use `toString()` to here too. (same line 599)

packages/api/src/utils/psa.js
```diff
@@ -132,11 +149,25 @@ export function transformAndValidate (payload) {
   // Validate CID.
   if (cid) {
     try {
-      normalizedCid = normalizeCid(cid)
-      opts.cid = cid
+      let c
```
francois-potato, line 152: Nit; I find the following code not easy to read. Not sure why the declaration of c is made here and not in the block below. Could we use only one try? Maybe define a type for the returned struct?

packages/api/test/pin.spec.js
```diff
@@ -562,6 +568,53 @@ describe('Pinning APIs endpoints', () => {
...
+      const cid = CID.create(1, 72, hash).toString()
+
+      const res = await fetch(new URL('pins', endpoint).toString(), {
```
francois-potato, line 575: Nit: fetch can use a URL object (no need to `toString`). (same line 603)

francois-potato (approved):
LGTM

packages/api/src/utils/psa.js
```diff
@@ -132,11 +149,25 @@ export function transformAndValidate (payload) {
...
+        normalizedCid = c.toV1().toString()
+        opts.cid = cid
+      } catch (e) {
+        throw new PSAErrorInvalidData(INVALID_CID)
+      }
+
+      if (!SUPPORTED_CODECS.includes(c.code)) {
+        let message = PSACodecNotSupported.MESSAGE
+        if (c.code === 72) {
+          message = `${message} If you're trying to pin using an IPNS record that isn't supported yet. Please +1 [this githib issue](https://github.com/web3-storage/web3.storage/issues/2155) if you want it to be.`
```
blucas, line 164:
```suggestion
          message = `${message} If you're trying to pin using an IPNS record that isn't supported yet. Please +1 [this github issue](https://github.com/web3-storage/web3.storage/issues/2155) if you want it to be.`
```
Does the client parsing this error message support markdown syntax for hyperlinks?

packages/api/src/utils/psa.js
```diff
@@ -132,11 +149,25 @@ export function transformAndValidate (payload) {
...
+        c = CID.parse(cid)
+        normalizedCid = c.toV1().toString()
+        opts.cid = cid
+      } catch (e) {
+        throw new PSAErrorInvalidData(INVALID_CID)
+      }
+
+      if (!SUPPORTED_CODECS.includes(c.code)) {
+        let message = PSACodecNotSupported.MESSAGE
+        if (c.code === 72) {
```
blucas, line 163: Should `72` be documented somewhere to identify its meaning?

codeflyer: I'll add it to a constant, to improve the readability.

flea89: Good point, saved it in a constant and added a bit of explanation.

packages/api/src/utils/psa.js
flea89: Good point, while obv this API can be used directly, I think it's safe to expect most of the requests would probably come from Kubo. And I don't know if it does support it.
I wonder if it's a real problem thogh. Even if it doesn't, Markdown syntax for links it's quite common anyway, and I expect users to be able to still understand and navigate to the issue.

What do you think?

codeflyer (approved):
LGTM

adamalton (approved):

alanshaw (approved):

PR #2174, opened 2022-12-30
feat: increase timeout for pins cron

ericssontadashi:


dchoi27 (approved):

.github/workflows/cron-pins.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         env: ['staging', 'production']
-    timeout-minutes: 30
```
vasco-santos, line 15: it will run each 30 minutes, so we will start having multiple jobs at same time. See line 5

vasco-santos (changes requested):

vasco-santos (approved):

olizilla:
The cron job to update pin status is set up so it basically runs all the time. It times out after 30mins and runs every 30mins. We can move it to run for 60mins and timeout every 60mins but it's basically the same. The idea is that the job picks up where it left off, updating the db as it goes, so the interval isn't too important. The key point is we expect the job to timeout out most of the time. The subsequent run just picks up where it left off.

vasco-santos:
@olizilla that might have been true at some point. However, we did do changes in the past to make it not the case anymore. Like moving heavy queries to the replica and removing Pin status remote rows.

If you see the github action, it even has an heartbeat which will throw alerts in case of failures. I agree that moving 30-60 minutes might be the same, and we probably just need to get some backlog catched up.

alanshaw:
Also github cron is approximate and often starts after the scheduled time, increasing the run time should reduce the number of periods when we're waiting on github to start the job, giving us more time to process.

PR #2179, opened 2023-01-05
feat: use pickup as `CLUSTER_API_URL`

olizilla:
Update `CLUSTER_API_URL` in wrangler.toml to have the api use pickup in place of cluster for handling pinning service requests.

The prod pickup is currently configured to proxy all requests back over to cluster. The plan is to get pickup in the loop so we can control what fraction of requests it should handle from pickyup config.

Note: Merging this PR will trigger a staging deployment, which will apply the change to the staging config. Merging the release-please PR will trigger a prod deploy and the prod config change will be applied then.

TODO
- [ ] update `CLUSTER_API_URL` secret on repo so cron-pins uses the same endpoint as the web3.storage api worker.
- [x] implement handler in pickup for `GET /pins?cids=baf1,...,bafN`

License: MIT

### DIFF
--- a/packages/api/wrangler.toml
+++ b/packages/api/wrangler.toml
@@ -40,7 +40,7 @@ r2_buckets = [
 # name = "web3-storage-production"
 account_id = "fffa4b4363a7e5250af8357087263b3a" # Protocol Labs CF account
 route = { pattern = "https://api.web3.storage/*", zone_id = "7eee3323c1b35b6650568604c65f441e" }
-vars = { CARPARK_URL = "https://carpark.web3.storage", CLUSTER_API_URL = "https://web3.storage.ipfscluster.io/api/", ENV = "production", PG_REST_URL = "https://web3-storage-pgrest-prod.herokuapp.com", GATEWAY_URL = "https://ipfs.io" }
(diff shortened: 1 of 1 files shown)


vasco-santos (approved):
🎉 🎉 🎉 🎉 🎉

olizilla:
Note that the cron-pins job is still using the `CLUSTER_API_URL` value as set as a secret on this repo. We must change that value at least by the time we start letting pickup handle some prod traffic.

We can probably update it now, as pickup also proxies through to cluster, but I need to verify that we're not using that variable for anything else.

olizilla:
the `CLUSTER_API_URL` value that is set as a secret on the repo (as opposed to the same key that is set in the wrangler.toml) is used in the cron-pins workflow

https://github.com/search?q=repo%3Aweb3-storage%2Fweb3.storage+secrets.CLUSTER_API_URL&type=code

olizilla:
the cron-pins job and the pinning service implementation call the `status` and `statusAll` methods on our cluster client lib.

- `status` fetches `GET /pins/:cid` which pickup is set to handle and proxy to cluster as needed
- `statusAll` fetches `GET /pins?cids=:cid1,:cid2,...:cidN` which we need to implement a handler for on pickup before switching over.

PR #2180, opened 2023-01-05
feat: force users to have a subscription to the free tier if they do not already have a subscription

e-schneid:
This is part of the early adopter migration plan.  All of the active and high usage early adopters have already been migrated, but there remain inactive or low usage early adopters.  We (see [here](https://www.notion.so/Early-Adopter-to-Custom-Plan-Migration-Requirements-6d39bbb145ef4106ad00febcaf02b94c)) want to migrate these early adopters to the existing Free tier, but there are 20k+ such users and most of these may never log into the app again.

So this code just migrates the above users _as_ they request their subscription info.  It has the added benefit of doing a bit of cleanup if it so happens that we create a web3.storage user but their Stripe subscription setup fails (protecting data integrity).

### DIFF
--- a/packages/api/src/utils/billing.js
+++ b/packages/api/src/utils/billing.js
@@ -140,7 +140,15 @@ export async function getPaymentSettings (ctx) {
   if (paymentMethod instanceof Error) {
     throw paymentMethod
   }
-  const subscription = await ctx.subscriptions.getSubscription(customer.id)
+  let subscription = await ctx.subscriptions.getSubscription(customer.id)
+
+  // This ensures that all users have at least a subscription to the Free tier.  This is both for gradual migration
+  // of inactive early adopters, but also to repair data integrity issues that may arise during new account creation.
+  if (!(subscription instanceof Error) && subscription.storage === null) {
+    await ctx.subscriptions.saveSubscription(customer.id, { storage: { price: storagePriceNames.free } })
+    subscription = await ctx.subscriptions.getSubscription(customer.id)
+  }
+
   if (subscription instanceof Error) {
     return subscription
   }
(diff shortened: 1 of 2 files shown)


packages/api/src/utils/billing.js
```diff
@@ -140,7 +140,15 @@ export async function getPaymentSettings (ctx) {
   if (paymentMethod instanceof Error) {
     throw paymentMethod
   }
-  const subscription = await ctx.subscriptions.getSubscription(customer.id)
+  let subscription = await ctx.subscriptions.getSubscription(customer.id)
+
+  // This ensures that all users have at least a subscription to the Free tier.  This is both for gradual migration
+  // of inactive early adopters, but also to repair data integrity issues that may arise during new account creation.
+  if (!(subscription instanceof Error) && subscription.storage === null) {
```
gobengo, line 147: can we do this somewhere other than in `getPaymentSettings`? I'd prefer to avoid putting a potential side effect in a function starting with `get` to avoid confusion.

e.g. instead of updating the subscription as payment settings are fetched, can we do this kind of check at login or something? We already have logic on login for newly created users to get set their subscription set to 'free'.

e-schneid: I can look into adding this at login, but my thought was that although there *may* be side-effects putting it here, this GET request is still idempotent.  I looked at this more as data correction.  But I'll go down the path of seeing what's possible at login.

e-schneid: we can probably do it [here](https://github.com/web3-storage/web3.storage/blob/main/packages/api/src/user.js#L167).  I'll go ahead and implement this.  Let me know if you were thinking of another path.

gobengo: @e-schneid yep initiating it from there on login could be better.

>  but my thought was that although there may be side-effects putting it here, this GET request is still idempotent

Yes, it's idempotent when this `BillingService` is being invoked in the context of an http handler, but I'd like to keep the class useful in other contexts too, and I'd like this method to continue being useful to other contexts as a 'pure function' with no side effects. (it can still be used by a function that has side effects)

PR #2197, opened 2023-02-07
feat: add IsEarlyAdopter user tag

e-schneid:
Since early adopters are no longer being modeled strictly as "users without Stripe customers," @gobengo suggested that we use our tag mechanism to keep track of who our early adopters are (so that we don't need to look at Stripe to make those inferences).

After this is merge and the included db migration is performed, there will be a separate migration to actually tag our early adopters.

### DIFF
--- a/packages/api/src/constants.js
+++ b/packages/api/src/constants.js
@@ -8,6 +8,7 @@ export const PIN_STATUSES = ['PinQueued', 'Pinning', 'Pinned', 'PinError']
 export const USER_TAGS = {
   ACCOUNT_RESTRICTION: 'HasAccountRestriction',
   DELETE_RESTRICTION: 'HasDeleteRestriction',
+  EARLY_ADOPTER: 'IsEarlyAdopter',
   PSA_ACCESS: 'HasPsaAccess'
 }
 export const CAR_CODE = 0x202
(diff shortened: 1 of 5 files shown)


packages/api/src/constants.js
```diff
@@ -8,6 +8,7 @@ export const PIN_STATUSES = ['PinQueued', 'Pinning', 'Pinned', 'PinError']
 export const USER_TAGS = {
   ACCOUNT_RESTRICTION: 'HasAccountRestriction',
   DELETE_RESTRICTION: 'HasDeleteRestriction',
+  EARLY_ADOPTER: 'IsEarlyAdopter',
```
gobengo, line 11: Would exactly would this indicate?

Basically user records that were inserted before we shipped the first version of tiered pricing? or would it be something more nuanced/intentional for that? e.g. if a user was in early adopter status, but later chose to add a payment method and sign up for our most expensive subscription plan, would you expect that user to still have this tag?

another thought experiment. If we know we shipped tiered pricing on 2022-09-01, then I think we can just say 'all the users that were every early adopter are all the users that were inserted before 2022-09-01'.

What's an example of a user that wouldn't match that inserted_at query but that we would want to tag 'IsEarlyAdopter'?

e-schneid: > if a user was in early adopter status, but later chose to add a payment method and sign up for our most expensive subscription plan, would you expect that user to still have this tag?

My initial thought was yes, but I'm totally on board with just using that user insertion date in order determine if someone is an early adopter.  It might have just been a miscommunication, but I think what you're suggesting would work.

gobengo: @e-schneid k. the only thing missing to use that heuristic would be the exact date to use as cuttover.

e-schneid: I'll close the PR and address the date concerns in a later separate PR if needed.  I think our main interest is that we don't lose information about early adopters, and I think we've established that we're safe there since we'll always know the user insertion date.

PR #2200, opened 2023-02-09
fix: manage Stripe race condition on login

e-schneid:


packages/api/src/utils/stripe.js
```diff
@@ -256,7 +258,26 @@ export class StripeCustomersService {
...
       email: options?.email
     })
-    await this.userCustomerService.upsertUserCustomer(user.id.toString(), createdCustomer.id)
+
+    try {
+      await this.userCustomerService.insertUserCustomer(user.id.toString(), createdCustomer.id)
+    } catch(error) {
+      // due to potential race conditions, we want to ensure we only return an actual associated stripe customer
+      // and remove stripe customers that we fail to associate
+      if (hasOwnProperty(error, "code") && error.code === "CONSTRAINT_ERROR_DB") {
```
gobengo, line 267: where does this error come from?

e-schneid: this comes from postgrest.  I have the method checking for codes that start with '23' ([constraint violation codes](https://www.postgresql.org/docs/current/errcodes-appendix.html))

packages/api/src/utils/stripe.js
```diff
@@ -256,7 +258,26 @@ export class StripeCustomersService {
...
+    try {
+      await this.userCustomerService.insertUserCustomer(user.id.toString(), createdCustomer.id)
+    } catch (error) {
+      // due to potential race conditions, we want to ensure we only return an actual associated stripe customer
+      // and remove stripe customers that we fail to associate
+      if (hasOwnProperty(error, 'code') && error.code === 'CONSTRAINT_ERROR_DB') {
+        // no need to block during deletion
+        this.stripe.customers.del(createdCustomer.id)
+
+        existingCustomer = await this.userCustomerService.getUserCustomer(user.id.toString())
```
gobengo, line 271: imo keep `existingCustomer` const and use a different `const` name here

packages/api/src/utils/billing.js
```diff
@@ -167,14 +167,14 @@ export function createMockUserCustomerService () {
   /**
    * @returns {Promise<void>}
    */
-  const upsertUserCustomer = async (userId, customerId) => {
+  const insertUserCustomer = async (userId, customerId) => {
     userIdToCustomerId.set(userId, customerId)
```
gobengo, line 171: you might want to update this to throw on duplicates like the postgrest one will do

packages/api/src/utils/billing-types.ts
```diff
@@ -132,7 +132,7 @@ export type PaymentSettings = {
 
 export interface UserCustomerService {
   getUserCustomer: (userId: string) => Promise<null|{ id: string }>
-  upsertUserCustomer: (userId: string, customerId: string) => Promise<void>
+  insertUserCustomer: (userId: string, customerId: string) => Promise<void>
```
gobengo, line 135: we should capture in the types in kind of 'normal failure' that may happen, e.g. due to a uniqueness constraint, e.g. `UserAlreadyHasCustomer`

packages/api/src/utils/stripe.js
```diff
@@ -256,7 +258,26 @@ export class StripeCustomersService {
...
       email: options?.email
     })
-    await this.userCustomerService.upsertUserCustomer(user.id.toString(), createdCustomer.id)
+
+    try {
+      await this.userCustomerService.insertUserCustomer(user.id.toString(), createdCustomer.id)
+    } catch (error) {
+      // due to potential race conditions, we want to ensure we only return an actual associated stripe customer
+      // and remove stripe customers that we fail to associate
+      if (hasOwnProperty(error, 'code') && error.code === 'CONSTRAINT_ERROR_DB') {
```
gobengo, line 267: Try to hide the implementation details inside `insertUserCustomer` interface.

e.g. detecting a `UserAlreadyHasCustomer` error. `insertUserCustomer`, when implementation is powered by a db, can throw it when its underlying db sees this error

packages/api/src/utils/stripe.js
```diff
@@ -256,7 +258,26 @@ export class StripeCustomersService {
...
+        // no need to block during deletion
+        this.stripe.customers.del(createdCustomer.id)
```
gobengo, line 269: I don't think it's important to have this clean it up in stripe. We can detect them later on a cron job like you had been implementing.

If we want to take some action here to make that easier, we could write into a table on our side of 'known orphaned customer ids' so we can use that db later to drive a delete job.

but tl;dr I think it would be fine to not put cleanup logic here, and instead just document that it's a known possible side effect of a known race condition, and that it shouldn't affect the running behavior of the app (e.g. no tests fail because of it)

packages/api/src/utils/stripe.js
gobengo: doing try/catch with a more abstract error is a good idea regardless to hide the db details from this stripe implementation service that otherwise is abstracted away via the 'service' interfaces that can be injected in.

I would *also* encourage then avoiding try/catch alltogether so you can also have the type checker make sure callers handle that case where insertUserCustomer cannot complete due to the uniqueness constraint that is inherent in our domain model. So personally I'd also recommend having `insertUserCustomer` return `Promise<void | UserAlreadyHasCustomer>`

packages/db/index.js
```diff
@@ -1321,6 +1321,28 @@ export class DBClient {
...
+    const { error } = await this._client
+      .from('user_customer')
+      .insert({
+        user_id: userId,
+        customer_id: customerId
+      })
+      .single()
+
+    if (error) {
+      if (error.code.startsWith('23')) {
```
gobengo, line 1338: can you include a link to why this magic `23` works? what does it mean? what kind of constraintError?

packages/api/src/utils/stripe.js
```diff
@@ -256,7 +258,26 @@ export class StripeCustomersService {
...
+
+        existingCustomer = await this.userCustomerService.getUserCustomer(user.id.toString())
+
+        if (!existingCustomer) {
+          throw new CustomerNotFoundError()
```
gobengo, line 274: ... this should never happen right?

is there a test case that necessitates it? what should the caller of this func do if they encounter `CustomerNotFoundError`?

it seems like maybe this should never actually happen... and if it did.... it would be pretty surprising.... like maybe all this should just be `throw error` here and no need for CustomerNotFoundError which to me looks like it means `UnableToCreateCustomerForUser` or something (because you're throwing this in the context of encountering an error doing insertUserCustomer)

e-schneid:
we probably won't need this one.  I'll close it and reopen it later if that changes.  Our plan seems to be that we'd really just like to yank out the code causing the race condition rather than try to fix it.

PR #2216, opened 2023-02-21
chore: add metric tracking time to pin a 3MiB file.

olizilla:
Adds a step to the cron upload test job to pin a file to web3.storage via kubo and record the time it takes in prometheus

Uses https://github.com/ibnesayeed/setup-ipfs to setup and start kubo

There is a PR to add retries to pickup that should get us past the sproadic connection refusals that are causing this new test to fail currently https://github.com/web3-storage/pickup/pull/115

License: MIT

### DIFF
--- a/.github/workflows/cron-psa.yml
+++ b/.github/workflows/cron-psa.yml
@@ -1,0 +1,46 @@
+name: Cron PSA Test
+
+on:
+  push:
+    paths:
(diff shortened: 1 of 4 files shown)


olizilla:
The upload test is failing now because pinning to the prod api is currently failing occasionally see #2217

gobengo (approved):

PR #2226, opened 2023-02-27
fix: stop creating a Stripe account during web3.storage account creation

### LINKED ISSUE #2225: remove code that automatically signs up new user to free tier
Originally, we had automatically signed up all new users to the Free Tier, but this no longer necessary. Removing the code will simplify management in the future.

e-schneid:
Fixes #2225 

This eliminates a race condition causing the creation of multiple customers in Stripe.  New customers are now treated as if they're in the free tier without actually giving them a stripe customer.


gobengo (approved):

e-schneid:
This is tested in staging by creating a new account and confirming that there is no corresponding customer being create in Stripe. 

For manual regression testing:

1. Ensure that adding a payment method still works while the user has no customer in Stripe (as expected, this _will_ create a new customer in Stripe).
2. Check that names and emails are properly being updated in Stripe on subsequent logins to web3.storage (after adding a payment method or subscription)

PR #2232, opened 2023-03-03
feat: update to latest web3.storage@4.5

olizilla:
To start using the 50MB chunk size.

Thank you to @hammertoe for the nudge. This was overlooked when we made the change to the client lib, and increase the max size of files we can serve through freeway.

License: MIT


vasco-santos (approved):

alanshaw:
Thanks @hammertoe!

PR #2235, opened 2023-03-06
fix: update pin status from GET /pins even if failed

### LINKED ISSUE #2234: getPin should update pin status to failed if upstream says its failed
We previously avoided setting a pin status to failed as cluster would periodically retry to fetch things. pickup wont do that, so we should let the user know early and set the failed status when the upstream says it's failed.

olizilla:
Update pin status for pins regardless of status. Previously we only updated if the state was positive.

Pickup will not retry pins that timeout, so the change here ensures we inform users promptly where that happens.

Where the user has a pending pin, the request will now synchronously go check cluster/pickup, update the db and return the response. Previously the worker would respond early, and poll for a positive pin status from the upstream for up to 30s. Just checking the upstream once per request seems sufficient. We may want to rate limit this end point in the future, but it's pretty low volume currently.

fixes #2234 

License: MIT

### DIFF
--- a/packages/api/src/utils/pin.js
+++ b/packages/api/src/utils/pin.js
@@ -1,3 +1,4 @@
+import retry from 'p-retry'
 /**
  * @typedef {import('@nftstorage/ipfs-cluster').API.TrackerStatus} TrackerStatus
  * @typedef {'Undefined'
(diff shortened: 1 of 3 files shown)


packages/api/src/utils/pin.js
```diff
@@ -155,15 +156,33 @@ export async function waitOkPins (cid, cluster, waitTime = MAX_PIN_STATUS_CHECK_
...
   return okPins
 }
+
+/**
+ * @param {string} contentCid
+ * @param {import('@web3-storage/db/db-client-types').PinUpsertInput[]} pins
+ * @returns {import('@web3-storage/db/db-client-types').PinsUpsertInput[]}
+ */
+export function toPinsUpsert (contentCid, pins) {
+  // @ts-ignore
```
olizilla, line 170: in my defence, i have simply refactored and moved the @ts-ignore that was already there on line 166.

olizilla: ah ha. Those types are only used in this file and id is not needed by the impl.

olizilla: i have fixed

alanshaw (approved):

Issue #2237, opened 2023-03-07
docs button on site not working


heyjay44:
clicking on docs button doesn't take user to docs site.

e-schneid:
It appears that the auto-events.js code is interpreting https://web3.storage/docs as the URL to a download.  Explicitly removing the 'doc' extension from the data-extensions attribute fixes this, but we're doing this as a workaround for what appears to be a bug.

<Script data-extensions="pdf,csv,docx,xlsx,zip,xls" src="https://track.web3.storage/auto-events.js" />

This is tested by clicking the Docs link and noticing that we now properly navigate to the /docs page.

PR #2238, opened 2023-03-07
fix: apply workaround to Simple Analytics blocking navigation to /docs

e-schneid:
The Simple Analytics event handlers being attached to all of our links is interpreting the path `/docs` as a link to a download and blocking the default action of this event.  It appears this is because it's mistaking this path as a link to a .doc file.  Explicitly setting the file extensions to their defaults but removing the `doc` extension seems to fix this.  Probably a bad regex on their end.

See #2237


gobengo (approved):
lgtm!

PR #2241, opened 2023-03-10
fix: use car block validator

vasco-santos:
Use new lib to validate blocks https://github.com/web3-storage/car-block-validator so that we can support all the hashers. Alternatively, we can block on linkdex `getDagStructure` which currently is done in `waitUntil`.

Thoughts?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4433,6 +4433,15 @@
         }
       }
     },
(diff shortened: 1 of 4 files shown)


gobengo (approved):
nice :)

olizilla (approved):
👍

PR #2278, opened 2023-06-15
fix: add types to package.json exports

tennox:
Needed for newer typescript with certain moduleResolution settings.

Found the explanation & solution here:
https://github.com/gxmari007/vite-plugin-eslint/pull/60

& the docs for the option here:
https://nodejs.org/api/packages.html#packages_exports (with `types` being defined here: https://nodejs.org/api/packages.html#conditional-exports)

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -75,6 +75,7 @@
   "main": "./dist/src/lib.cjs",
   "exports": {
     ".": {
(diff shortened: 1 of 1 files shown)


tennox:
Test using:
```
pnpm add github:tennox/web3.storage#HACK-subrepo
```

gobengo (approved):

PR #2291, opened 2023-07-18
feat: write content claims

alanshaw:


Gozala (approved):

packages/api/src/car.js
```diff
@@ -116,22 +120,23 @@ export async function handleCarUpload (request, env, ctx, car, uploadType = 'Car
...
-  await Promise.all([
+  const [,, { cid: indexCid, carCid: indexCarCid }] = await Promise.all([
     putToS3(env, s3Key, carBytes, carCid, sourceCid, structure),
     putToR2(env, r2Key, carBytes, carCid, sourceCid, structure),
-    writeSatNavIndex(env, carCid, carBytes),
+    writeSatNavIndex(env, carCid, blockOffsets),
     writeDudeWhereIndex(env, contentCid, carCid)
   ])
 
+  await writeContentClaims(env, rootCid, carCid, indexCid, indexCarCid, structure, linkIndex)
```
vasco-santos, line 138: could this be done in parallel as above? We should have enough concurrent sub-requests

vasco-santos (approved):
LGTM, a few details commented inline.

packages/api/src/car.js
alanshaw: We need the index CID from the put to SATNAV, but I guess we could parallelise with the rest...

packages/api/test/car.spec.js
```diff
@@ -30,7 +33,9 @@ import {
 Object.assign(global, { fetch })
 
 describe('POST /car', () => {
-  it('should eventually add posted CARs to Cluster', async function () {
+  // FIXME: add to cluster failing with "multipart: NextPart: EOF"
+  // Code that is tested is not run in production...
+  it.skip('should eventually add posted CARs to Cluster', async function () {
```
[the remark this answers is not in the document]
alanshaw: https://github.com/web3-storage/web3.storage/issues/2300

Issue #2274, opened 2023-06-07
Put fails with ERR_INVALID_THIS


legowerewolf:
`push-storage.js`
```js
import { Web3Storage, getFilesFromPath } from "web3.storage";

const storage = new Web3Storage({ token: process.env.STORAGE_TOKEN });

const files = await getFilesFromPath("./ipfs/");

console.debug("files:", files);

const cid = await storage.put(files);

console.log("wrote CID:", cid);
```

```sh
> node push-storage.js
files: [
  {
    name: '/ipfs/index.html',
    stream: [Function: stream],
    mode: undefined,
    mtime: undefined,
    size: 778
  }
]
node:internal/errors:490
    ErrorCaptureStackTrace(err);
    ^

TypeError [ERR_INVALID_THIS]: Value of "this" must be of type URLSearchParams
    at new NodeError (node:internal/errors:399:5)
    at Proxy.has (node:internal/url:511:13)
    at Proxy.<anonymous> (file:///~~~/node_modules/@web-std/fetch/src/headers.js:140:44)
    at new Request (file:///~~~/node_modules/@web-std/fetch/src/request.js:105:38)
    at file:///~~~/node_modules/@web-std/fetch/src/index.js:40:19
    at new Promise (<anonymous>)
    at fetch (file:///~~~/node_modules/@web-std/fetch/src/index.js:38:9)
    at pRetry.retries (file:///~~~/node_modules/web3.storage/src/lib.js:221:30)
    at async RetryOperation._fn (~~~\node_modules\p-retry\index.js:50:12) {
  code: 'ERR_INVALID_THIS'
}

Node.js v20.2.0
```

Node 20.2.0
web3.storage 4.5.4
Windows 11 22H2 (22621.1778)

codingbiro:
Any update on this?

tomjohnhall:
Also hitting this error at the moment 😞

legowerewolf:
Downgrading isn't a long-term solution, though. It's been a month and it doesn't seem like the issue's even been reviewed by maintainers, which is a little disheartening.

PR #2292, opened 2023-07-19
fix: this param for fetch calls

alanshaw:
* Updates `@web-std/fetch` to `v4` with fix for `this` param
* Adds temporary fix for Node.js 20 - pass an `ArrayBuffer` as `fetch` body instead of a `Blob`
* Switches smoke to listen on `127.0.0.1` instead of `localhost` so node doesn't auto resolve a IPv6 address and fail to connect
* Tests client on Node.js 16, 18 and 20
* Splits bundlesize from `npm test` tasks since it cannot be run on node.js 20 (depends on unmaintained native module `iltorb` which does not compile on node.js 20)

resolves https://github.com/web3-storage/web3.storage/issues/2274


packages/client/package.json
```diff
@@ -53,7 +53,7 @@
     "@rollup/plugin-json": "^4.1.0",
     "@rollup/plugin-node-resolve": "^13.1.3",
     "@types/mocha": "9.0.0",
-    "bundlesize": "^0.18.1",
+    "body-parser": "^1.20.2",
```
alanshaw, line 56: ...was a missing dependency

packages/client/src/lib.js
```diff
@@ -184,10 +184,10 @@ class Web3Storage {
     }
     const targetSize = maxChunkSize
     const url = new URL('car', endpoint)
-    let headers = Web3Storage.headers(token)
-
-    if (name) {
-      headers = { ...headers, 'X-Name': encodeURIComponent(name) }
+    const headers = {
+      ...Web3Storage.headers(token),
+      'Content-Type': 'application/vnd.ipld.car',
```
alanshaw, line 189: Ensures we always get a `Content-Type` header even when we do the Node.js fix below.

packages/client/package.json
```diff
@@ -14,12 +14,12 @@
   "homepage": "https://github.com/web3-storage/web3.storage#readme",
   "scripts": {
     "test": "npm-run-all -p -r mock:api test:all",
-    "test:all": "run-s test:web test:cjs test:esm test:size",
-    "test:size": "bundlesize",
+    "test:all": "run-s test:web test:cjs test:esm",
+    "test:size": "npx bundlesize",
```
alanshaw, line 18: `npx` so that `npm install` in ndoe.js 20 doesn't fail because native module can't be compiled.

vasco-santos (approved):

packages/client/src/lib.js
Gozala: FYI: I have created https://github.com/ipld/js-car/pull/142 that we could reference official mime type, instead of having to copy that string all over.

PR #2296, opened 2023-07-23
fix: make blocking in ToS explicit

dchoi27:


dchoi27:
not sure why these tests are failing?

alanshaw:
e2e tests have been failing for a while...not sure why

PR #2298, opened 2023-07-24
fix: temporarily disable e2e tests

alanshaw:
These are currently failing as magic.link is failing to successfully log users in using the special test email `test+success@magic.link` (https://magic.link/docs/auth/introduction/test-mode#usage):

Screenshot 2023-07-24 at 16 58 19

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -48,132 +48,133 @@ jobs:
       - uses: bahmutov/npm-install@v1
       - run: npm run lint -w packages/website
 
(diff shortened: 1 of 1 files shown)


olizilla (approved):
you can disable jobs with an `if: false` property too...

https://stackoverflow.com/questions/75905932/how-to-always-disable-skip-a-job-in-github-actions-not-step-action

vasco-santos:
Should we create an issue to re-enable?

PR #2319, opened 2023-11-08
feat: add w3up-launch announcement banner (when env.NEXT_PUBLIC_W3UP_LAUNCH_SUNSET_ANNOUNCEMENT_START is set)

gobengo:
Motivation:
* https://github.com/web3-storage/secrets/issues/20

Tactics:
* there is a new env var `NEXT_PUBLIC_W3UP_LAUNCH_SUNSET_ANNOUNCEMENT_START` that is an ISO8601 datetime of when the announcement should appear. If unset, the announcement will never appear. My intention is to set it only on the staging environment for acceptance testing. Once we're happy on staging, we can enable it on production by setting this var to whatever datetime we want it to appear.

Scope:
* when enabled at after the appropriate time, the banner should appear on the following routes
  * /
  * /account
  * /account/payment
  * /docs
* the github actions `website` workflow now uses github environments. This is because I want to use them to configure the staging environment to have a particular value for `NEXT_PUBLIC_W3UP_LAUNCH_SUNSET_ANNOUNCEMENT_START`

### DIFF
--- a/packages/website/pages/index.js
+++ b/packages/website/pages/index.js
@@ -1,6 +1,7 @@
 import { useEffect } from 'react';
 
 import IndexPageData from '../content/pages/index.json';
+import Scroll2Top from '../components/scroll2top/scroll2top.js';
 import BlockBuilder from '../components/blockbuilder/blockbuilder.js';
 import { initFloaterAnimations } from '../lib/floater-animations.js';
 import { W3upMigrationRecommendationCopy, shouldShowSunsetAnnouncement } from '../components/w3up-launch.js';
@@ -35,6 +36,8 @@ export default function Home() {
           <BlockBuilder id={`section_${index + 1}`} key={`section_${index + 1}`} subsections={section} />
         ))}
       </main>
+
+      <Scroll2Top />
     </>
   );
 }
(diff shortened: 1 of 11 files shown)


packages/website/pages/index.js
```diff
@@ -23,13 +24,17 @@ export default function Home() {
...
+        </PageBannerPortal.PageBanner>
+      )}
+
       <main className="page page-index">
         {sections.map((section, index) => (
           <BlockBuilder id={`section_${index + 1}`} key={`section_${index + 1}`} subsections={section} />
         ))}
       </main>
-
-      <Scroll2Top />
```
gobengo, line 32: When this was present, it was weirdly somehow resulting in an element from inside the scroll2top component getting mixed up with the PageBanner children. It happened in firefox and chrome, and I have no idea how.... The scroll2top component code seems relatively innocuous so I don't know why it would affect dom manip. The only way to fix was to not use the component here. I know its dumb, but without this, the bug in the dom that react rendered made the home page banner not render at all.

workaround: remove this Scroll2Top for now. We can add it back in later if we really need it, but we probably don't, and this whole website is being sunset anyway!

PR #2329, opened 2023-11-13
ci: api.yml upgrade wrangler-action to 3.3.2 to use vars var

gobengo:
Motivation:
* i want to use `vars` for https://github.com/web3-storage/web3.storage/pull/2328 but it didn't work due to old wrangler action


gobengo:
I manually triggered this that runs the staging deploy using this new version, and it succeeded https://github.com/web3-storage/web3.storage/actions/runs/6856853839/job/18644979250

PR #2347, opened 2024-01-09
feat: listUploads includes parts property with list of links to CARs that contain it

gobengo:
Motivation:
* https://github.com/web3-storage/web3.storage/issues/2343

Why touch the sql files?
* the db tests wouldn't run cleanly without these changes. It would try to drop tables but fail due to dependencies on the uuid extension, other cols, etc. I think these changes are safe/good and will help others too but I can remove them if needed


gobengo:
example upload list item:
```json
{
    _id: '2',
    type: 'Upload',
    name: 'rand-07270196086601688',
    created: '2024-01-09T01:17:28.22+00:00',
    updated: '2024-01-09T01:17:28.22+00:00',
    cid: 'QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR',
    dagSize: 1000,
    pins: [ [Object] ],
    partOf: [
      'ipfs://bagbaiera6xcx7hiicm7sc523axbjf2otuu5nptt6brdzt4a5ulgn6qcfdwea'
    ],
    deals: []
}
```

packages/db/db-client-types.ts
```diff
@@ -218,6 +219,8 @@ export type UploadItemOutput = {
   dagSize?: definitions['content']['dag_size']
   pins: Array<PinItemOutput>,
   deals: Array<Deal>
+  // array of links to things containing this Upload (e.g. CARs)
+  partOf: Array<string>
```
alanshaw, line 223: Maybe just pluralise as `parts`? I like that this is similar to a [partition content claim](https://github.com/web3-storage/content-claims#partition-claim).

packages/db/test/upload.spec.js
```diff
@@ -278,6 +278,40 @@ describe('upload', () => {
...
+    assert.ok(uploads.length > 0)
+    for (const upload of uploads) {
+      // backupUrls raw is private
+      assert.ok(!('backupUrls' in upload), 'upload does not have backupUrls property')
+      assert.ok(Array.isArray(upload.partOf), 'upload.partOf is an array')
+    }
+    const namedUpload = uploads.find(u => u.name === name)
+    assert.deepEqual(namedUpload.partOf, [
+      // this corresponds to `exampleCarParkUrl`
+      'ipfs://bagbaiera6xcx7hiicm7sc523axbjf2otuu5nptt6brdzt4a5ulgn6qcfdwea'
```
alanshaw, line 311: I'd drop `ipfs://`, we're only going to strip it off when consuming this API and it'll never be anything different...

packages/db/utils.js
```diff
@@ -6,17 +6,39 @@
...
+    partOf
+  }
+}
+
+/**
+ * given array of backup_urls from uploads table, return a set of ipfs:// URIs for any CAR files in the backup_urls
+ * @param {string[]} backupUrls
+ * @returns {Iterable<string>}
+ */
+function carUrlsFromBackupUrls (backupUrls) {
```
alanshaw, line 32: You probably want something like this so you can read CAR CID from S3 URLs also:

https://github.com/web3-storage/content-claims/blob/50932114d51cfe34da113ab065c7b35927dbc6c5/packages/lambda/src/lib/store/block-index.js#L171-L192

alanshaw (changes requested):

packages/db/db-client-types.ts
gobengo: Will do.

Just to check my understanding. Do we ever expect there to be more than one CAR that the Upload is sharded across? In my light testing it looked like only CARs with all blocks included. So I was thinking we could make a slightly different claim than just 'it's spread across these parts' but instead 'it's entirely within this CAR, i.e. it is a subset/partOf these CARs'. But I did switch it to `parts` because that's a less opinionated claim here that might fit the data better.

packages/db/utils.js
gobengo: done!

packages/db/test/upload.spec.js
```diff
@@ -278,6 +278,41 @@ describe('upload', () => {
...
+    const { uploads } = await client.listUploads(user._id, { page: 1 })
+    assert.ok(uploads.length > 0)
+    for (const upload of uploads) {
+      // backupUrls raw is private
+      assert.ok(!('backupUrls' in upload), 'upload does not have backupUrls property')
+      assert.ok(Array.isArray(upload.parts), 'upload.parts is an array')
+    }
+    const namedUpload = uploads.find(u => u.name === name)
+    assert.deepEqual(namedUpload.parts, [
+      // this corresponds to `exampleCarParkUrl`
```
gobengo, line 311:
```suggestion
      // this corresponds to `exampleCarParkUrl` and the same CIDv1 but
      // parsed from the base32 multihash path of `exampleS3Url`
```

packages/db/db-client-types.ts
alanshaw: It can be multiple. If they send another CAR with the same root CID it _should_ be added to the list.

gobengo:
This patch would add in support for the backup table https://github.com/web3-storage/web3.storage/pull/2352

packages/api/src/magic.link.js
```diff
@@ -58,7 +58,12 @@ function createMagicTestmodeBypasss () {
 export const magicLinkBypassForE2ETestingInTestmode = createMagicTestmodeBypasss()
 
 function isMagicTestModeToken (token) {
-  const parsed = JSON.parse(globalThis.atob(token))
+  let parsed
+  try {
+    parsed = JSON.parse(globalThis.atob(token))
+  } catch (error) {
```
alanshaw, line 64:
```suggestion
  } catch {
```

alanshaw (approved):

gobengo:
✅ tested on staging
```shell
⚡ npm -s start -w packages/w3 -- list --json | head -n1 | jq .
{
  "_id": "315318962269342672",
  "type": "Car",
  "name": "Upload at 2024-01-22T19:10:44.048Z",
  "created": "2024-01-22T19:10:44.048+00:00",
  "updated": "2024-01-22T19:10:44.048+00:00",
  "cid": "bafybeieevwnu57cbcp5u6jsy6wxpj2waq5gfq5gc4spss4skpzk34vvxyy",
  "dagSize": 8848144,
  "pins": [
    {
      "status": "Pinned",
      "updated": "2024-01-22T19:10:44.048+00:00",
      "peerId": "bafzbeibhqavlasjc7dvbiopygwncnrtvjd2xmryk5laib7zyjor6kf3avm",
      "peerName": "elastic-ipfs",
      "region": null
    }
  ],
  "parts": [
    "bagbaierakuersmo7wndedhwk43e5xwcpzwenuda3dhpcsvkfibewg5gxl7oa"
  ],
  "deals": []
}
```

that part CID corresponds to https://bafybeieevwnu57cbcp5u6jsy6wxpj2waq5gfq5gc4spss4skpzk34vvxyy.ipfs-staging.w3s.link/

PR #2352, opened 2024-01-09
feat: list uploads joins to backup table to get backup urls

gobengo:
Motivation:
* add support for backup urls from backup table to https://github.com/web3-storage/web3.storage/pull/2347

### DIFF
--- a/packages/db/index.js
+++ b/packages/db/index.js
@@ -22,6 +22,7 @@ const uploadQuery = `
         created:inserted_at,
         updated:updated_at,
         backupUrls:backup_urls,
+        backup( url ),
         content(cid, dagSize:dag_size, pins:pin(status, updated:updated_at, location:pin_location(_id:id, peerId:peer_id, peerName:peer_name, ipfsPeerId:ipfs_peer_id, region)))
       `
 
--- a/packages/db/postgres/pg-rest-api-types.d.ts
+++ b/packages/db/postgres/pg-rest-api-types.d.ts
@@ -356,6 +356,102 @@ export interface paths {
       };
     };
   };
+  "/backup": {
+    get: {
+      parameters: {
+        query: {
+          id?: parameters["rowFilter.backup.id"];
+          upload_id?: parameters["rowFilter.backup.upload_id"];
+          url?: parameters["rowFilter.backup.url"];
+          inserted_at?: parameters["rowFilter.backup.inserted_at"];
+          /** Filtering Columns */
+          select?: parameters["select"];
+          /** Ordering */
+          order?: parameters["order"];
+          /** Limiting and Pagination */
+          offset?: parameters["offset"];
+          /** Limiting and Pagination */
+          limit?: parameters["limit"];
+        };
+        header: {
+          /** Limiting and Pagination */
+          Range?: parameters["range"];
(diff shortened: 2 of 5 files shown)


vasco-santos (commented):
Could we get more info about why we are adding this table? We actually had it before and moved towards having a column within upload table later on.

Also a point in migration that you may not know, we have a folder https://github.com/web3-storage/web3.storage/tree/main/packages/db/postgres/migrations where when we change SQL code we should create a migration to ease the release by running it. So these changes should be added there

gobengo:
> Could we get more info about why we are adding this table? We actually had it before and moved towards having a column within upload table later on.

@vasco-santos I'm mainly adding it here to reflect the reality of this table existing in staging/production, and so that the local dev flow creates it so the tests can query against it and pass, and I did that because I thought you, me, and @alanshaw all wanted to make sure that this uploads list api would include backup urls from that backup table in prod.

Now, I just want to 'add it' to the code so the local dev and ci/test setup have access to it. I don't want to (re-?)add it in staging/production, because I think it's already there. That's why I didn't add a migration, because I don't want to run a migration on staging/production. I want the localdev to reflect what is running in staging/prod, but only insofar as I can run tests locally that use that table.

Does that seem reasonable why there is not a migration or do you still think I should add one @vasco-santos ?

alanshaw:
@vasco-santos we never ran a migration to get the data out of the old backup table so it still exists in production. @gobengo just needs it to exist again for local dev and testing so that in production it all works.

vasco-santos:
Interesting we did not run the migration, the script was created https://github.com/web3-storage/web3.storage/pull/1305

Perhaps we could just run it now and get rid of backup table instead of making the list more complex?

alanshaw:
:warning: the migration may have run but a quick check and I have found entries in `backup` that are not in `upload.backup_urls`.

packages/db/index.js
```diff
@@ -22,6 +22,7 @@ const uploadQuery = `
         created:inserted_at,
         updated:updated_at,
         backupUrls:backup_urls,
+        backup( url ),
```
vasco-santos, line 25: I believe you should run `https://github.com/web3-storage/web3.storage/blob/main/packages/db/package.json#L13C5-L13C63` to update the types? but I do not remember anymore how this used to work

vasco-santos (approved):
LGTM

alanshaw:
@gobengo does this need to get deployed?

packages/db/index.js
gobengo: Thanks for suggesting this. I wouldn't have known. I ran that script and it produced this diff https://github.com/web3-storage/web3.storage/pull/2352/commits/df36cfbc116c89fa7988da50991c08a7d7a08d08

gobengo:
After deploying to staging via [this](https://github.com/web3-storage/web3.storage/actions/runs/8237844444), I verified `w3 list` calling the list uploads api still works ok. Going to release to prod.

gobengo:
i had to do this other PR to ensure this change to db package got incorporated into released api service https://github.com/web3-storage/web3.storage/pull/2362

verified in prod after [deploy](https://github.com/web3-storage/web3.storage/actions/runs/8238701978/job/22530426458)

PR #2374, opened 2025-02-14
fix(api): re-enable tokens endpoints outside read-write

hannahhoward:
# Goals

Re-enable creating API tokens

# Implementation

Currently the whole API is in read only mode. This allows tokens to be created and deleted by moving the handler outside the read/write mode protections


travis (approved):
seems right?!

